PR #52, opened 2019-08-16
Feature/authorizations component

annashamray:
**Changes:**
1. Include `authorizations` component to openzaak


src/openzaak/components/besluiten/api/permissions.py
```diff
@@ -1,34 +1,11 @@
...
-    Look at the scopes required for the current action and at informatieobjecttype and vertrouwelijkheidaanduiding
-    of related informatieobject and check that they are present in the AC for this client
-    """
-
-    permission_fields = ("besluittype",)
-    obj_path = "besluit"
-
-
-class BesluitRelatedAuthScopesRequired(RelatedObjAuthScopesRequired):
+class BesluitAuthRequired(AuthRequired):
```
sergei-maertens: Copy paste issue?

src/openzaak/components/documenten/api/viewsets.py
```diff
@@ -206,7 +204,7 @@ def get_serializer_class(self):
         """
         if self.action in ["update", "partial_update"]:
             return EnkelvoudigInformatieObjectWithLockSerializer
-        return EnkelvoudigInformatieObjectSerializer
+        return self.serializer_class
```
sergei-maertens:
```suggestion
        return super().get_serializer_class()
```
?

src/openzaak/utils/middleware.py
```diff
@@ -15,3 +33,72 @@ def __call__(self, request):
 
     def log(self, request: HttpRequest):
         logger.debug("Request headers for %s: %r", request.path, request.headers)
+
+
+class JWTAuth(_JWTAuth):
```
sergei-maertens: I'd move this to `components.authorizations.middleware` instead of keeping it in `utils`

src/openzaak/utils/middleware.py
```diff
@@ -15,3 +33,72 @@ def __call__(self, request):
...
+    def __init__(self, encoded: str = None):
```
sergei-maertens: wouldn't it make more sense to declare `component = None` as a class attribute and not override the `__init__`?

src/openzaak/utils/middleware.py
```diff
@@ -15,3 +33,72 @@ def __call__(self, request):
...
+        self.encoded = encoded
+        self.component = None
+
+    @property
+    def applicaties(self) -> Union[list, None]:
```
sergei-maertens: maybe wouldn't override the entire property, but re-implement `def _request_auth(self)` so that it always returns an empty list?

src/openzaak/utils/middleware.py
```diff
@@ -15,3 +33,72 @@ def __call__(self, request):
...
+
+        for autorisatie in autorisaties:
+            scopes_provided.update(autorisatie.scopes)
+
+        return scopes.is_contained_in(list(scopes_provided))
+
+
+class AuthMiddleware(_AuthMiddleware):
+    def extract_jwt_payload(self, request):
+        authorization = request.META.get(self.header, "")
```
sergei-maertens: nowadays there's `request.HEADERS` iirc

src/openzaak/utils/middleware.py
```diff
@@ -15,3 +33,72 @@ def __call__(self, request):
...
+        prefix = f"{self.auth_type} "
+        if authorization.startswith(prefix):
```
sergei-maertens: I'm not sure what's happening here?

src/openzaak/utils/middleware.py
annashamray: thanks! replaced with `request.headers`

src/openzaak/utils/middleware.py
annashamray: removing "Bearer " from the "Authorization" header value I suppose

src/openzaak/utils/middleware.py
annashamray: fixed

src/openzaak/utils/middleware.py
annashamray: `_request_auth(self)` is changed, `applicaties` is inherited from the parent class

src/openzaak/utils/middleware.py
annashamray: done

src/openzaak/utils/middleware.py
sergei-maertens: oooh of course

sergei-maertens:
One more go at resolving conflicts and then we can merge 🙂

annashamray:
@sergei-maertens one more rebase is done

sergei-maertens:
lgtm!

PR #60, opened 2019-08-22
Bring back ObjectInformatieObject

sergei-maertens:
- [x] Re-enable all tests
- [x] Make tests pass


README.rst
```diff
@@ -44,6 +44,13 @@ API's te realiseren die gebruikt kunnen worden bij gemeenten.
 
 .. _`referentie implementaties van VNG Realisatie`: https://github.com/VNG-Realisatie/gemma-zaken
 
+Requirements
+============
+
+* PostgreSQL database
+* Install extensions ``postgres`` and ``uuid-ossp``. If the DB user is a
```
joeribekker: Is this still relevant without triggers?

If so, make this a subbullet from the PostgreSQL database bullet. And, shouldn't postgis be here as well (with link to https://postgis.net/install/)?

src/openzaak/components/documenten/api/serializers.py
```diff
@@ -438,3 +442,55 @@ class Meta:
...
+        object_field = self.fields["object"]
+        if object_type == "besluit":
+            object_field.view_name = "besluit-detail"
+            object_field.queryset = Besluit.objects
+        else:
+            object_field.view_name = "zaak-detail"
+            object_field.queryset = Zaak.objects
+
+    def to_internal_value(self, data):
+        object_type = data.get("object_type")
```
joeribekker: Will `data` always include `object_type`? I don't think so if a partial update omits it.

It's safer to do:

```suggestion
        object_type = data.get("object_type", self.instance.object_type)
```

src/openzaak/components/documenten/api/tests/test_audittrails.py
```diff
@@ -76,7 +78,8 @@ def test_create_enkelvoudiginformatieobject_audittrail(self):
             informatieobject_create_audittrail.nieuw, informatieobject_data
         )
 
-    @skip("ObjectInformatieObject is not implemented yet")
+    @tag("oio")
+    @skip("Creation and destory OIO via its resource is not supported")
```
joeribekker:
```suggestion
    @skip("Creation and destroy OIO via its resource is not supported until support for external API's is added.")
```

src/openzaak/components/documenten/api/serializers.py
annashamray: but OIO resource doesn't support `update` and `partitial_update` operations

README.rst
annashamray: No, it's not relevant now. Fixed.

annashamray:
> I missed the code to prevent the `Manager/QuerySet.update` and `Manager/QuerySet.delete` functions where signals are used. If `.update` or `.delete` is used on a queryset, an exception should be raised if it involves the "URL-fields".

@joeribekker there are signals that process `delete` action on ZIO and BIO objects. And OIO resource doesn't' support `update` action. Should I still raise exception in case there is `update` of the queries?

src/openzaak/components/documenten/api/filters.py
```diff
@@ -37,3 +43,40 @@ class Meta:
             "startdatum": ["lt", "lte", "gt", "gte"],
             "einddatum": ["lt", "lte", "gt", "gte"],
         }
+
+
+def object_queryset(request):
+    object_value = request.query_params.get("object", "")
+    if "zaken" in object_value:
```
sergei-maertens: that's not a very elegant way to figure this out. Isn't it possible to leverage `vng_api_common.utils.get_viewset_for_path`, and then use `viewset.get_queryset().model._default_manager.all()`?

src/openzaak/components/documenten/api/filters.py
```diff
@@ -37,3 +43,40 @@ class Meta:
...
+        return Zaak.objects.all()
+    return Besluit.objects.all()
+
+
+class ObjectFilter(URLModelChoiceFilter):
+    def filter(self, qs, value):
+        if isinstance(value, Zaak):
+            self.field_name = "zaak"
+        else:
```
sergei-maertens: I'd make an explicit check here for `Besluit` instance and if it's none of both, raise a `TypeError`. Now this will silently do weird stuff if a third model gets added at some point.

src/openzaak/components/documenten/api/serializers.py
sergei-maertens:
```suggestion
        object_type = data["object_type"]
```
if you are not handling the `None` case from `.get(...)`, I prefer explicit key access.

src/openzaak/components/documenten/api/serializers.py
```diff
@@ -438,3 +442,55 @@ class Meta:
...
+        self.set_object_properties(object_type)
+        res = super().to_internal_value(data)
+        return res
+
+    def to_representation(self, instance):
+        object_type = instance.object_type
+        self.set_object_properties(object_type)
+        return super().to_representation(instance)
+
+    class Meta:
```
sergei-maertens: code style: can you define the meta class _before_ any methods?

src/openzaak/components/documenten/api/serializers.py
```diff
@@ -438,3 +442,55 @@ class Meta:
...
+        model = ObjectInformatieObject
+        fields = ("url", "informatieobject", "object", "object_type")
+        extra_kwargs = {"url": {"lookup_field": "uuid"}}
+
+    def __init__(self, *args, **kwargs):
```
sergei-maertens: code style: please put magic methods first in the class definition, then non-magic methods

src/openzaak/components/documenten/api/tests/test_audittrails.py
```diff
@@ -76,7 +78,10 @@ def test_create_enkelvoudiginformatieobject_audittrail(self):
             informatieobject_create_audittrail.nieuw, informatieobject_data
         )
 
-    @skip("ObjectInformatieObject is not implemented yet")
+    @tag("oio")
+    @skip(
+        "Creation and destroy OIO via its resource is not supported until support for external API's is added."
```
sergei-maertens: hrm, shouldn't this also just result in an audit trail entry? As in - the POST gets accepted, but nothing changes in the database, except the audit trail entry that gets created?

src/openzaak/components/documenten/api/viewsets.py
```diff
@@ -438,3 +450,132 @@ class EnkelvoudigInformatieObjectAuditTrailViewSet(AuditTrailViewSet):
...
+                        "object"
+                    ],
+                    "informatieobject": serializer.validated_data["informatieobject"],
+                }
+            )
+        except ObjectInformatieObject.DoesNotExist:
+            raise ValidationError(
+                {
+                    api_settings.NON_FIELD_ERRORS_KEY: _(
+                        "The relation between object and informatieobject don't exist"
```
sergei-maertens:
```suggestion
                        "The relation between object and informatieobject doesn't exist"
```

src/openzaak/components/documenten/api/viewsets.py
```diff
@@ -438,3 +450,132 @@ class EnkelvoudigInformatieObjectAuditTrailViewSet(AuditTrailViewSet):
...
+        if (
+            instance.object_type == "zaak"
+            and ZaakInformatieObject.objects.filter(
+                informatieobject=instance.informatieobject, zaak=instance.zaak
+            ).exists()
+        ):
+            raise ValidationError(
+                {
+                    api_settings.NON_FIELD_ERRORS_KEY: _(
+                        "The relation between zaak and informatieobject still exist"
```
sergei-maertens:
```suggestion
                        "The relation between zaak and informatieobject still exists"
```

src/openzaak/components/documenten/api/viewsets.py
```diff
@@ -438,3 +450,132 @@ class EnkelvoudigInformatieObjectAuditTrailViewSet(AuditTrailViewSet):
...
+        if (
+            instance.object_type == "besluit"
+            and BesluitInformatieObject.objects.filter(
+                informatieobject=instance.informatieobject, besluit=instance.besluit
+            ).exists()
+        ):
+            raise ValidationError(
+                {
+                    api_settings.NON_FIELD_ERRORS_KEY: _(
+                        "The relation between besluit and informatieobject still exist"
```
sergei-maertens:
```suggestion
                        "The relation between besluit and informatieobject still exists"
```

src/openzaak/components/documenten/migrations/0007_install_uuid_extension.py
```diff
@@ -0,0 +1,11 @@
...
+
+from django.db import migrations
+from django.contrib.postgres.operations import CreateExtension
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [("documenten", "0006_auto_20190822_1327")]
+
+    operations = [CreateExtension("uuid-ossp")]
```
sergei-maertens: this is no longer needed - entire migration can be deleted :fire:

sergei-maertens:
@annashamray @joeribekker w/r to the queryset/manager methods:

* `bulk_create` 100% bypasses signals, and should therefore be blocked on `ZaakInformatieObject`, `BesluitInformatieObject` and probably on `ObjectInformatieObject` as well
* `update` 100% bypasses signals - same as above. I'd say don't even bother with checking if the URL fields are involved or not, because you are not allowed to change relation information.
* `delete` sometimes bypasses signals, sometimes not. It depends on if related objects exist or not. The safe approach is to block it, and only allow `instance.delete(...)` which 100% fires signals.

joeribekker:
> @joeribekker there are signals that process delete action on ZIO and BIO objects. And OIO resource doesn't' support update action. Should I still raise exception in case there is update of the queries?

On QuerySet level I want to prevent these.

So, `ZIO-instance.delete()` should be allowed, but `ZIO.objects.all().delete()` should not.

src/openzaak/components/documenten/api/filters.py
annashamray: Thanks for the advice. Changed with the use of `get_viewset_for_path`

src/openzaak/components/documenten/api/filters.py
annashamray: fixed

src/openzaak/components/documenten/api/serializers.py
annashamray: done

src/openzaak/components/documenten/api/serializers.py
annashamray: done

src/openzaak/components/documenten/api/tests/test_audittrails.py
annashamray: I turned off `notifications` because objects are not created here and the notifications of creation ZIO and BIO would be just duplicated.
I turned off `audittrails` under the same logic, but I perhaps was mistaken. So `audittrails` are turned on again for OIO resource and the test case is unskipped

src/openzaak/components/documenten/migrations/0007_install_uuid_extension.py
annashamray: Thanks! I forgot to clean it

src/openzaak/components/documenten/api/tests/test_audittrails.py
sergei-maertens: Brought it under upstreams attention: https://github.com/VNG-Realisatie/gemma-zaken/issues/1283

sergei-maertens:
@annashamray the blocked `queryset.update`, `queryset.delete` and `manager.bulk_create` methods are still missing, or am I blind?

annashamray:
> @annashamray the blocked `queryset.update`, `queryset.delete` and `manager.bulk_create` methods are still missing, or am I blind?

@sergei-maertens blocking these methods is added to `BIO`, `ZIO` and `OIO` queries

src/openzaak/components/besluiten/models/tests/test_bio_block_change.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def test_update(self):
+        self.assertRaises(
+            TypeError,
+            BesluitInformatieObject.objects.update,
+            aard_relatie=RelatieAarden.hoort_bij,
+        )
+
+    def test_delete(self):
+        self.assertRaises(TypeError, BesluitInformatieObject.objects.delete)
```
sergei-maertens: `Manager.delete()` normally doesn't exist - see [the docs](https://docs.djangoproject.com/en/2.2/topics/db/managers/#creating-a-manager-with-queryset-methods)

```suggestion
        self.assertRaises(TypeError, BesluitInformatieObject.objects.all().delete)
```

src/openzaak/components/documenten/models/tests/test_oio.py
```diff
@@ -0,0 +1,106 @@
...
+        ZaakInformatieObjectFactory.create()
+        cls.oio = ObjectInformatieObject.objects.get()
+
+    def test_update(self):
+        self.assertRaises(
+            TypeError, ObjectInformatieObject.objects.update, object_type="besluit"
+        )
+
+    def test_delete(self):
+        self.assertRaises(TypeError, ObjectInformatieObject.objects.delete)
```
sergei-maertens:
```suggestion
        self.assertRaises(TypeError, ObjectInformatieObject.objects.all().delete)
```

src/openzaak/components/zaken/models/tests/test_zio_block_change.py
```diff
@@ -0,0 +1,31 @@
...
+    def setUpTestData(cls):
+        super().setUpTestData()
+
+        cls.zio = ZaakInformatieObjectFactory.create()
+
+    def test_update(self):
+        self.assertRaises(TypeError, ZaakInformatieObject.objects.update, titel="new")
+
+    def test_delete(self):
+        self.assertRaises(TypeError, ZaakInformatieObject.objects.delete)
```
sergei-maertens:
```suggestion
        self.assertRaises(TypeError, ZaakInformatieObject.objects.all().delete)
```

src/openzaak/utils/query.py
```diff
@@ -0,0 +1,19 @@
+from django.db.models import QuerySet
+
+
+class BlockChangeMixin:
+    @property
+    def msg(self):
+        return f"This method is forbidden for {self.model.__name__} queryset"
+
+    def bulk_create(self, objs, batch_size=None, ignore_conflicts=False):
+        raise TypeError(self.msg)
```
sergei-maertens: I'd probably change `TypeError` into a custom exception - it's a bit too generic and doesn't fit the purpose exactly.

annashamray:
> @annashamray please scroll through the feedback, I'll wrap up this branch with the changes. Nothing exciting, except maybe the bit that `delete` method does not exist on a mananger, but only on querysets.

Thank you for catching it and explaining. Now I know

PR #63, opened 2019-08-28
Fixes #55 -- Pin all (dev/test/production) dependencies

### LINKED ISSUE #55: Dependencies aren't pinned exactly
**Describe the bug**

Only the `requirements/base.in` is used as input to pip-compile. The production/dev requirements should also be frozen, as we've run into strange issues between minor versions of isort for example.

sergei-maertens:
Interesting things happen if the minor versions of dev-tooling aren't
the same in all developers environments, which is why _everything_
needs to be pinned.


joeribekker:
I understand the issue. Originally I removed these pins to make the structure more clear.

I still don't like "test" and "testing" confusion, and in comments it says "CI"... Can we pick one? :)

sergei-maertens:
@joeribekker renamed them to `ci.txt` and kept `testing.in`, is that okay?

PR #73, opened 2019-09-04
Fixes #71 -- remove duplicated tests & unused static assets

### LINKED ISSUE #71: As a maintainer, I want to prevent/clean up technical debt
... so that I am not distracted by quirkiness of the code while implementing new features.

**Description**

Since we based the code on the reference implementations, there's some code duplication going on, especially in the generic tests.

It makes sense to reorganize the code so that only useful stuff is there and it has a logical structure.

sergei-maertens:
Clean up of duplicated DSO API strategy tests

* moved them to the 'generic' part of the codebase
* refactored the tests to not use actual API resources (= guarantees we test global behaviour)
* made a TODO of features to be expected and tested, and deleted the `@skip`ed tests
* refactored some tests from ZTC that are 'global' and un-`@skip`ed them


annashamray:
Are test cases, that request OAS schemas, left only for `authorizations` component?
I thought test cases that get `schema-json-{component}` for each component are quite useful.

sergei-maertens:
@annashamray oh yeah I missed that - currently they're all the same version but those will start diverging after a while :thinking: 

I'll make the test cases so that the component is pluggable and we've got all of them covered.

PR #84, opened 2019-09-18
Fix broken link to license.md

ElenaFdR:
Switched from British to American spelling to match filename.


sergei-maertens:
Thanks for spotting this!

PR #95, opened 2019-09-24
Support external besluittypen

sergei-maertens:
Adds support for external resources, i.e. resources referenced by URL that do not live in our own database.


src/openzaak/components/besluiten/openapi.yaml
```diff
@@ -52,7 +101,6 @@ paths:
         required: false
         schema:
           type: string
-          format: uri
```
sergei-maertens: Must not dissapear

Issue #75, opened 2019-09-05
Creating Besluit with a Zaak gives zaaktype-mismatch even when besluittype is in Zaaktype.besluittypen


stevenbal:
**Describe the bug**
If I create a `Besluittype`, then a `Zaaktype` (with the previously created `Besluittype` in `Zaaktype.besluittypen`), then a `Zaak` with the created `Zaaktype` and then a `Besluit` with the created `Zaak` as `Besluit.zaak`, the `Besluit` creation fails and gives the following error:

```json
{
    "type": "http://localhost:8000/ref/fouten/ValidationError/",
    "code": "invalid",
    "title": "Invalid input.",
    "status": 400,
    "detail": "",
    "instance": "urn:uuid:ea08eede-56da-42d9-b668-ef42518f57ba",
    "invalidParams": [
        {
            "name": "nonFieldErrors",
            "code": "zaaktype-mismatch",
            "reason": "De referentie hoort niet bij het zaaktype van de zaak."
        }
    ]
}
```

Even though I explicitly added the `Besluittype` to the `Zaaktype`.

annashamray:
@stevenbal Have you published your `zaaktype` and `besluittype` objects after their creation? 
When the `besluit` with `zaak` attribute is being created the validator checks if the relation exists between **published** related  `zaaktype` and `besluittype` (i.e. `concept` field = False).
I tried to recreate the bug, but I could do it only with non-published `zaaktype` and `besluittype`

@sergei-maertens @joeribekker Now I start to think that there is a logic gap here. Because it is possible to create `besluit` or `zaak` related to non-published `besluittype` or `zaaktype`, but the relation between them is impossible to create. 
I think it should be forbidden to create any `zaak`, `besluittype` or `informatieobjecttype` related to non-published *type.

sergei-maertens:
Correct observation from what I quickly read. Can you bring this up in
gemma-zaken?

stevenbal:
> Have you published your zaaktype and besluittype objects after their creation?
When the besluit with zaak attribute is being created the validator checks if the relation exists between published related zaaktype and besluittype (i.e. concept field = False).
I tried to recreate the bug, but I could do it only with non-published zaaktype and besluittype

This is currently not the case in the reference components, there the validator does not filter out non-concept zaak/besluittypes when creating besluiten

annashamray:
> This is currently not the case in the reference components, there the validator does not filter out non-concept zaak/besluittypes when creating besluiten

[...]

PR #106, opened 2019-10-08
Feature/concept validation

annashamray:
Issue - https://github.com/open-zaak/open-zaak/issues/75

**Changes:**
1. Forbid creation of `zaak`, `informatieobject` and `besluit` if the relative type is not published
1. Forbid publishing `zaaktype` if it has relations with unpublished `besluittype` or `informatieobjecttype`
1. Remove filter on `concept=False` in relation validator for `Zaak` and `Besluit`


src/openzaak/components/catalogi/api/viewsets/zaaktype.py
```diff
@@ -84,3 +91,25 @@ class ZaakTypeViewSet(
...
+
+        # check related objects
+        besluittypen = instance.besluittype_set.all()
+        informatieobjecttypen = instance.heeft_relevant_informatieobjecttype.all()
+
+        for types in [besluittypen, informatieobjecttypen]:
+            for relative_type in types:
+                if relative_type.concept:
+                    msg = _("All relative resources should be published")
+                    raise PermissionDenied(detail=msg)
```
sergei-maertens: Should be 400 response

src/openzaak/components/catalogi/tests/test_zaaktype.py
```diff
@@ -277,6 +278,33 @@ def test_publish_zaaktype(self):
...
+    def test_publish_zaaktype_fail_not_concept_besluittype(self):
+        zaaktype = ZaakTypeFactory.create()
+        besluittype = BesluitTypeFactory.create()
+        zaaktype.besluittype_set.add(besluittype)
+
+        zaaktype_url = get_operation_url("zaaktype_publish", uuid=zaaktype.uuid)
+
+        response = self.client.post(zaaktype_url)
+
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
```
sergei-maertens: should be a 400 response

src/openzaak/components/catalogi/tests/test_zaaktype.py
```diff
@@ -277,6 +278,33 @@ def test_publish_zaaktype(self):
...
+
+    def test_publish_zaaktype_fail_not_concept_iotype(self):
+        zaaktype = ZaakTypeFactory.create()
+        ZaakInformatieobjectTypeFactory.create(zaaktype=zaaktype)
+
+        zaaktype_url = get_operation_url("zaaktype_publish", uuid=zaaktype.uuid)
+
+        response = self.client.post(zaaktype_url)
+
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
```
sergei-maertens: should be a 400 response

src/openzaak/components/besluiten/api/validators.py
```diff
@@ -29,9 +29,7 @@ def __call__(self, attrs):
         if not zaak:
             return
 
-        if not besluittype.zaaktypes.filter(
-            id=zaak.zaaktype_id, concept=False
-        ).exists():
+        if not besluittype.zaaktypes.filter(id=zaak.zaaktype_id).exists():
```
sergei-maertens: I don't understand this change?

When you're creating `besluit`, the `besluittype` must be published. If you are relating it to a `zaak`, the zaak must exist and have a published zaaktype. So, both besluittype and zaaktype should be published.

Or is that the logic you're using because the `concept` check actually doesn't matter here anymore?

annashamray: Yes, the idea was:
* make realization consistent with reference implementation realization, where there is no filter on `concept`
* get rid off excessive code: in this case we already know that `zaaktype` is published since related `zaak` exists

src/openzaak/components/catalogi/api/viewsets/zaaktype.py
annashamray: PermissionDenied is replaced with ValidationError

src/openzaak/components/catalogi/tests/test_zaaktype.py
annashamray: done

src/openzaak/components/catalogi/tests/test_zaaktype.py
annashamray: done

Issue #132, opened 2019-10-22
Filtering on m2m fields in Catalogi API returns internal server error


stevenbal:
**Describe the bug**
When I try to filter `BesluitType` by the m2m fields `Besluittype.zaaktypes` or `Besluittype.besluittypes`, I get an internal server error

**To Reproduce**

Make a request to the `BesluitType` endpoint and filter on `zaaktypes`:
```
GET localhost:8000/catalogi/api/v1/besluittypen?zaaktypes=http%3A%2F%2Flocalhost%3A8000%2Fcatalogi%2Fapi%2Fv1%2Fzaaktypen%2Fbd2f53b3-249f-4360-9412-9be9d7c1f9a4
```

Response:
```json
ObjectDoesNotExist at /catalogi/api/v1/besluittypen
URL did not resolve

Request Method: GET
Request URL: http://localhost:8000/catalogi/api/v1/besluittypen?zaaktypes=http%3A%2F%2Flocalhost%3A8000%2Fcatalogi%2Fapi%2Fv1%2Fzaaktypen%2Fbd2f53b3-249f-4360-9412-9be9d7c1f9a4
Django Version: 2.2.4
Python Executable: /home/steven/Documents/VNG/open-zaak/env/bin/python
Python Version: 3.7.4
```

stevenbal:
issue in gemma-zaken https://github.com/VNG-Realisatie/gemma-zaken/issues/1411

Issue #122, opened 2019-10-15
When filtering on resources in the APIs with invalid urls, an empty list should be returned


stevenbal:
**Describe the bug**
When I try to filter on a resource, for instance I try to filter `Gebruiksrechten` by `informatieobject`, and I pass a url that does not point to an informatieobject, I get a 400 error with code invalid

**To Reproduce**
Try to filter on any url on gebruiksrechten:
```html
{{drc_url}}/gebruiksrechten?informatieobject=https://google.com
```

and this response is returned:
```json
{
    "type": "http://localhost:8000/ref/fouten/ValidationError/",
    "code": "invalid",
    "title": "Invalid input.",
    "status": 400,
    "detail": "",
    "instance": "urn:uuid:eb2a107d-e52e-4930-ad56-474b4ddc9cc4",
    "invalidParams": [
        {
            "name": "informatieobject",
            "code": "invalid",
            "reason": "Ongeldige resource URL opgegeven"
        }
    ]
}
```

**Expected behavior**
I would expect filtering on urls that do not point to the correct resource to return an empty list, instead of a 400 error

stevenbal:
see also issue in gemma-zaken: https://github.com/VNG-Realisatie/gemma-zaken/issues/1381

PR #133, opened 2019-10-22
Issue/filtering on non existent gives 400

stevenbal:
fixes: https://github.com/open-zaak/open-zaak/issues/132
and once https://github.com/VNG-Realisatie/vng-api-common/pull/107 is merged and the version is bumped, will also fix: https://github.com/open-zaak/open-zaak/issues/122


src/openzaak/components/catalogi/api/filters.py
```diff
@@ -143,6 +155,7 @@ class BesluitTypeFilter(FilterSet):
             "Het INFORMATIEOBJECTTYPE van informatieobjecten waarin besluiten van dit "
             "BESLUITTYPE worden vastgelegd."
         ),
+        validators=[URLValidator()],
```
annashamray: Do I get it right that we couldn't use urls without hostname (for example '/api/v1/zaken/<uuid>') as filter parameters anymore?

src/openzaak/components/catalogi/tests/test_filters.py
```diff
@@ -0,0 +1,237 @@
...
+                response = self.client.get(reverse(BesluitType), {query_param: "bla"})
+
+                self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+
+                error = get_validation_errors(response, query_param)
+                self.assertEqual(error["code"], "invalid")
+
+    def test_filter_by_valid_url_object_does_not_exist(self):
+        for query_param in ["catalogus", "zaaktypes", "informatieobjecttypes"]:
+            with self.subTest(query_param=query_param):
```
annashamray: can you please create `Besluittype` object with no m2m relations to zaaktypes and informatieobjecttypes before making a request?

src/openzaak/components/catalogi/api/filters.py
stevenbal: that's correct

annashamray: I don't think it is consistent with other operations since we still can include urls without hostname in the body in POST action. And it adds extra work to modify performance tests, which I don't want to, because hostname will have to be hardcoded.
But if you and @sergei-maertens are ok with it, I am also ok

src/openzaak/conf/includes/base.py
```diff
@@ -280,6 +280,7 @@
             "level": "DEBUG",
             "propagate": False,
         },
+        "vng_api_common": {"handlers": ["console"], "level": "INFO", "propagate": True},
```
stevenbal: @sergei-maertens it seemed that logging from vng_api_common did not work yet, so I added it to openzaak

src/openzaak/components/catalogi/tests/test_filters.py
annashamray: could you please add `concept=False` when creating Besluittype? This can affect the result of the request

stevenbal: I have added `concept=False` (https://github.com/open-zaak/open-zaak/pull/133/commits/22337ffa7ffe86af9dcaa7fb6498ea0bd106dde5) where needed and the tests did indeed fail after that, so I fixed the issue in https://github.com/VNG-Realisatie/vng-api-common/pull/107

I also added factories to the other filter tests and that made them fail as well, so I fixed that too in the vng_api_common PR

Thanks for the feedback

src/openzaak/components/catalogi/api/filters.py
sergei-maertens: The URLs without hostname in bodies working as expected is an unintended side-effect. We always had the intention to use fully qualified URLs

sergei-maertens:
vng-api-common@1.0.27 released

src/openzaak/components/besluiten/tests/test_filters.py
```diff
@@ -32,3 +36,48 @@ def test_filter_besluittype(self):
...
+
+    def test_filter_by_valid_url_object_does_not_exist(self):
+        response = self.client.get(
+            reverse(Besluit), {"besluittype": "https://google.com"}
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(
+            response.data,
+            OrderedDict(
```
sergei-maertens: you can compare with a regular dict instead of `OrderedDict`

Issue #121, opened 2019-10-15
PUT and PATCH should be possible on resources in Catalogi API


stevenbal:
**Describe the bug**
Currently it is not possible to do (partial) updates on resources in the Catalogi API, this should be possible

Validation rules should apply to this as described in the [standard](https://github.com/VNG-Realisatie/gemma-zaken/blob/master/docs/_content/standaard/catalogi/index.md#concepten)

see also the PR for this issue in the reference implementation: https://github.com/VNG-Realisatie/gemma-zaaktypecatalogus/pull/75

joeribekker:
@stevenbal Can you port this to Open Zaak?

PR #137, opened 2019-10-24
Feature/put patch operations

stevenbal:
fixes https://github.com/open-zaak/open-zaak/issues/121


src/openzaak/components/catalogi/api/viewsets/mixins.py
```diff
@@ -55,29 +56,26 @@ class ConceptMixin(ConceptPublishMixin, ConceptDestroyMixin, ConceptFilterMixin)
...
-            raise PermissionDenied(detail=msg)
-
-        super().perform_create(serializer)
-
-
 class ZaakTypeConceptDestroyMixin(ConceptDestroyMixin):
     def get_concept(self, instance):
         return instance.zaaktype.concept
 
+    def perform_destroy(self, instance):
```
annashamray: I don't quite understand why to rewrite this method, because logic remains the same. An error code can be put into class variable I think.
Or if you want to create your own `perform_destroy` the inheritance of `ConceptDestroyMixin` is not needed. In current realization you will invoke `ConceptDestroyMixin.perform_destroy` method which will never raise an error.

src/openzaak/components/catalogi/api/viewsets/zaaktype.py
```diff
@@ -80,7 +74,47 @@ class ZaakTypeViewSet(
...
+    def perform_create(self, serializer):
+        for field_name in self.relation_fields:
+            field = serializer.validated_data.get(field_name, [])
+            for related_object in field:
+                # TODO fix url lookup to check if concept
+                if not related_object["gerelateerd_zaaktype"]:
+                    msg = _(
+                        f"Relations to non-concept {field_name} object can't be created"
+                    )
+                    raise PermissionDenied(detail=msg)
```
annashamray: can you please change `PermissionDenied` into `ValidationError`? I think all `PermissionDenied` in context of `concept field should be replaced to `ValidationError`. @sergei-maertens can you confirm (or not confirm ) it? 

I did it for ZTC in reference implementation (PR - https://github.com/VNG-Realisatie/gemma-zaaktypecatalogus/pull/77)

src/openzaak/components/catalogi/api/viewsets/zaaktype.py
```diff
@@ -80,7 +74,47 @@ class ZaakTypeViewSet(
...
+        super().perform_create(serializer)
+
+    def perform_update(self, instance):
+        for field_name in self.relation_fields:
+            field = getattr(self.get_object(), field_name)
+            # TODO fix url lookup to check if concept
+            related_non_concepts = field.filter(zaaktype__concept=False)
+            if related_non_concepts.exists():
+                msg = _(f"Objects related to non-concept {field_name} can't be updated")
+                raise PermissionDenied(detail=msg)
```
annashamray: please see comment above

src/openzaak/components/catalogi/api/viewsets/zaaktype.py
```diff
@@ -80,7 +74,47 @@ class ZaakTypeViewSet(
...
+    def perform_destroy(self, instance):
+        for field_name in self.relation_fields:
+            field = getattr(instance, field_name)
+            # TODO fix url lookup to check if concept
+            related_non_concepts = field.filter(zaaktype__concept=False)
+            if related_non_concepts.exists():
+                msg = _(
+                    f"Objects related to non-concept {field_name} can't be destroyed"
+                )
+                raise PermissionDenied(detail=msg)
```
annashamray: please see comment above

src/openzaak/components/catalogi/api/viewsets/zaaktype.py
stevenbal: This code was for the concept business logic between related zaaktypen (`gerelateerdeZaaktypen`), but I discussed with Joeri and we have decided to not apply this logic to related zaaktypes, because otherwise it would be impossible to publish them (i.e. if I have zaaktype X and zaaktype Y that are related, I can't publish X because Y is still a concept, and I can't publish Y because X is still a concept).

So I will remove this code altogether

Issue #118, opened 2019-10-15
As a developer I want a scope that allows deleting of published resources in Catalogi API


stevenbal:
... allowing these resources to be deleted when testing the standard or the OAS

see also: https://github.com/VNG-Realisatie/gemma-zaken/issues/1351

PR #150, opened 2019-10-31
Feature/scope force delete published types

stevenbal:
implements https://github.com/open-zaak/open-zaak/issues/118


joeribekker:
*Keeping the branch to make it easy to apply in the standard later*

Issue #154, opened 2019-11-01
As a developer I want to optimize the amount of DB queries


annashamray:
LIST actions of all resources should be analyzed for the amount of DB queries. 
(for dev config they log into "logs/django.log" file)

This amount of DB queries should be minimized using `select_related` and `prefetch_related`

PR #168, opened 2019-11-06
optimize db queries - part 2 (Infromatieobject related)

annashamray:
Issue - https://github.com/open-zaak/open-zaak/issues/154

DB queries reduction (for 100 observations):

1. Zaken
* `GET /zaken/api/v1/zaakinformatieobjecten`
Before: 309
After: 10

2. Documenten
* `GET /documenten/api/v1/objectinformatieobjecten`
Before: 309
After: 10

* `GET /documenten/api/v1/gebruiksrechten`
Before: 206
After: 7

3. Besluiten
* `GET /besluiten/api/v1/besluitinformatieobjecten`
Before: 306
After: 7


src/openzaak/components/documenten/models.py
```diff
@@ -317,6 +318,7 @@ class Meta:
         verbose_name = _("Enkelvoudige informatie object")
         verbose_name_plural = _("Enkelvoudige informatie objecten")
         indexes = [models.Index(fields=["canonical", "-versie"])]
+        ordering = ["canonical", "-versie"]
```
sergei-maertens: I'm generally not a fan of this because it messes badly with `GROUP BY` queries, but in this case, I'll allow it. Django is actually working on not using that `ordering` in those queries in the future.

Issue #151, opened 2019-10-31
ZaakInformatieObject allows PUT and PATCH but updates are blocked by signals in Documenten API


stevenbal:
**Describe the bug**
According to the API spec of [Zaken API](https://test.openzaak.nl/zaken/api/v1/schema/#operation/zaakinformatieobject_update), it should be possible to do PUT and PATCH on `ZaakInformatieObject`, specifically on the fields `titel` and `beschrijving` (as `zaak` and `informatieobject` are immutable). However, when I try to do an update on a `ZaakInformatieObject`, I get the following error:

```python
File "/home/steven/Documents/VNG/open-zaak/src/openzaak/components/documenten/signals.py", line 40, in sync_oio
    raise RuntimeError("Updates to relation information are not allowed.")
RuntimeError: Updates to relation information are not allowed.
```

Should updates be allowed on the fields `titel` and `beschrijving`? Or should updates not be allowed altogether?

sergei-maertens:
They should be allowed on those fields, this is a bug. We should probably check what change(s), check if it's a local FK and then allow the update to happen.

The API validation should have the relation fields immutable, the rest should work.

PR #208, opened 2019-11-08
Issue/zio put patch

stevenbal:
fixes https://github.com/open-zaak/open-zaak/issues/151

also removed PUT and PATCH from BesluitInformatieObject, as the only two fields (zaak and informatieobject) are immutable


src/openzaak/components/besluiten/openapi.yaml
```diff
@@ -1368,12 +1368,10 @@ paths:
           format: uri
       - name: informatieobject
         in: query
-        description: URL-referentie naar het INFORMATIEOBJECT (in de Documenten API)
```
joeribekker: Please put this back

See: https://github.com/VNG-Realisatie/gemma-zaken/blob/master/api-specificatie/brc/1.0.x/openapi.yaml#L1395

src/openzaak/components/besluiten/openapi.yaml
```diff
@@ -1368,12 +1368,10 @@ paths:
...
-          waarin (een deel van) het besluit beschreven is.
+        description: ''
         required: false
         schema:
           type: string
-          format: uri
```
joeribekker: And this.

See: https://github.com/VNG-Realisatie/gemma-zaken/blob/master/api-specificatie/brc/1.0.x/openapi.yaml#L1400

stevenbal:
I have put back the description and type in the OAS

PR #209, opened 2019-11-08
Admin fixes/improvements

sergei-maertens:
Changed all interaction with selectielijst into dropdowns. If zaaktype has procestype selected, then resultaattypes selectielijst values are filtered on that procestype.

All API calls are cached - selectielijst-things are cached for a full day as it shouldn't really change in that time span, resources that may change are cached for 1 hour.


joeribekker:
@sergei-maertens can you look at all related issues and close or update them accordingly?

I changed most of them this morning with additional issue descriptions.

sergei-maertens:
Build is failing because of HTTP 429 on the mock URLs :/

Issue #160, opened 2019-11-01
Admin inlines for Catalogus do not show objects that belong to it


stevenbal:
**Describe the bug**
In the admin for Catalogi, besluittypen/zaaktypen/informatieobjecttypen that belong to a certain `Catalogus` are not visible in the inlines at the bottom of the page

![Screenshot from 2019-11-01 14-52-59](https://user-images.githubusercontent.com/29249171/68029368-595d1d00-fcb7-11e9-9b99-a56eae470bcd.png)

annashamray:
@stevenbal 
Can you check it now please? I can't reproduce the bug
`Zaaktypen`:
![image](https://user-images.githubusercontent.com/47745906/68401325-d554da80-0179-11ea-9c35-ce91df935206.png)

`Besluittypen` and `Informatieobjecttypen` are also visible

stevenbal:
@annashamray I checked it again and the issue only seems to occur for types that I created using the API. If I create a type via the admin interface, it does appear in the inline

annashamray:
@stevenbal 
Nice catch! Didn't even think it's possible
PR - https://github.com/open-zaak/open-zaak/pull/230

PR #230, opened 2019-11-12
change condition to represent inline objects in templates

annashamray:
Issue - https://github.com/open-zaak/open-zaak/issues/160


annashamray:
@stevenbal thankss, rebasing helped!

Issue #215, opened 2019-11-11
As admin, I want to see the audittrail in the admin for the main objecttypes


joeribekker:
... so I can view its history.

This can replace the "history" button on the detail page for all objecttypes that concern components.

PR #234, opened 2019-11-12
Feature/audittrail in admin

stevenbal:
for https://github.com/open-zaak/open-zaak/issues/215#event-2791569862

### DIFF
--- a/src/openzaak/templates/admin/object_history.html
+++ b/src/openzaak/templates/admin/object_history.html
@@ -15,7 +15,7 @@
             <th scope="col">{% trans 'Resultaat' %}</th>
             <th scope="col">{% trans 'Gebruiker' %}</th>
             <th scope="col">{% trans 'Applicatie' %}</th>
-            <th scope="col">{% trans 'Hoofdobject' %}</th>
+            <th scope="col">{% trans 'Object' %}</th>
             <th scope="col">{% trans 'Toelichting' %}</th>
             <th scope="col">{% trans 'Wijzigingen' %}</th>
         </tr>
@@ -29,7 +29,7 @@
                 <td>{{ audit.resultaat }}</td>
                 <td>{{ audit.gebruikers_weergave }}</td>
                 <td>{{ audit.applicatie_weergave }}</td>
-                <td>{{ audit.hoofd_object }}</td>
+                <td>{{ audit.resource_url }}</td>
                 <td>{{ audit.toelichting }}</td>
                 <td><button class="btn btn-primary" data-toggle="modal" data-target="#myModal-{{forloop.counter}}">{% trans 'Toon wijzigingen' %}</button></td>
             </tr>
--- a/src/openzaak/templates/admin/object_history.html
+++ b/src/openzaak/templates/admin/object_history.html
@@ -1,22 +1,11 @@
-{% extends "admin/base_site.html" %}
+{% extends "admin/object_history.html" %}
 {% load i18n admin_urls %}
 {% load static %}
 
-{% block breadcrumbs %}
-<div class="breadcrumbs">
-<a href="{% url 'admin:index' %}">{% trans 'Home' %}</a>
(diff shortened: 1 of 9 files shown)


src/openzaak/templates/admin/object_history.html
```diff
@@ -0,0 +1,117 @@
+{% extends "admin/base_site.html" %}
```
joeribekker: Extend from `admin/object_history.html` instead of `admin/base.html`

src/openzaak/templates/admin/object_history.html
```diff
@@ -0,0 +1,117 @@
...
+    </table>
+{% endif %}
+</div>
+</div>
+
+<link rel="stylesheet" href="{% static 'vng_api_common/libs/bootstrap/css/bootstrap.min.css' %}">
+<script src="{% static 'vng_api_common/libs/jquery/jquery-3.3.1.slim.min.js' %}"></script>
+<script src="{% static 'vng_api_common/libs/popper/popper.min.js' %}"></script>
+<script src="{% static 'vng_api_common/libs/bootstrap/js/bootstrap.min.js' %}"></script>
+
```
joeribekker: Can you add:

```
{% if object.audittrail and action_list %}
<h1>{% trans "Changes made via the admin." %}
{% endif %}
{{ block.super }}
```

src/openzaak/templates/admin/object_history.html
```diff
@@ -0,0 +1,117 @@
...
+        </tbody>
+    </table>
+{% endif %}
+</div>
+</div>
+
+<link rel="stylesheet" href="{% static 'vng_api_common/libs/bootstrap/css/bootstrap.min.css' %}">
+<script src="{% static 'vng_api_common/libs/jquery/jquery-3.3.1.slim.min.js' %}"></script>
+<script src="{% static 'vng_api_common/libs/popper/popper.min.js' %}"></script>
+<script src="{% static 'vng_api_common/libs/bootstrap/js/bootstrap.min.js' %}"></script>
```
joeribekker: Not sure these scripts and styles are needed?

src/openzaak/templates/admin/object_history.html
```diff
@@ -0,0 +1,117 @@
...
+&rsaquo; <a href="{% url opts|admin_urlname:'change' object.pk|admin_urlquote %}">{{ object|truncatewords:"18" }}</a>
+&rsaquo; {% trans 'History' %}
+</div>
+{% endblock %}
+
+{% block content %}
+<div id="content-main">
+<div class="module">
+
+{% if object.audittrail %}
```
joeribekker: 1. Can you move the if-statement above the `div#content-main` (and move the bottom `endif` accordingly?

2. And, rename the ID `content-main` to `content-main-audittrail` so that it won't conflict with the superblock at the bottom (which should be added by my comment :p)

src/openzaak/templates/admin/object_history.html
```diff
@@ -0,0 +1,117 @@
...
+
+{% block breadcrumbs %}
+<div class="breadcrumbs">
+<a href="{% url 'admin:index' %}">{% trans 'Home' %}</a>
+&rsaquo; <a href="{% url 'admin:app_list' app_label=opts.app_label %}">{{ opts.app_config.verbose_name }}</a>
+&rsaquo; <a href="{% url opts|admin_urlname:'changelist' %}">{{ module_name }}</a>
+&rsaquo; <a href="{% url opts|admin_urlname:'change' object.pk|admin_urlquote %}">{{ object|truncatewords:"18" }}</a>
+&rsaquo; {% trans 'History' %}
+</div>
+{% endblock %}
```
joeribekker: Shouldn't be needed when extend from object_history

src/openzaak/templates/admin/object_history.html
stevenbal: One wasn't needed indeed, but the other three are needed for the modal to show the changes

Issue #169, opened 2019-11-07
As developer, I want to create an audit entry in the API if a change was made in the admin


joeribekker:
... so changes can not be done without leaving an audit trail.

**Description**
Please research if we can make this somewhat generic and discuss if we cannot.

PR #236, opened 2019-11-15
log Admin changes to AuditTrail

annashamray:
issue - https://github.com/open-zaak/open-zaak/issues/169

### DIFF
--- a/src/openzaak/components/besluiten/tests/admin/test_besluit_audittrail.py
+++ b/src/openzaak/components/besluiten/tests/admin/test_besluit_audittrail.py
@@ -3,19 +3,20 @@ import uuid
 from django.test import TestCase
 from django.urls import reverse
 
-from openzaak.components.besluiten.models import Besluit
-from ..factories import BesluitFactory
-from openzaak.components.catalogi.tests.factories import BesluitTypeFactory
 from vng_api_common.audittrails.models import AuditTrail
-from openzaak.utils.tests import AdminTestMixin
-from ..utils import get_operation_url
 
+from openzaak.components.besluiten.models import Besluit
+from openzaak.components.catalogi.tests.factories import BesluitTypeFactory
+from openzaak.utils.tests import AdminTestMixin
+
+from ..factories import BesluitFactory
+from ..utils import get_operation_url
 
 inline_data = {
     "besluitinformatieobject_set-TOTAL_FORMS": 0,
     "besluitinformatieobject_set-INITIAL_FORMS": 0,
     "besluitinformatieobject_set-MIN_NUM_FORMS": 0,
-    "besluitinformatieobject_set-MAX_NUM_FORMS": 1000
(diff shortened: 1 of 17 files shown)


src/openzaak/utils/admin.py
```diff
@@ -118,3 +121,58 @@ class DynamicArrayMixin:
...
+
+    def save_model(self, request, obj, form, change):
+        model = obj.__class__
+        basename = model._meta.object_name.lower()
+
+        viewset = self.get_viewset(request)
+        version, scheme = viewset.determine_version(request, version=api_settings.DEFAULT_VERSION, uuid=obj.uuid)
+        request.version, request.versioning_scheme = version, scheme
+
+        action = 'update' if change else 'create'
```
joeribekker: Isn't this an enum somewhere?

annashamray: yeah, `CommonResourceAction`, I'll change it. It's a bit raw now.

src/openzaak/components/besluiten/tests/admin/test_besluit_audittrail.py
```diff
@@ -0,0 +1,159 @@
...
+        self.assertEqual(audittrail.hoofd_object, f"http://testserver{besluit_url}"),
+        self.assertEqual(audittrail.resource, "besluit"),
+        self.assertEqual(audittrail.resource_url, f"http://testserver{besluit_url}"),
+        self.assertEqual(audittrail.resource_weergave, besluit.unique_representation()),
+
+        old_data, new_data = audittrail.oud, audittrail.nieuw
+        self.assertEqual(old_data["toelichting"], "old")
+        self.assertEqual(new_data["toelichting"], "new")
+
+    def test_delete_besluit_action(self):
```
stevenbal: This means that the AuditTrail remains in the database, even after the main resource (`Zaak` in this case) is deleted right? Where would you be able to see the audittrail for this action, if the resource is already deleted?

@joeribekker @sergei-maertens currently in the audittrail in vng-api-common the audittrail is deleted if the main resource is deleted via the API, is this okay?

joeribekker: > This means that the AuditTrail remains in the database, even after the main resource (Zaak in this case) is deleted right? Where would you be able to see the audittrail for this action, if the resource is already deleted?

This is indeed not correct. The audit trail should be deleted if the main resource is deleted.

> @joeribekker @sergei-maertens currently in the audittrail in vng-api-common the audittrail is deleted if the main resource is deleted via the API, is this okay?

Yes. Delete really means destroy (even if I don't like this from a developer perspective). It's the law :)

joeribekker:
@annashamray Can you make it so the deletion of the main resource will remove the audit trail as well? I'm not sure how this is done via the API but I thought it was just a cascade delete operation... (evidently, its not).

Issue #194, opened 2019-11-07
As maintainer I want to have a clear indicator of the concept status of a zaaktype as well as a button for publishing 


Hugo-ter-Doest:
... so I always know what the concept status is.

**Tasks**

- [ ] Remove concept field from the admin everywhere.
- [ ] Add a publish button on the objects detail page (like where delete is also)
- [ ] Add the string "(CONCEPT)" after a Zaaktype's __str__ method if concept=True

PR #241, opened 2019-11-19
Feature/publishing catalogi admin

stevenbal:
for https://github.com/open-zaak/open-zaak/issues/194

### DIFF
--- a/src/openzaak/components/catalogi/admin/mixins.py
+++ b/src/openzaak/components/catalogi/admin/mixins.py
@@ -1,3 +1,5 @@
+from django.contrib import messages
+from django.http import HttpResponseRedirect
 from django.utils.translation import ugettext_lazy as _
 
 
@@ -16,3 +18,21 @@ class ConceptAdminMixin(object):
     def get_fieldsets(self, request, obj=None):
         fieldsets = super().get_fieldsets(request, obj)
         return tuple(fieldsets) + ((_("Concept"), {"fields": ("concept",)}),)
+
+
+class PublishAdminMixin:
+    def response_post_save_change(self, request, obj):
+        if "_publish" in request.POST:
+            # Clear messages
+            storage = messages.get_messages(request)
+            for i in storage:
+                pass
+
+            obj.concept = False
+            obj.save()
+            msg = _("The resource has been published successfully!")
+            self.message_user(request, msg, level=messages.SUCCESS)
+
+            return HttpResponseRedirect(request.path)
+        else:
+            return super().response_post_save_change(request, obj)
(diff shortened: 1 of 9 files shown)


src/openzaak/templates/admin/submit_line.html
```diff
@@ -0,0 +1,12 @@
+{% extends "admin/submit_line.html" %}
+{% load i18n admin_urls %}
+
+{% block submit-row %}
+    {% if original.concept == True %}
+        <input type="submit" value="{% trans 'Publish' %}" name="_publish"/>
+
+        {% comment %} <a href="{% url 'admin:zaaktype-publish' original.id %}">{% trans 'Publish' %}</a> {% endcomment %}
```
joeribekker: This should not be needed :)

src/openzaak/components/catalogi/models/zaaktype.py
```diff
@@ -358,7 +358,10 @@ class Meta:
         verbose_name_plural = _("Zaaktypen")
 
     def __str__(self):
-        return "{} - {}".format(self.catalogus, self.zaaktype_identificatie)
+        representation = "{} - {}".format(self.catalogus, self.zaaktype_identificatie)
+        if self.concept:
+            representation = "{} (CONCEPT)".format(representation)
+        return representation
```
joeribekker: We also need this for Besluittype and InformatieObjecttype

src/openzaak/components/catalogi/admin/mixins.py
```diff
@@ -16,3 +18,21 @@ class ConceptAdminMixin(object):
     def get_fieldsets(self, request, obj=None):
         fieldsets = super().get_fieldsets(request, obj)
         return tuple(fieldsets) + ((_("Concept"), {"fields": ("concept",)}),)
+
+
+class PublishAdminMixin:
+    def response_post_save_change(self, request, obj):
```
joeribekker: I see that you copied this code for Zaaktype with validation. That's not very DRY.

Why not add a new function like:

```python
def _publish_validation(...):
    pass  # default

def response_post_save_change(self, request, obj):
    if "_publish" in request.POST:
        # ...
        if not self._publish_validation():
            return
        # ...
```

And override `_publish_validation` with the code you added for Zaaktype?

src/openzaak/components/catalogi/admin/zaaktypen.py
```diff
@@ -1,6 +1,6 @@
-from django.contrib import admin
+from django.contrib import admin, messages
```
sergei-maertens: please clean up the unused imports

Issue #196, opened 2019-11-07
AS maintainer of the ZTC I want to have proper version control over my Zaaktypes


Hugo-ter-Doest:
... so my life will be easier.

We need to introduce a better concept for versions in the admin. To accomplish this should:

**Tasks**

- [ ] Change "versiedatum" to readonly in the admin and set it by default to the "beginDatumGeldigheid" when the object is saved (in the admin).
- [ ] Group versiedatum with begin/eind datum fields.
- [ ] Add a button "Create new version" to create a new version of Zaaktype on the detail page.
- [ ] Throw a validation error if there is no `eindeGeldigheid` filled in.
- [ ] A new version will copy ALL information and related objects to NEW objects (where appropriate, m2ms can ofcourse be the same objects) with the following changes:
   - The concept flag should be true for the new object
   - The `beginGeldigheid` should be set to the `eindeGeldigheid` of the original
  - `eindeGeldigheid` should be set to `None`.

PR #247, opened 2019-11-20
Feature/zaaktype add new version in admin

annashamray:
Issue - https://github.com/open-zaak/open-zaak/issues/196

### DIFF
--- a/src/openzaak/components/catalogi/api/validators.py
+++ b/src/openzaak/components/catalogi/api/validators.py
@@ -472,9 +472,9 @@ class GeldigheidVersiedatumValidator:
 
         if datum_begin_geldigheid != versiedatum:
             msg = _(
-                    "De datum_begin_geldigheid moet gelijk zijn aan een "
-                    "Versiedatum van het gerelateerde zaaktype."
-                )
+                "De datum_begin_geldigheid moet gelijk zijn aan een "
+                "Versiedatum van het gerelateerde zaaktype."
+            )
             raise ValidationError(msg, code=self.code)
 
         # current_einde_geldigheid = (
--- a/src/openzaak/components/catalogi/api/validators.py
+++ b/src/openzaak/components/catalogi/api/validators.py
@@ -1,3 +1,5 @@
+from datetime import timedelta
+
 from django.db.models import Q
 from django.utils.translation import ugettext_lazy as _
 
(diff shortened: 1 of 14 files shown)


src/openzaak/components/catalogi/api/validators.py
```diff
@@ -443,3 +445,48 @@ def __call__(self, attrs):
...
+        versiedatum = attrs.get("versiedatum") or self.instance.versiedatum
+
+        if datum_begin_geldigheid != versiedatum:
+            msg = _(
+                "De datum_begin_geldigheid moet gelijk zijn aan een "
+                "Versiedatum van het gerelateerde zaaktype."
+            )
+            raise ValidationError(msg, code=self.code)
+
+        # current_einde_geldigheid = (
```
annashamray: @sergei-maertens @joeribekker 
I've copied logic here from `GeldigheidMixin._clean_geldigheid` because I removed this check from `ZaakType` model
I've commented this part because it seems to contradict the code above:
`versiedatum` cant be equal to `datum_begin_geldigheid` and to `datum_einde_geldigheid + 1` at the same time. 

Is it correct? Should I uncomment it? Should I comment it in the `GeldigheidMixin._clean_geldigheid`?

stevenbal:
@annashamray rebase on master should fix the failing newman tests

src/openzaak/components/catalogi/api/validators.py
```diff
@@ -443,3 +445,48 @@ def __call__(self, attrs):
                             f"Objects can't be updated with a relation to non-concept {field_name}"
                         )
                         raise ValidationError(msg, code=self.code)
+
+
+class GeldigheidVersiedatumValidator:
```
stevenbal: @joeribekker @sergei-maertens @annashamray if the `versiedatum` is read-only in the admin, shouldn't it also be read-only in the API?

joeribekker: @stevenbal Good observation but the answer is no (for now).

We create a UI where we make assumptions about how to fill the `versiedatum`. If this is the defacto standard way of doing things, then I guess it will be changed in the API spec but for now it's fine.

sergei-maertens: only if it's properly described how the versioning machinery works, which isn't the case currently. if the 1.0.0 API spec says writable, then we have to stick to that :thinking:

sergei-maertens:
I'll have to dive into this later today as it's a big PR :)

src/openzaak/components/catalogi/api/validators.py
sergei-maertens: I think that commented code can be removed. It is possible that version x exists from `[a,b]` and version y exists from `[c,d]`, with a gap between `b` and `c`. In practice, that means that between b & c no cases of that type can be created.

sergei-maertens:
I suspect that the newman tests failing is because of the validators in the `ZaaktypeSerializer`

annashamray:
> I suspect that the newman tests failing is because of the validators in the `ZaaktypeSerializer`

@sergei-maertens I removed `GeldigheidVersiedatumValidator` as it was discussed

Issue #245, opened 2019-11-20
As developer, I want to create an audit entry in the API if a change was made in the admin inlines...


sergei-maertens:
so changes cannot be made without leaving an audit trail.

**Description**

This is the sequel to #169 - during [code review](https://github.com/open-zaak/open-zaak/pull/236#pullrequestreview-319768617) it became apparent that `Zaak.status` inlines etc. aren't caught by the audit trail integration.

The implementation looks solid and it should be fairly straight-forward to add it for inlines.

annashamray:
Implemented in bea5f97b24eb5b493f83bfb94cb8c1f0cd605b93

PR #251, opened 2019-11-21
Feature/audittrail admin inline

annashamray:
Issue - https://github.com/open-zaak/open-zaak/issues/245

### DIFF
--- a/src/openzaak/components/documenten/admin.py
+++ b/src/openzaak/components/documenten/admin.py
@@ -19,7 +19,9 @@ class GebruiksrechtenInline(AuditTrailInlineAdminMixin, admin.TabularInline):
     viewset = viewsets.GebruiksrechtenViewSet
 
 
-class EnkelvoudigInformatieObjectInline(AuditTrailInlineAdminMixin, admin.StackedInline):
+class EnkelvoudigInformatieObjectInline(
+    AuditTrailInlineAdminMixin, admin.StackedInline
+):
     model = EnkelvoudigInformatieObject
     extra = 1
     viewset = viewsets.EnkelvoudigInformatieObjectViewSet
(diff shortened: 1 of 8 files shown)


src/openzaak/components/documenten/admin.py
```diff
@@ -40,6 +46,9 @@ def get_not_lock_display(self, obj) -> bool:
     get_not_lock_display.short_description = "free to change"
     get_not_lock_display.boolean = True
 
+    def get_viewset(self, request):
```
joeribekker: Why is this here?

src/openzaak/utils/admin.py
```diff
@@ -223,9 +227,75 @@ def delete_model(self, request, obj):
...
+        # add new
+        for obj in formset.new_objects:
+            data_after = self.get_serializer_data(request, viewset, obj)
+            self.trail(
+                obj, viewset, request, CommonResourceAction.create, None, data_after
+            )
+
+
+class AuditTrailInlineAdminMixin(object):
+    viewset = None
```
joeribekker: Maybe we should add a check if viewset is still None on init and raise an ImproperlyConfigured exception to prevent accidental left outs.

src/openzaak/components/documenten/admin.py
annashamray: `EnkelvoudigInformatieObjectCanonical` doesn't have matching viewset, so the changes in the admin can't be logged. But in the admin page it has inline objects - `EnkelvoudigInformatieObject` and `Gebruiksrechten` which have related viewsets with AuditTrail. So we can log only changes in the inline objects here (it's a special case).

If we don't override `get_viewset` method the exception will be raised since this class doesn't have `viewset` attribute.

src/openzaak/utils/admin.py
annashamray: In this case we couldn't trail changes in objects with no related viewset. Like EnkelvoudigInformatieObjectCanonical

Issue #189, opened 2019-11-07
As a municipality I want to export/import catalogues


Hugo-ter-Doest:
... so I can transfer catalogues from accept to production environment, and share catalogues with other municipalities.

**Description**

- [x] Export Catalog (complete with everything)
- [x] Import Catalog (as a new Catalog)

**Implementation hints**
We're going to use the API serializers to create an export per resource. We'll use a dependency graph between resources to do the import within a single database transaction.

**Tasks**

- [x] Create Django command to: Export resource using the API serializer to a file named after the resource (ie. zaaktypen.json)
- [x] Extend command to: Group all exported JSON files into a ZIP-file.
- [x] Create a dependency graph to read the files in the correct order (ie. 1. catalogi.json, 2. zaaktypen.json, 3 statussen.json, etc.).
- [x] Create Django command to: Import ZIP-file to create all objects.
- [x] Throw error when catalog (the unique field) already exists
- [x] Add export button to Catalog detail Admin UI
- [x] Add import button to Catalog list Admin UI

NOTE: UUIDs are not imported and moved this to another US: #249
NOTE: The import/export of Zaaktypes is moved to a different US: #214

sergei-maertens:
@joeribekker do you want to export the uuid here as well or generate a new one? Exporting them along would result in duplicates across environments, which shouldn't happen, so I'd say that `UUID` must be generated at import time.

joeribekker:
@sergei-maertens good question. I think it should export with uuid.

On import it should ask whether to generate new ones or not. I can imagine that migrating a catalog from one instance to another should not change the uuids. We should indicate when you should say yes or no.

joeribekker:
@sergei-maertens @stevenbal let's discuss at earliest convenience.

sergei-maertens:
yeah we need to discuss, tomorrow/Friday is fine

joeribekker:
Updated original User Story with more details after our discussion.

We decided to use the API to export data, for the following reasons:

- Can be done in a database transaction
- Format adheres to the API format and should port more easily to other suppliers
- No database information is leaked (database agnostic)

Issue #256, opened 2019-11-22
ResultaatType.brondatumArchiefprocedure fields are incorrectly serialized for empty values


stevenbal:
**Describe the bug**
The gegevensgroep `ResultaatType.brondatumArchiefprocedure` has validation which states that certain fields should be empty strings dependent on the value of `afleidingswijze` (see [standard](https://github.com/VNG-Realisatie/gemma-zaken/blob/master/docs/_content/standaard/catalogi/index.md#ztc-003)).

So when I create a `ResultaatType` with the following valid data:
```json
{
	"zaaktype": "{{zaaktype_url}}",
	"omschrijving": "Klaar",
	"resultaattypeomschrijving": "https://ref.tst.vng.cloud/referentielijsten/api/v1/resultaattypeomschrijvingen/e6a0c939-3404-45b0-88e3-76c94fb80ea7",
	"selectielijstklasse": "{{selectielijstklasse_url}}",
	"brondatumArchiefprocedure": {
		"afleidingswijze": "gerelateerde_zaak",
		"procestermijn": null,
		"datumkenmerk": "",
		"einddatumBekend": false,
		"objecttype": "",
		"registratie": ""
	}
}
```

I get the following response:
```json
{
    "url": "http://localhost:8000/catalogi/api/v1/resultaattypen/716f8cf5-b093-41e9-af1f-b77764fa54e9",
    "zaaktype": "http://localhost:8000/catalogi/api/v1/zaaktypen/14d0d28b-4b10-44cd-8fbb-c65b758ef8f4",
    "omschrijving": "Klaar",
    "resultaattypeomschrijving": "https://ref.tst.vng.cloud/referentielijsten/api/v1/resultaattypeomschrijvingen/e6a0c939-3404-45b0-88e3-76c94fb80ea7",
    "omschrijvingGeneriek": "Afgewezen",
    "selectielijstklasse": "https://referentielijsten-api.vng.cloud/api/v1/resultaten/3c35d5f5-8348-4ff3-b23a-18e9d6fd6a1a",
    "toelichting": "",
    "archiefnominatie": "vernietigen",
    "archiefactietermijn": "P1827D",
    "brondatumArchiefprocedure": {
        "afleidingswijze": "gerelateerde_zaak",
        "datumkenmerk": null,
        "einddatumBekend": false,
        "objecttype": null,
        "registratie": null,
        "procestermijn": null
    }
}
```

[...]

PR #259, opened 2019-11-22
Feature/import export catalogi

stevenbal:
fixes https://github.com/open-zaak/open-zaak/issues/189

exporting ResultaatType doesn't work yet, due to issues with the serializer: https://github.com/open-zaak/open-zaak/issues/256


src/openzaak/components/catalogi/tests/admin/test_import_export_catalogus.py
```diff
@@ -0,0 +1,140 @@
...
+        form = response.form
+        with open(self.filename, "rb") as f:
+            form["file"] = (
+                self.filename,
+                f.read(),
+            )
+            response = form.submit("_import")
+
+        self.assertIn(
+            "validation error occurred while deserializing a Catalogus", response.text
```
joeribekker: This can fail if we translate this string.

Use the exact same translatable string as you did when raising the exception.

src/openzaak/templates/admin/change_list.html
```diff
@@ -0,0 +1,24 @@
...
+  {% comment %} {% if has_add_permission %}
+  {<li>
+    {% url cl.opts|admin_urlname:'add' as add_url %}
+    <a href="{% add_preserved_filters add_url is_popup to_field %}" class="addlink">
+      {% blocktrans with cl.opts.verbose_name as name %}Add {{ name }}{% endblocktrans %}
+    </a>
+  </li
+  {% endif %} {% endcomment %}
+  {{ block.super }}
+    {% if cl.opts.verbose_name == "catalogus" %}
```
joeribekker: Since this code is only for the catalogus admin, I prefer making a new (copy) template file and point to it from the admin class (something like `template_name=...`) instead of using an if-statement.

src/openzaak/templates/admin/submit_line.html
```diff
@@ -6,5 +6,9 @@
         <input type="submit" value="{% trans 'Publish' %}" name="_publish"/>
     {% endif %}
 
+    {% if original.domein %}
+        <input type="submit" value="{% trans 'Export' %}" name="_export"/>
```
joeribekker: This becomes messy now with all the actions and I definitly don't like the if-statement nor the attribute checking. But you can leave it for now and we'll see how to make it more elegant afterwards. I made #268 for this.

src/openzaak/templates/import_catalogus.html
```diff
@@ -0,0 +1,13 @@
+{% extends "admin/base_site.html" %}
```
joeribekker: Can you move this file to template/admin/catalogi/?

src/openzaak/components/catalogi/management/commands/export.py
```diff
@@ -0,0 +1,69 @@
...
+
+            setattr(request, "versioning_scheme", URLPathVersioning())
+            setattr(request, "version", "1")
+
+            for instance in objects:
+                serialized = serializer(instance, context={"request": request})
+                data = serialized.data
+
+                if resource == "BesluitType":
+                    data["zaaktypen"] = []
```
joeribekker: I remember we spoke of this, but why is this here again?

I think a code-comment should explain why this is here.

src/openzaak/components/catalogi/admin/catalogus.py
```diff
@@ -52,6 +64,94 @@ class CatalogusAdmin(ListObjectActionsAdminMixin, admin.ModelAdmin):
...
+                id_list.append([",".join(str(id) for id in ids)])
+
+        return resource_list, id_list
+
+    def response_post_save_change(self, request, obj):
+        if "_export" in request.POST:
+            # Clear messages
+            storage = messages.get_messages(request)
+            for i in storage:
+                pass
```
joeribekker: Can you check how @annashamray did this for creating a new version?

I can't remember this was needed there, but then again, it might just show 2 messages...heh.

src/openzaak/components/catalogi/management/commands/export.py
stevenbal: This is needed, because when exporting, `BesluitType.zaaktypen` and `ZaakType.besluittypen` both contain references, but then creating the `BesluitType` first will fail, because the `ZaakType` does not exist yet. I'll add a comment for it

src/openzaak/components/catalogi/api/serializers/relatieklassen.py
```diff
@@ -9,7 +9,7 @@
 from ..validators import ZaakInformatieObjectTypeCatalogusValidator
 
 
-class ZaakTypeInformatieObjectTypeSerializer(serializers.HyperlinkedModelSerializer):
+class ZaakInformatieobjectTypeSerializer(serializers.HyperlinkedModelSerializer):
```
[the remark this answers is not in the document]
stevenbal: Yes, but the model is named `ZaakInformatieobjectType` and I use the model name to retrieve the serializer. I suppose I could also rename the model from `ZaakInformatieobjectType` -> `ZaakTypeInformatieObjectType`.

src/openzaak/components/catalogi/admin/catalogus.py
stevenbal: Yes, if I remove it, it will show 2 messages: the custom one I added and the original message that says that the resource was successfully edited or whatever

src/openzaak/components/catalogi/api/serializers/relatieklassen.py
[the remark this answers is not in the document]
joeribekker: Please do.

joeribekker:
Thanks, please do the rename as suggested here: https://github.com/open-zaak/open-zaak/pull/259#discussion_r350858787

src/openzaak/components/catalogi/admin/mixins.py
```diff
@@ -122,3 +133,101 @@ def response_change(self, request, obj):
...
+                    "pk", flat=True
+                )
+            )
+
+        resource_list = []
+        id_list = []
+        for resource, ids in resources.items():
+            if ids:
+                resource_list.append([resource])
+                id_list.append([",".join(str(id) for id in ids)])
```
sergei-maertens: can't you pass an actual python list instead of turning it into a comma-separated list? This is passed to the management command args right?

src/openzaak/components/catalogi/management/commands/export.py
```diff
@@ -0,0 +1,73 @@
...
+            action="append",
+            nargs="*",
+            help=_("Name of the resource to export objects for"),
+        )
+        parser.add_argument(
+            "--ids",
+            help=_("IDs of objects to be exported for the resource"),
+            action="append",
+            nargs="*",
+            type=str,
```
sergei-maertens: I'd expect `type=int` because they are PKs?

src/openzaak/components/catalogi/management/commands/export.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def handle(self, *args, **options):
+        archive_name = options.pop("archive_name")
+        resources = options.pop("resource")
+        ids = options.pop("ids")
+
+        filenames = []
+
+        for resource, id_list in zip(resources, ids):
+            resource = resource[0]
```
sergei-maertens: that's kinda weird?

src/openzaak/components/catalogi/management/commands/export.py
```diff
@@ -0,0 +1,73 @@
...
+            serializer = getattr(serializers, f"{resource}Serializer")
+            objects = model.objects.filter(id__in=parsed_ids)
+
+            factory = APIRequestFactory()
+            request = factory.get("/")
+
+            setattr(request, "versioning_scheme", URLPathVersioning())
+            setattr(request, "version", "1")
+
+            for instance in objects:
```
sergei-maertens: can't this be replaced by a `serializer(instance=objects, many=True)` call?

src/openzaak/components/catalogi/management/commands/import.py
```diff
@@ -0,0 +1,73 @@
...
+    "ZaakInformatieobjectType",
+    "ResultaatType",
+    "RolType",
+    "StatusType",
+    "Eigenschap",
+]
+
+
+class Command(BaseCommand):
+    help = _("Import Catalogi data from a .zip file")
```
sergei-maertens:
```suggestion
    help = "Import Catalogi data from a .zip file"
```

src/openzaak/components/catalogi/management/commands/import.py
```diff
@@ -0,0 +1,73 @@
...
+                    data = zip_file.read(f"{resource}.json").decode()
+                    for old, new in uuid_mapping.items():
+                        data = data.replace(old, new)
+
+                    data = json.loads(data)
+
+                    model = apps.get_model("catalogi", resource)
+                    serializer = getattr(serializers, f"{resource}Serializer")
+
+                    factory = APIRequestFactory()
```
sergei-maertens: stuff like this can be outside of the for-loop, right?

src/openzaak/components/catalogi/tests/admin/test_import_export_catalogus.py
```diff
@@ -0,0 +1,147 @@
...
+        )
+        response = form.submit("_import")
+
+        self.assertIn(
+            _("A validation error occurred while deserializing a Catalogus"),
+            response.text,
+        )
+        self.assertEqual(Catalogus.objects.count(), 1)
+
+    def tearDown(self):
```
sergei-maertens: please put these methods at the top of the class

also - it's better to not define `def tearDown`, but add it to the cleanup in `def setUp`:

```python
self.addCleanup(lambda: os.remove(self.filename))
```

Cleanup always runs, even if errors happen. tearDown doesn't always run

src/openzaak/components/catalogi/tests/management/test_import_export.py
```diff
@@ -0,0 +1,274 @@
...
+
+class ExportCatalogiTests(TestCase):
+    def setUp(self):
+        self.filepath = os.path.join(PATH, "export_test.zip")
+
+    def test_export_catalogus(self):
+        catalogus = CatalogusFactory.create()
+
+        call_command(
+            "export", self.filepath, resource=[["Catalogus"]], ids=[[str(catalogus.id)]]
```
sergei-maertens: the arguments are weird, should be fixed in the command itself I think and then adapt the calling code

```suggestion
            "export", self.filepath, resource=["Catalogus"], ids=[[catalogus.id]]
```

Command should probably also validate that there are equal amounts of elements in `resource` as in `ids`

src/openzaak/templates/admin/submit_line.html
joeribekker: Since #268 was merged, this branch will have merge conflicts with this change.

You can copy (in master branch) `templates/admin/catalogi/change_form_zaaktype.html` to `change_form_catalog.html` and use the same template logic to add this button without the if-statement.

Then use this template by specifying `change_form_template = "admin/catalogi/change_form_catalog.html"` in the 
`CatalogusAdmin` class.

stevenbal:
@sergei-maertens I have changed it a bit, now when importing a zip, the file is not saved and the content is passed to the import management command. Similarly for exporting, I now pass the response to the export command and write the output to that response, instead of writing it to a file, then sending that file and then deleting that file

src/openzaak/components/catalogi/management/commands/export.py
```diff
@@ -0,0 +1,96 @@
...
+            nargs="*",
+            type=int,
+        )
+
+    def handle(self, *args, **options):
+        archive_name = options.pop("archive_name")
+        response = options.pop("response")
+
+        if response and archive_name:
+            raise CommandError(
```
sergei-maertens: for a future exercise, but just informing you - with argparse you can define mutually exclusive options. https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_mutually_exclusive_group

src/openzaak/components/catalogi/openapi.yaml
```diff
@@ -5887,7 +5887,7 @@ paths:
         format: uuid
   /zaaktype-informatieobjecttypen:
     get:
-      operationId: zaakinformatieobjecttype_list
+      operationId: zaaktypeinformatieobjecttype_list
```
sergei-maertens: that's a breaking change - clients operating on `operationId` (of which gemma-zds-clients is one) will crash on this change.

The schema introspection has a `get_operation_id` method, probably we should override it to keep the old one here.

Issue #257, opened 2019-11-22
As developer, I want to add the missing archive-actions


joeribekker:
... so all archive actions actually work.

**Description**
Some archive actions are now missing in the reference implementation. We need to fill these in to make it work.

- [ ] Find out which actions are missing
- [ ] Add missing actions
- [ ] Enable tests for these missing actions in the test suite.

sergei-maertens:
CTRL + F for `raise NotImplementedError`

stevenbal:
PR for documentation of this archiving behavior in standaard https://github.com/VNG-Realisatie/gemma-zaken/pull/1479

PR #264, opened 2019-11-25
Feature/zaak archiving

stevenbal:
fixes https://github.com/open-zaak/open-zaak/issues/257

I will also update the newman tests for zaak archiving

### DIFF
--- a/src/openzaak/components/zaken/brondatum.py
+++ b/src/openzaak/components/zaken/brondatum.py
@@ -1,7 +1,7 @@
 from datetime import date, datetime
 from typing import Union
 
-from django.db.models import Max
+from django.db.models import Max, Min
 from django.utils.translation import ugettext_lazy as _
 
 from dateutil.relativedelta import relativedelta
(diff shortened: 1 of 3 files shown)


src/openzaak/components/zaken/brondatum.py
```diff
@@ -1,7 +1,7 @@
 from datetime import date, datetime
 from typing import Union
 
-from django.db.models import Max
+from django.db.models import Max, Min
```
sergei-maertens: unused `Min` import?

src/openzaak/components/zaken/brondatum.py
```diff
@@ -187,7 +196,16 @@ def get_brondatum(
...
-        # TODO: Relation from Zaak to Besluit is not implemented yet...
-        raise NotImplementedError
+        zaakbesluiten = zaak.besluit_set.all()
+        if not zaakbesluiten.exists():
+            # Cannot use vervaldatum_besluit if Zaak has no Besluiten
+            raise DetermineProcessEndDateException(
+                _("Geen besluiten aan zaak gekoppeld om brondatum uit af te leiden.")
+            )
+
+        max_vervaldatum = zaakbesluiten.aggregate(Max("vervaldatum"))[
```
sergei-maertens: Can you add a test where there _does_ exist a `ZaakBesluit`, BUT the `vervaldatum` is `None`? That should return a validation error saying the vervaldatum must be set first or something like that?

@joeribekker sanity check that I'm not talking nonsense?

joeribekker: It can be `None` so thats a valid case.

Issue #219, opened 2019-11-11
Support external zaak.zaaktype


sergei-maertens:
Part of #6 - it should be possible to create a zaak of a zaaktype in an external catalogus.

PR #267, opened 2019-11-25
change Zaak.zaaktype field to loose-fk

annashamray:
Depends on https://github.com/maykinmedia/django-loose-fk/pull/13
Issue - https://github.com/open-zaak/open-zaak/issues/219


sergei-maertens:
Loose FK 0.5.8 is released

src/openzaak/components/zaken/api/serializers/zaken.py
```diff
@@ -254,6 +245,17 @@ class Meta:
...
+            "zaaktype": {
+                "lookup_field": "uuid",
+                "max_length": 1000,
+                "min_length": 1,
+                "validators": [
+                    LooseFkResourceValidator("ZaakType", settings.ZTC_API_SPEC),
+                    LooseFkIsImmutableValidator(),
+                    PublishValidator(),
+                ],
+                "help_text": get_help_text("zaken.Zaak", "zaaktype"),
```
sergei-maertens: this sould already happen by default, I think this line can be removed

src/openzaak/components/zaken/tests/test_zaken.py
```diff
@@ -574,3 +577,131 @@ def test_zaak_archiefactiedatum_afleidingswijze_ingangsdatum_besluit(self):
...
+@tag("external-urls")
+class ZaakCreateExternalURLsTests(JWTAuthMixin, APITestCase):
+    heeft_alle_autorisaties = True
+    list_url = get_operation_url("zaak_create")
+
+    def test_create_external_zaaktype(self):
+        catalogus = "https://externe.catalogus.nl/api/v1/catalogussen/1c8e36be-338c-4c07-ac5e-1adf55bec04a"
+        zaaktype = "https://externe.catalogus.nl/api/v1/zaaktypen/b71f72ef-198d-44d8-af64-ae1932df830a"
+
+        with requests_mock.Mocker(real_http=True) as m:
```
sergei-maertens: why the `real_http`? I'd prefer the test to error if a URL is fetched that accesses the real network. Those calls should be mocked.

src/openzaak/components/zaken/tests/test_zaken.py
```diff
@@ -574,3 +577,131 @@ def test_zaak_archiefactiedatum_afleidingswijze_ingangsdatum_besluit(self):
...
+            m.register_uri(
+                "GET", zaaktype, json=get_zaaktype_response(catalogus, zaaktype),
+            )
+            m.register_uri(
+                "GET",
+                catalogus,
+                json={
+                    "url": catalogus,
```
sergei-maertens: probably better to make a `get_catalogus_response` as well?

src/openzaak/components/zaken/tests/test_zaken.py
```diff
@@ -574,3 +577,131 @@ def test_zaak_archiefactiedatum_afleidingswijze_ingangsdatum_besluit(self):
         self.assertEqual(
             zaak.archiefactiedatum, date(2030, 5, 3)  # 2020-05-03 + 10 years
         )
+
+
+@tag("external-urls")
+class ZaakCreateExternalURLsTests(JWTAuthMixin, APITestCase):
```
sergei-maertens: can you add a test with an external zaaktype that is not published (`concept=True`)?

src/openzaak/components/zaken/api/serializers/zaken.py
annashamray: `help_text` is removed

src/openzaak/components/zaken/tests/test_zaken.py
annashamray: In the `LooseFkResourceValidator` `ZTC_API_SPEC` is requested. I thought we use it in tests without mocking it. And the tests for `Besluit.besluittype` also don't mock it.

src/openzaak/components/zaken/tests/test_zaken.py
annashamray: `get_catalogus_response` is added to the utils

src/openzaak/components/zaken/tests/test_zaken.py
annashamray: testcase is added

src/openzaak/components/zaken/tests/test_zaken.py
sergei-maertens: ah yeah... now I remember that we have some calls going through.

I'll create a new issue so we can leave it as it is. API spec mocking is annoying

PR #273, opened 2019-11-27
Add NLX inway

sergei-maertens:
Added while doing performance testing with/without NLX


deployment/files/traefik/deployment.yml
```diff
@@ -28,10 +28,9 @@ spec:
               cpu: "250m"
             limits:
               memory: "200Mi"
-              cpu: "500m"
+              cpu: "2"
```
sergei-maertens: noticed that Traefik was hitting the limit, so increased it to use a full node. CPU usage is high during TLS termination.

deployment/roles/nlx/tasks/ingress.yml
```diff
@@ -0,0 +1,21 @@
+---
+
+- name: Install HTTPS Ingress Route
```
sergei-maertens: I need to remove this again, we don't want to let traffic go through Traefik before hitting the inway. Inway is exposed by a LoadBalancer service already.

deployment/roles/nlx/templates/inway-svc.yml.j2
```diff
@@ -0,0 +1,20 @@
...
+kind: Service
+apiVersion: v1
+metadata:
+  labels:
+    app.kubernetes.io/name: inway
+    app.kubernetes.io/component: inway
+    app.kubernetes.io/part-of: nlx
+    app.kubernetes.io/managed-by: Ansible
+spec:
+  type: LoadBalancer
```
sergei-maertens: this gives us/you a public IP address -> DNS must be configured to this

deployment/roles/openzaak/templates/deployment.yml.j2
```diff
@@ -74,7 +74,7 @@ spec:
           - name: DJANGO_SETTINGS_MODULE
             value: openzaak.conf.docker
           - name: ALLOWED_HOSTS
-            value: "localhost,{{ domain }}"
+            value: "localhost,nginx.openzaak-test,{{ domain }}"
```
sergei-maertens: nasty - inway could be adapted to pass a HTTP_HOST header, or we can set it in nginx config to pass the canonical Host header.

deployment/roles/nlx/defaults/main.yml
```diff
@@ -0,0 +1,63 @@
+---
+
+# Kubernetes namespace
+nlx_namespace: nlx
+nlx_instance: proeftuin
```
joeribekker: proeftuin?

deployment/roles/nlx/defaults/main.yml
```diff
@@ -0,0 +1,63 @@
...
+# nlx_inway_services:
+#   - name: zrc
+#     endpoint_url: http://gemma-zrc.k8s.dc1.proeftuin.utrecht.nl/api/v1
+#     documentation_url = "http://gemma-zrc.k8s.dc1.proeftuin.utrecht.nl"
+#     authorization_model = "none"
+#     api_specification_document_url = "http://gemma-zrc.k8s.dc1.proeftuin.utrecht.nl/api/v1/schema/openapi.json"
+
+nlx_inway_replicas: 1
+nlx_inway_directory_registration_address: directory-registration-api.demo.nlx.io:443
+nlx_inway_self_address: inway.utrechtproeftuin.nl
```
joeribekker: Should be inway.openzaak.nl?

deployment/roles/nlx/templates/inway.yml.j2
```diff
@@ -0,0 +1,84 @@
...
+              value: "{{ nlx_inway_self_address }}:443"
+
+            - name: TLS_NLX_ROOT_CERT
+              value: "/certs/root.crt"
+            - name: TLS_ORG_CERT
+              value: "/certs/org.crt"
+            - name: TLS_ORG_KEY
+              value: "/certs/org.key"
+
+            # TODO: enable logdb
```
joeribekker: I guess we need this or NLX looses half its value?

deployment/roles/openzaak/templates/deployment.yml.j2
joeribekker: > we can set it in nginx config to pass the canonical Host header

This sounds like a reasonable solution.

deployment/roles/nlx/defaults/main.yml
sergei-maertens: 3 guesses where this role was originally created :-)

deployment/roles/nlx/defaults/main.yml
sergei-maertens: yeah these are defaults anyway, so I'll change it to a more generic case like `inway.gemeente.nl`

deployment/roles/nlx/templates/inway.yml.j2
sergei-maertens: yep, but that requires initializing the database and for the proof-of-concept/getting things up and running, this is not as relevant yet

joeribekker: Created #279 for this

PR #274, opened 2019-11-27
Inline API credentials with applications

sergei-maertens:
Refs. #4 - manage credentials inline with the application

Uses the same models under the hood, only the admin does some clever stuff to keep things in sync.


src/openzaak/components/authorizations/admin.py
```diff
@@ -1,5 +1,49 @@
...
+    formset = BaseModelFormSet
+    fields = ("identifier", "secret")
+    extra = 1
+
+    # Disable system checks, since this model is not related at all to Applicatie
+    def check(self, *args, **kwargs):
+        return []
+
+    def get_formset(self, request, obj=None, **kwargs):
+        return CredentialsFormSet
```
joeribekker: Is there a reason why this is here instead of just assigned to the class formset attribute?

sergei-maertens:
Yeah, the class attribute gets fed to inlineformset_factory, while we
already provide the formset class to use (result of modelformset_factory)

Issue #218, opened 2019-11-11
Support external EIO.informatieobjecttype


sergei-maertens:
Part of #6 - it should be possible to create documents from a type in an external catalogus.

annashamray:
implemented in https://github.com/open-zaak/open-zaak/commit/6f6f4f515898fd569cea797a9146ecf377480f7b

PR #275, opened 2019-11-27
Feature/loose fk eio informatietype

annashamray:
Issue - https://github.com/open-zaak/open-zaak/issues/218


src/openzaak/components/documenten/query.py
```diff
@@ -76,7 +76,7 @@ def filter_for_authorizations(
             )
             vertrouwelijkheidaanduiding_whens.append(
                 When(
-                    **{"informatieobjecttype": informatieobjecttype},
+                    **{"_informatieobjecttype": informatieobjecttype},
```
annashamray: @joeribekker @sergei-maertens 
This change is just workaround. After adding all loose-fk fields the authorization should be modified.
But this can be tricky since in the `AuthorizationsFilterMixin` we directly construct DB queries

sergei-maertens: @annashamray can you create a follow up issue to validate permissions with remote types?

src/openzaak/components/documenten/tests/test_validation.py
```diff
@@ -14,6 +14,7 @@
 from .factories import EnkelvoudigInformatieObjectFactory
 
 
+@override_settings(ALLOWED_HOSTS=["testserver"])
```
sergei-maertens: FYI - it can't hurt, but the django test runner inserts `testserver` into ALLOWED_HOSTS already :-)

src/openzaak/components/zaken/api/validators.py
```diff
@@ -141,13 +141,11 @@ def __call__(self, attrs):
...
-            io_type_id = informatieobject.enkelvoudiginformatieobject_set.values_list(
-                "informatieobjecttype_id", flat=True
-            ).first()
+            io_type = informatieobject.latest_version.informatieobjecttype
         else:
-            io_type_id = informatieobject.informatieobjecttype.id
+            io_type = informatieobject.informatieobjecttype
 
-        if not zaak.zaaktype.informatieobjecttypen.filter(id=io_type_id).exists():
+        if not zaak.zaaktype.informatieobjecttypen.filter(uuid=io_type.uuid).exists():
```
sergei-maertens: this is probably going to break if/when you have both external zaaktype and external IOT. Something to be solved in django-loose-fk

src/openzaak/components/documenten/tests/test_validation.py
annashamray: Yes, but I suppose it also keeps wildcards, i.e. all the urls in loose-fk fields are processed as locals
So this test doesn't work without `override_settings`

sergei-maertens: Yep, correct. Because of the `dev` settings having the wildcard. The system check should've complained about that, no?

src/openzaak/components/zaken/api/validators.py
sergei-maertens: Created #283 for this

src/openzaak/components/documenten/query.py
annashamray: Issue - https://github.com/open-zaak/open-zaak/issues/284

src/openzaak/components/documenten/tests/test_validation.py
annashamray: Yes, but it's just a warning message :)

Issue #287, opened 2019-11-29
ZaakEigenschap.eigenschap should be in Zaak.zaaktype.eigenschappen


stevenbal:
**Describe the bug**
According to https://github.com/VNG-Realisatie/gemma-zaken/blob/master/docs/_content/standaard/zaken/index.md#zrc-018, when adding a `ZaakEigenschap` to a `Zaak`, it should be validated that `ZaakEigenschap.eigenschap` is in `Zaak.zaaktype.eigenschappen`, if not, the API should return 400

issue in gemma-zaken https://github.com/VNG-Realisatie/gemma-zaken/issues/1476

PR #288, opened 2019-11-29
Feature/zaakeigenschap validation

stevenbal:
fixes https://github.com/open-zaak/open-zaak/issues/287


stevenbal:
I have also updated the newman tests and it works locally, when this PR is approved I will update the testsuite on the API test platform (which will also add the tests for archiving that were missing)

Issue #283, opened 2019-11-28
Validation of IO/Zaak with remote IOT/ZT will probably crash


sergei-maertens:
See https://github.com/open-zaak/open-zaak/pull/275#discussion_r351881898

PR #293, opened 2019-12-02
Feature/loose fk zaaktype informatieobjecttype validator

annashamray:
issue - https://github.com/open-zaak/open-zaak/issues/283

### DIFF
--- a/src/openzaak/utils/validators.py
+++ b/src/openzaak/utils/validators.py
@@ -159,10 +159,5 @@ class ObjecttypeInformatieobjecttypeRelationValidator:
             objecttype_data = AuthorizedRequestsLoader.fetch_object(
                 objecttype_url, do_underscoreize=False
             )
-            related_iotypes = [
-                iotype
-                for iotype in objecttype_data.get("informatieobjecttypen", [])
-                if iotype == iotype_url
-            ]
-            if not related_iotypes:
+            if iotype_url not in objecttype_data.get("informatieobjecttypen", []):
                 raise serializers.ValidationError(message, code=code)
--- a/src/openzaak/utils/validators.py
+++ b/src/openzaak/utils/validators.py
@@ -6,6 +6,7 @@ from django_loose_fk.drf import FKOrURLField, FKOrURLValidator
 from rest_framework import serializers
 from vng_api_common.oas import fetcher, obj_has_shape
 from vng_api_common.validators import IsImmutableValidator
+
 from openzaak.components.documenten.models import EnkelvoudigInformatieObject
 
 from ..loaders import AuthorizedRequestsLoader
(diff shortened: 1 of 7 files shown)


src/openzaak/utils/validators.py
```diff
@@ -115,3 +117,52 @@ def __call__(self, value: str):
...
+        objecttype = getattr(object, self.objecttype_field)
+
+        if not isinstance(informatieobject, EnkelvoudigInformatieObject):
+            io_type = informatieobject.latest_version.informatieobjecttype
+        else:
+            io_type = informatieobject.informatieobjecttype
+
+        # zaaktype/besluittype and informatieobjecttype should be both internal or external
+        if bool(objecttype.pk) != bool(io_type.pk):
+            raise serializers.ValidationError(message, code=code)
```
stevenbal: Maybe I'm nitpicking, but wouldn't it be a bit better to have a different error message here that states that both types should be either internal or external, instead of just returning the general message `The informatieobjecttype is not related to the zaaktype of the zaak.`?

src/openzaak/utils/validators.py
```diff
@@ -115,3 +117,52 @@ def __call__(self, value: str):
...
+            iotype_url = io_type._loose_fk_data["url"]
+            objecttype_data = AuthorizedRequestsLoader.fetch_object(
+                objecttype_url, do_underscoreize=False
+            )
+            related_iotypes = [
+                iotype
+                for iotype in objecttype_data.get("informatieobjecttypen", [])
+                if iotype == iotype_url
+            ]
+            if not related_iotypes:
```
stevenbal: Can't you replace this and the list comprehension above with something simpler like:
```python
if not iotype_url in objecttype_data.get("informatieobjecttypen", []):
    raise serializers.ValidationError(message, code=code)
```

annashamray: Done. Thank you. Now it's much neater

stevenbal: :+1:

src/openzaak/utils/validators.py
annashamray: Honestly I don't know what to write in the message in this case to make it clear from the client's point of view. Maybe something like this:
`The informatieobjecttype and the zaaktype should belong to the same component` ?

stevenbal: Maybe something like `The informatieobjecttype and the zaaktype should be hosted in the same component`?

annashamray: Yeah, this version is better. The message is changed

sergei-maertens: I've changed the error message into specifying that both types must belong to the same catalogue, I think that's the semantic meaning that is expressed behind this, and that requirement is specified in the catalogue API as well.

Issue #237, opened 2019-11-18
Creating ZaakType in the admin without trefwoorden and/or verantwoordingsrelatie gives internal server error


stevenbal:
**Describe the bug**
In the `ZaakType` admin, the fields `trefwoorden` and `verantwoordingsrelatie` are not marked as required, but when trying to create a `ZaakType` without either of these two fields, the page crashes, because the fields have not null constraints:

```html
IntegrityError at /admin/catalogi/zaaktype/add/

null value in column "trefwoorden" violates not-null constraint
DETAIL:  Failing row contains (180, 2019-11-18, null, t, 612c8cd0-bdaf-43c1-92c3-95845443ea93, 1, dasd, , openbaar, sadasd, sadasd, , extern, sadasd, sdadas, asda, 10 days, null, f, f, null, null, f, , null, 2019-11-18, {https://ref.tst.vng.cloud/standaard/}, https://referentielijsten-api.vng.cloud/api/v1/procestypen/e1b73..., s, , 2).
```

**Expected behavior**
If these fields are indeed required, they should also be marked as required in the admin, raising a ValidationError when they are not filled in

sergei-maertens:
Was fixed via https://github.com/open-zaak/open-zaak/commit/5e26ca6c614431ffcb4dd26456ac1a92edda7c18 already

PR #297, opened 2019-12-03
Issue/zaaktype required fields admin

stevenbal:
fixes https://github.com/open-zaak/open-zaak/issues/237

### DIFF
--- a/src/openzaak/components/catalogi/admin/forms.py
+++ b/src/openzaak/components/catalogi/admin/forms.py
@@ -38,9 +38,12 @@ class ZaakTypeForm(forms.ModelForm):
         self.fields[
             "trefwoorden"
         ].help_text += " Gebruik een komma om waarden van elkaar te onderscheiden."
+        self.fields["trefwoorden"].required = True
+
         self.fields[
             "verantwoordingsrelatie"
         ].help_text += " Gebruik een komma om waarden van elkaar te onderscheiden."
+        self.fields["verantwoordingsrelatie"].required = True
 
     def clean(self):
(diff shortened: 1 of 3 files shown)


annashamray:
I thought that this problem with default `ArrayField` values was solved in https://github.com/open-zaak/open-zaak/commit/5e26ca6c614431ffcb4dd26456ac1a92edda7c18
I can't reproduce the issue now on the master branch

src/openzaak/components/catalogi/admin/forms.py
```diff
@@ -38,9 +38,12 @@ def __init__(self, *args, **kwargs):
         self.fields[
             "trefwoorden"
         ].help_text += " Gebruik een komma om waarden van elkaar te onderscheiden."
+        self.fields["trefwoorden"].required = True
```
sergei-maertens: Hrm no, this is not intended behaviour - making fields required will only result in people filling in nonsensical data to get rid of the error message.

I also think the above help text is no longer relevant, will verify!

src/openzaak/components/catalogi/admin/forms.py
```diff
@@ -38,9 +38,12 @@ def __init__(self, *args, **kwargs):
...
+
         self.fields[
             "verantwoordingsrelatie"
         ].help_text += " Gebruik een komma om waarden van elkaar te onderscheiden."
+        self.fields["verantwoordingsrelatie"].required = True
```
sergei-maertens: same remarks as for `trefwoorden`

Issue #262, opened 2019-11-25
Informatieobject attribute description and format are missing


joeribekker:
**Describe the bug**
The `informatieobject` attribute on various locations have no description or format defined. The official spec does have this.

**Besluiten**
```diff
--- "src/openzaak/components/besluiten/openapi.official.yaml"	2019-11-25 11:09:46.246230000 +0100
+++ "src/openzaak/components/besluiten/openapi.yaml"	2019-11-25 11:08:56.850645800 +0100
@@ -1398,12 +1341,10 @@
           format: uri
       - name: informatieobject
         in: query
-        description: URL-referentie naar het INFORMATIEOBJECT (in de Documenten API)
-          waarin (een deel van) het besluit beschreven is.
+        description: ''
         required: false
         schema:
           type: string
-          format: uri
       responses:
         '200':
           description: OK
```

**Zaken**
```diff
--- "src/openzaak/components/zaken/openapi.official.yaml"	2019-11-25 11:09:47.578509600 +0100
+++ "src/openzaak/components/zaken/openapi.yaml"	2019-11-25 11:09:41.484947800 +0100
@@ -2713,12 +2649,10 @@
           format: uri
       - name: informatieobject
         in: query
-        description: URL-referentie naar het INFORMATIEOBJECT (in de Documenten API),
-          waar ook de relatieinformatie opgevraagd kan worden.
+        description: ''
         required: false
         schema:
           type: string
-          format: uri
       responses:
         '200':
           description: OK
```

PR #300, opened 2019-12-03
Fix API specs

annashamray:
Issue - https://github.com/open-zaak/open-zaak/issues/262 (description part)

& took the opportunity to dutchify the authorizations URL (-> autorisaties)


annashamray:
`GenerateSchemaTestCase` was added to git by mistake. It was only for local use and debugging and `doesn't` assert anything
@sergei-maertens thanks for renaming it, but I think it's better to remove it. Don't you mind it?

sergei-maertens:
@annashamray doesn't hurt to have a test that ensures at least it doesn't error :)

sergei-maertens:
After dependency update, verified that the schema stays the same as in this PR originally :-)

Issue #284, opened 2019-11-28
support for external urls in permissions and ListFilterByAuthorizationsMixin


annashamray:
1. Add support for external urls in permissions and in `ListFilterByAuthorizationsMixin`
2. Write authorization tests for external `zaaktype`, `besluittype` and `informatieobjecttype`

sergei-maertens:
Fixed via #305

PR #305, opened 2019-12-04
Feature/loose fk authorizations

annashamray:
Issue - https://github.com/open-zaak/open-zaak/issues/284

### DIFF
--- a/src/openzaak/utils/query.py
+++ b/src/openzaak/utils/query.py
@@ -44,23 +44,43 @@ class LooseFkAuthorizationsFilterMixin:
     vertrouwelijkheidaanduiding_use = True
     authorizations_lookup = None
 
+    @property
+    def prefix(self):
+        return (
+            "" if not self.authorizations_lookup else f"{self.authorizations_lookup}__"
+        )
+
     def get_loose_fk_object(
         self, authorization, local=True
     ) -> Union[models.Model, str]:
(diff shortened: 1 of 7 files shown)


src/openzaak/utils/query.py
```diff
@@ -23,3 +36,113 @@ def delete(self, *args, **kwargs):
...
+
+        else:
+            del filters["_va_order__lte"]
+            queryset = self.filter(**filters)
+        return queryset
+
+    def get_filters(self, scope, authorizations, local=True) -> dict:
+        prefix = self.prefix
+        loose_fk_field = (
+            f"_{self.loose_fk_field}" if local else f"_{self.loose_fk_field}_url"
```
sergei-maertens: Works and is a pragmatic approach, but as soon as we change the naming scheme, this breaks. Something to keep in mind :)

sergei-maertens:
@annashamray can you take a look at the failing tests and rebase on `master` - the `authorizations` -> `autorisaties` rename is done in there and I'm not sure if it affects this PR or not.

Newman tests may be a matter of setting `ALLOWED_HOSTS` correctly :)

annashamray:
@sergei-maertens I am looking at it rn. Only one test is failing, so I don't think it's a problem with `ALLOWED_HOSTS`. We'll see

Issue #214, opened 2019-11-11
As a municipality I want to export/import zaaktypes


joeribekker:
... so I can transfer zaaktypes from accept to production environment, and share zaaktypes with other municipalities.

The export of a Zaaktype should include all underlying (Statustype, etc.) AND related objects (BesluitType, InformatieObjectType).

The import is more complex: We need a small step before we can do the actual import:

1. Import all data into a temporary cache/table
2. Show all imported BesluitType and InformatieObjectTypes
3. Next to each BesluitType and InformatieObjectType, show a list of all existing BesluitTypes and existing InformatieObjectType. If you select one, it will not create this object, but uses the existing one. If you select nothing, the object is created.
4. Make all newly created objects have `concept=True`.

**Description**

- [ ] Export (individual) Zaaktype
- [ ] Import Zaaktype (in existing Catalog)
- [ ] Show intermediate screen to link BesluitType en InformatieObjectType to existing objects
- [ ] Perform actual import

joeribekker:
@Hugo-ter-Doest Deze even losgetrokken van #189 want het heeft wat voeten in de aarde. Kan jij aangeven wat de prio van deze is?

joeribekker:
Dit is niet goed automated gegaan. #306 had geen betrekking op dit ticket.

PR #306, opened 2019-12-04
Set up a way to compare relative deltas - fixes catalogus export bug

sergei-maertens:
... dealing with time is stupid.


joeribekker:
I don't think this fixes #214 which is about exporting zaaktypes specifically but this fixes a bug when exporting catalogs which was done in #189.

sergei-maertens:
I doubt that, because then you have to start transforming integer amount of working days to calendar days (hi holidays!), and correctly transform incoming precisions like +1 month to the correct number of working days :-)

Issue #296, opened 2019-12-02
As a developer I want to support external urls in ZaakBesluit


annashamray:
Now ZaakBesluit is a Proxy model based on `Besluit` model. Adding support of external urls in Besluit.zaak we also need to:

1. Add creating ZaakBesluit in the external API if Besluit.zaak was external
2. (not sure about this one though). If `besluiten` component is extenral and `zaken` is internal when we need `ZaakBesluit` to be a real endpoint to receive `create` and `destroy` actions

annashamray:
Implemented in https://github.com/open-zaak/open-zaak/pull/312

PR #312, opened 2019-12-05
Feature/loose fk zaakbesluit

annashamray:
Issue - https://github.com/open-zaak/open-zaak/issues/296


src/openzaak/components/besluiten/models.py
```diff
@@ -157,6 +157,29 @@ class Besluit(AuditTrailMixin, APIMixin, models.Model):
...
+        _("externe previous zaak"), blank=True, max_length=1000,
+    )
+    _previous_zaak = models.ForeignKey(
+        "zaken.Zaak",
+        on_delete=models.PROTECT,
+        null=True,
+        blank=True,  # een besluit kan niet bij een zaak horen
+        related_name="previous_zaak",
+    )
+    previous_zaak = FkOrURLField(
```
sergei-maertens: not sure why you're adding these as actual model fields?

src/openzaak/components/besluiten/models.py
```diff
@@ -170,6 +193,17 @@ def __str__(self):
     def save(self, *args, **kwargs):
         if not self.identificatie:
             self.identificatie = generate_unique_identification(self, "datum")
+
+        # save previous zaak for triggers
+        if self.pk:
```
sergei-maertens: doesn't it make more sense to override `__init__` and store `self._previous_zaak = self.zaak`? You can optimize this by keeping the loose-fk underlying attributes and define a property

```py
@property
def previous_zaak(self):
    ...
```

to handle it, but I don't think keeping that information in sync in the DB will go well. I'm worried that will give all weird kind of problems to debug :thinking:

src/openzaak/components/zaken/models/zaken.py
```diff
@@ -842,3 +848,57 @@ def save(self, *args, **kwargs):
...
+
+    class Meta:
+        verbose_name = _("zaakbesluit")
+        verbose_name_plural = _("zaakbesluiten")
+        unique_together = ("zaak", "_besluit")
+        constraints = [
+            models.UniqueConstraint(
+                fields=["zaak", "_besluit_url"],
+                condition=~models.Q(_besluit_url=""),
+                name="unique_zaak_and_besluit_document",
```
sergei-maertens:
```suggestion
                name="unique_zaak_and_besluit",
```

src/openzaak/components/besluiten/models.py
annashamray: `previous_zaak` fields are removed from db

src/openzaak/components/besluiten/models.py
annashamray: the realization is changed to use `__init__` and `@property` instead of model fields

Issue #310, opened 2019-12-05
As sysadmin, I want to see when notifications from Open Zaak to the NRC are failing


joeribekker:
... so I can act on this

**Description**
Currently, if a notification from OpenZaak to the NRC fails it does so silently. This needs to be logged in an admin-accessible interface.

We can hook into the logger (that only warns now) and create entries in the database in addition.

PR #327, opened 2019-12-10
Feature/failed notification admin

stevenbal:
issue https://github.com/open-zaak/open-zaak/issues/310


src/openzaak/components/zaken/api/viewsets.py
```diff
@@ -429,7 +434,9 @@ class ZaakObjectViewSet(
 
 
 class ZaakInformatieObjectViewSet(
+    FailedNotificationMixin,
     NotificationCreateMixin,
+    NotificationDestroyMixin,
```
stevenbal: @sergei-maertens this one was missing here, but not on the viewset for `BesluitInformatieObject`, so I added it here as well

src/openzaak/notificaties/mixins.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+class FailedNotificationMixin:
+    def notify(self, status_code, data, instance=None):
+        stream = StringIO()
+        console = logging.StreamHandler(stream=stream)
+        logger = logging.getLogger("vng_api_common.notifications.viewsets")
+        logger.addHandler(console)
+
+        logger.addFilter(FailedNotificationFilter())
```
joeribekker: This doesn't seem like the right approach to me.

Why not create a log handler that puts stuff in a model? And register it via settings?

This whole mixin concept seems inappropriate.

Let's discuss with @sergei-maertens

sergei-maertens: Indeed, I had something like https://pypi.org/project/django-db-logger/ in mind instead

stevenbal:
closed in favor of https://github.com/open-zaak/open-zaak/pull/354

Issue #329, opened 2019-12-13
Exporting Catalogus/ZaakType in docker gives internal server error


stevenbal:
**Describe the bug**
When I try to export a `Catalogus` or a `ZaakType` in the admin and I am running Openzaak in docker, I get the following error:

```html
DisallowedHost at /admin/catalogi/catalogus/56/change/

Invalid HTTP_HOST header: 'testserver'. You may need to add 'testserver' to ALLOWED_HOSTS.
```

PR #330, opened 2019-12-13
Issue/import export request host

stevenbal:
fixes https://github.com/open-zaak/open-zaak/issues/329

### DIFF
--- a/src/openzaak/components/catalogi/management/commands/export.py
+++ b/src/openzaak/components/catalogi/management/commands/export.py
@@ -3,6 +3,7 @@ import json
 import zipfile
 
 from django.apps import apps
+from django.conf import settings
 from django.core.management.base import BaseCommand, CommandError
 from django.http import HttpResponse
 from django.utils.translation import ugettext_lazy as _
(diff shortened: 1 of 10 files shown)


sergei-maertens:
Yeah relying on the sites framework is the best bet, we already do so in the `sync` calls anyway.

src/openzaak/components/catalogi/management/commands/export.py
```diff
@@ -57,7 +58,10 @@ def handle(self, *args, **options):
             )
 
         factory = APIRequestFactory()
-        request = factory.get("/")
+        server_name = (
```
sergei-maertens: For completeness sake, use `Site.objects.get_current()` here instead

src/openzaak/components/catalogi/tests/admin/test_import_export_catalogus.py
```diff
@@ -350,3 +350,42 @@ def test_import_catalogus_already_exists(self):
...
+            rsin="000000000",
+            domein="TEST",
+            contactpersoon_beheer_naam="bla",
+            contactpersoon_beheer_telefoonnummer="0612345678",
+            contactpersoon_beheer_emailadres="test@test.nl",
+        )
+
+        url = reverse("admin:catalogi_catalogus_change", args=(catalogus.pk,))
+
+        response = self.app.get(url)
```
sergei-maertens: I wouldn't do this as a `webtest`, but rather just use `call_command(...)` instead.

Issue #261, opened 2019-11-25
The `zaak` attribute on Besluit differs slightly


joeribekker:
**Describe the bug**
The official OAS spec indicates the `zaak` attribute on Besluit is not nullable and has no minimum length.

**Besluiten**
```diff
--- "src/openzaak/components/besluiten/openapi.official.yaml"	2019-11-25 11:09:46.246230000 +0100
+++ "src/openzaak/components/besluiten/openapi.yaml"	2019-11-25 11:08:56.850645800 +0100
@@ -2048,6 +1989,8 @@
           type: string
           format: uri
           maxLength: 200
+          minLength: 1
+          nullable: true
         datum:
           title: Datum
           description: De beslisdatum (AWB) van het besluit.
```

annashamray:
implemented in https://github.com/open-zaak/open-zaak/commit/14668b270e08a1af600b78ce03a2e498c55aa1cb

PR #331, opened 2019-12-17
Issue/besluit zaak schema diff

annashamray:
issue - https://github.com/open-zaak/open-zaak/issues/261


annashamray:
> Should we port this to https://github.com/maykinmedia/django-loose-fk/tree/develop/django_loose_fk/inspectors?

I'm not sure. The only thing this inspector does - getting rid of `nullable` in the schema. I thought this is specific `openzaak` issue, not `django-loose-fk`

Issue #322, opened 2019-12-10
As a developer I want to support external urls in Status


annashamray:
**Description**
1. `Status.statustype` field in `zaken` component should be loose-fk

annashamray:
Implemented in 2b480160ebc77b379dfeaa3b48e0096476515e6c

PR #334, opened 2019-12-17
Feature/loose fk statustype

annashamray:
Issue - https://github.com/open-zaak/open-zaak/issues/322

Depends on:
* https://github.com/maykinmedia/django-loose-fk/pull/17
* https://github.com/open-zaak/open-zaak/pull/336


sergei-maertens:
#336 was merged - can you rebase so the diff is smaller?

Issue #323, opened 2019-12-10
As a developer I want to support external urls in Resultaat


annashamray:
**Description**
1. `Resultaat.resultaattype` field in `zaken` component should be `loose-fk`

annashamray:
implemented in b1351bff57365057abd60b249d106577c96c804b

PR #336, opened 2019-12-18
Feature/loose fk resultaattype

annashamray:
issue - https://github.com/open-zaak/open-zaak/issues/323

Depends on https://github.com/maykinmedia/django-loose-fk/pull/17


src/openzaak/components/zaken/api/serializers/zaken.py
```diff
@@ -727,16 +727,6 @@ def create(self, validated_data):
 
 
 class ResultaatSerializer(serializers.HyperlinkedModelSerializer):
```
stevenbal: Will `validators = [CorrectZaaktypeValidator("resultaattype")]` still work with external FKs? I would also expect tests for the zaaktype match validation with both external and internal zaaktype/resultaattype.

annashamray: The validator should work after https://github.com/maykinmedia/django-loose-fk/pull/17 is included into `django-loose-fk`
The test with successful validation for external 'resultaattype` - https://github.com/open-zaak/open-zaak/pull/336/files#diff-c095540d689add98e8e09e8376218c7dR25
I didn't include tests for internal `resultaattype` since it was implemented before

Or do you mean tests where the validation fails?

stevenbal: Yes, I meant a test where validation fails (with external zaaktype and external resultaattype, because the test for this with internal zaaktype and resultaattype is probably already implemented)

sergei-maertens: Failing test as described above would be nice to have indeed, other than that it looks good :+1:

annashamray: the test with failing `zaaktype-mismatch` validations is added (https://github.com/open-zaak/open-zaak/pull/336/files#diff-c095540d689add98e8e09e8376218c7dR115)

Issue #325, opened 2019-12-10
As a developer I want to support external urls in ZaakEigenschap


annashamray:
**Description**
`ZaakEigenschap.eigenschap` field in zaken component should be `loose-fk`

PR #339, opened 2019-12-20
Feature/loose fk roltype

annashamray:
Issue - https://github.com/open-zaak/open-zaak/issues/324

Depends on https://github.com/open-zaak/open-zaak/pull/336

PR #342, opened 2019-12-20
Feature/loose fk eigenschap

annashamray:
Issue - https://github.com/open-zaak/open-zaak/issues/325


annashamray:
the branch is rebased on https://github.com/open-zaak/open-zaak/pull/339

Issue #321, opened 2019-12-10
As a developer I want to support external urls in ObjectInformatieObject


annashamray:
**Description**
1. `ObjectInformatieObject.object` can be external
2. When destroying `ZIO` or `BIO` object with external `Informatieobject` the `delete` request to the external DRC should be made

PR #345, opened 2019-12-23
Feature/loose fk oio

annashamray:
Issue - https://github.com/open-zaak/open-zaak/issues/321

Depends on https://github.com/VNG-Realisatie/vng-api-common/pull/119


src/openzaak/components/besluiten/models.py
```diff
@@ -250,6 +250,11 @@ class BesluitInformatieObject(models.Model):
             "API) waarin (een deel van) het besluit beschreven is.",
         ),
     )
+    _objectinformatieobject_url = models.URLField(
+        blank=True,
+        max_length=1000,
+        help_text="URL of related IbjectInformatieObject object in the other API",
```
sergei-maertens: Small typo & should be marked as translatable

```suggestion
        help_text=_("URL of related ObjectInformatieObject object in the other API"),
```

src/openzaak/components/documenten/api/filters.py
```diff
@@ -48,26 +45,13 @@ class Meta:
...
-
-
-class ObjectFilter(URLModelChoiceFilter):
+class ObjectFilter(FkOrUrlFieldFilter):
     def filter(self, qs, value):
-        if isinstance(value, Zaak):
-            self.field_name = "zaak"
-        elif isinstance(value, Besluit):
+        path = urlparse(value).path
+        if "/besluiten/" in path:
```
sergei-maertens: Don't like these hard coded URLs. Can't you use `vng_api_common.utils.get_viewset_for_path` instead for this? You would then be able to figure out the model from `viewset.queryset.model` and map that to the relevant field.

annashamray: But if the path to the external url would differ from openzaak url patterns it couldn't be resolved
For example, the path: "https://externe.zaken.nl/api/v1/zaken/{uuid}" can't be resolved with `get_viewset_for_path` because the url lacks subpath `zaken`.

sergei-maertens: Right, I missed that. Can you add that as a comment? I'd also do stricter validation then - `{whatever}/besluiten/{uuid}` should match as pattern

src/openzaak/components/besluiten/migrations/0014_besluitinformatieobject__objectinformatieobject_url.py
```diff
@@ -0,0 +1,22 @@
...
+        ("besluiten", "0013_besluit__zaakbesluit_url"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="besluitinformatieobject",
+            name="_objectinformatieobject_url",
+            field=models.URLField(
+                blank=True,
+                help_text="URL of related IbjectInformatieObject object in the other API",
```
stevenbal:
```suggestion
                help_text="URL of related ObjectInformatieObject object in the other API",
```

This typo doesn't appear in models.py though

annashamray: Thanks! I probably fixed it in the models.py after `makemigrations` was used

src/openzaak/components/documenten/api/filters.py
annashamray: a comment and stricter validation are added

PR #347, opened 2019-12-24
Specify viewset queryset ordering

sergei-maertens:
Ordering is relevant for pagination & it silences the warnings in the test output.

Dev settings are modified to hard-error when you forget to specify the ordering on viewset querysets.

### DIFF
--- a/src/openzaak/components/zaken/api/viewsets.py
+++ b/src/openzaak/components/zaken/api/viewsets.py
@@ -326,7 +326,7 @@ class StatusViewSet(
 
     """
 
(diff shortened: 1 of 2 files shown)


src/openzaak/components/zaken/api/viewsets.py
```diff
@@ -725,7 +727,9 @@ class ResultaatViewSet(
 
     """
 
-    queryset = Resultaat.objects.select_related("_resultaattype", "zaak").all()
+    queryset = Resultaat.objects.select_related("_resultaattype", "zaak").order_by(
+        "-pk"
```
annashamray: ugh, black

PR #354, opened 2020-01-02
Feature/failed notifications

stevenbal:
I used django-db-logger as Sergei suggested https://github.com/open-zaak/open-zaak/pull/327#pullrequestreview-333327042

issue https://github.com/open-zaak/open-zaak/issues/310
requires https://github.com/VNG-Realisatie/vng-api-common/pull/120 to be merged and bumped first


src/openzaak/components/besluiten/tests/test_besluit_create.py
```diff
@@ -114,6 +114,7 @@ def test_us162_voeg_besluit_toe_aan_zaak(self):
                 besluit.besluitinformatieobject_set.get().informatieobject, io.canonical
             )
 
+    @override_settings(ALLOWED_HOSTS=["openzaak.nl", "testserver"])
```
sergei-maertens: not sure what exactly makes these changes required?

stevenbal: for some reason I was getting 400's due to disallowed host when testing locally, I'm not sure why that was happening though

sergei-maertens: Can you try with a local `.env` file?

```bash
(open-zaak) ➜  open-zaak git:(master) cat .env
DEBUG=1
ALLOWED_HOSTS=localhost,127.0.0.1,testserver,openzaak.nl,example.com
DB_NAME=utrecht
```

stevenbal: thanks, that fixes it

PR #379, opened 2020-01-07
Add support for ADFS login to the admin

sergei-maertens:
Going to abstract the majority of the code into a third party package, as it's originally built for https://github.com/isprojects/


joeribekker:
Looks good overall. I guess indeed this needs to be a package, so we can also use it in Open Notifications.

sergei-maertens:
Yep, and also beheerconsole and whatever end-user applications come into existence :-)

sergei-maertens:
Replaced the implementation with django-auth-adfs-db, next week I can test this out at Utrecht, so until then, please do not merge yet!

sergei-maertens:
@stevenbal I think the envvars for Newman tests aren't set for external PRs, making the build fail. Can we discuss this tomorrow on what the best approach is?

sergei-maertens:
@joeribekker tested at Utrecht and confirmed to be functional. Rebased on master, this is ready for check in imo.

PR #382, opened 2020-01-08
[WIP] Config for Gemeente Buren installation

sergei-maertens:
Currently keeping a branch/PR to track/share the changes, but this will be moved to its own repository.


joeribekker:
Any recommendation on what to do with this @sergei-maertens ?

I think the Docker deploy stuff for Open Notificaties should be in that repo.

sergei-maertens:
@joeribekker yes it should be, just keeping this PR open so I don't forget about it and so I can copy over the code for it.

Even better, the ansible roles for Open-Zaak/Open-Notificaties should be in their own repositories completely, then we can point Ansible Galaxy at those repos. It's similar to publishing a package to PyPI.

sergei-maertens:
Not going to merge this, the various functionalities have been distributed into different projects.

PR #387, opened 2020-01-09
Make Newman/Docker stages conditional

sergei-maertens:
Via #379 it became clear that forks (untrusted) do not have access
to the necessary environment variables to:
* run Newman tests
* publish docker images

This should disable the relevant stages for forks, keeping the Django
tests and code style checks stage. Newman tests and docker builds/push
will happen once the branch is merged into master.

Note that this introduces a small regression risk - changes could be
introduced that break the API tests from the ATP and we'd only
realize _after_ the merge to master.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,6 +41,12 @@ before_script:
   # https://github.com/travis-ci/travis-ci/issues/8982#issuecomment-354357640
   - python -c "import fcntl; fcntl.fcntl(1, fcntl.F_SETFL, 0)"
 
+stages:
+  - name: Newman tests
+    if: fork = false
+  - name: Push Docker image(s)
+    if: fork = false
+
 jobs:
   include:
     # Run jobs in parallel in stage Tests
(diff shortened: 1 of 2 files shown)


sergei-maertens:
@stevenbal the problem is that those secrets now become public to the world, as anyone can fork and make a PR that prints out the values to the Travis log.

stevenbal:
but they are only used inside the docker used by travis, so they can just be dummy variables right? If they can't be public, they shouldn't be in https://github.com/open-zaak/open-zaak/tree/master/fixtures either

.travis.yml
```diff
@@ -96,6 +96,7 @@ jobs:
         - docker-compose rm -f
 
     - stage: Push Docker image(s)
+      if: fork = false
```
stevenbal: @sergei-maertens I moved the conditional here, not sure if it works, but if it is declared before the jobs, it will push the docker image before running all the tests (which is not desirable I assume?)

sergei-maertens: I have an external PR I can rebase on this branch, we'll see in a moment :-)

sergei-maertens: seems to work as intended, can you approve this PR so I can merge it?

PR #390, opened 2020-01-09
Issue/zaaktype publish permissiondenied

stevenbal:
fixes https://github.com/open-zaak/open-zaak/issues/385

### DIFF
--- a/src/openzaak/utils/permissions.py
+++ b/src/openzaak/utils/permissions.py
@@ -46,6 +46,13 @@ class AuthRequired(permissions.BasePermission):
         return getattr(obj, permission_main_object)
 
     def has_permission(self, request: Request, view) -> bool:
+        # permission checks run before the handler is determined. if there is no handler,
+        # a "method is not allowed" must be raised, not an HTTP 403 (see #385)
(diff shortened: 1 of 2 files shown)


src/openzaak/utils/permissions.py
```diff
@@ -46,6 +47,9 @@ def get_main_object(self, obj, permission_main_object):
         return getattr(obj, permission_main_object)
 
     def has_permission(self, request: Request, view) -> bool:
+        if not view.action:
+            raise MethodNotAllowed(request.method)
```
sergei-maertens: This is a bit bold - currently we do use `ViewSet` almost everywhere, but this will break for bare `APIView` views.

edit: I'm changing the implementation based on DRFs view code in `rest_framework.views.APIView.dispatch`

```python
            self.initial(request, *args, **kwargs)

            # Get the appropriate handler method
            if request.method.lower() in self.http_method_names:
                handler = getattr(self, request.method.lower(),
                                  self.http_method_not_allowed)
            else:
                handler = self.http_method_not_allowed
```

Issue #357, opened 2020-01-02
Authorizations should support external zaaktypen


joeribekker:
**Describe the bug**
Authorizations should support external zaaktypen

See: https://github.com/open-zaak/open-zaak/pull/307/files#diff-f40996d24e342971921c649c3de886a1R378

PR #427, opened 2020-01-20
Feature/authorization external zaaktype

stevenbal:
issue https://github.com/open-zaak/open-zaak/issues/357

### DIFF
--- a/src/openzaak/components/autorisaties/admin.py
+++ b/src/openzaak/components/autorisaties/admin.py
@@ -1,4 +1,5 @@
 from django.contrib import admin
+from django.core.exceptions import ObjectDoesNotExist
 from django.forms import BaseModelFormSet
 from django.shortcuts import redirect
 from django.urls import path, reverse
@@ -50,13 +51,20 @@ class AutorisatieInline(admin.TabularInline):
                 "<strong>Maximale vertrouwelijkheidaanduiding</strong>: "
                 "{va}"
             )
-            zaaktype = get_related_object(obj)
+            try:
+                zaaktype = get_related_object(obj)
+                admin_url = reverse(
+                    "admin:catalogi_zaaktype_change", kwargs={"object_id": zaaktype.pk}
+                )
+                zt_repr = str(zaaktype)
+            except ObjectDoesNotExist:
+                admin_url = obj.informatieobjecttype
+                zt_repr = f"{obj.zaaktype} (EXTERN)"
+
             return format_html(
                 template,
-                admin_url=reverse(
-                    "admin:catalogi_zaaktype_change", kwargs={"object_id": zaaktype.pk}
-                ),
-                zt_repr=str(zaaktype),
+                admin_url=admin_url,
+                zt_repr=zt_repr,
                 va=obj.get_max_vertrouwelijkheidaanduiding_display(),
             )
 
@@ -68,14 +76,21 @@ class AutorisatieInline(admin.TabularInline):
                 "<strong>Maximale vertrouwelijkheidaanduiding</strong>: "
                 "{va}"
             )
-            informatieobjecttype = get_related_object(obj)
-            return format_html(
-                template,
-                admin_url=reverse(
+            try:
+                informatieobjecttype = get_related_object(obj)
+                admin_url = reverse(
                     "admin:catalogi_informatieobjecttype_change",
                     kwargs={"object_id": informatieobjecttype.pk},
-                ),
-                iot_repr=str(informatieobjecttype),
+                )
+                iot_repr = str(informatieobjecttype)
+            except ObjectDoesNotExist:
+                admin_url = obj.informatieobjecttype
+                iot_repr = f"{obj.informatieobjecttype} (EXTERN)"
+
+            return format_html(
+                template,
+                admin_url=admin_url,
+                iot_repr=iot_repr,
                 va=obj.get_max_vertrouwelijkheidaanduiding_display(),
             )
 
@@ -84,15 +99,18 @@ class AutorisatieInline(admin.TabularInline):
                 "<strong>Besluittype</strong>: "
                 '<a href="{admin_url}" target="_blank" rel="noopener">{bt_repr}</a>'
             )
-            besluittype = get_related_object(obj)
-            return format_html(
-                template,
-                admin_url=reverse(
+            try:
+                besluittype = get_related_object(obj)
+                admin_url = reverse(
                     "admin:catalogi_besluittype_change",
                     kwargs={"object_id": besluittype.pk},
-                ),
-                bt_repr=str(besluittype),
-            )
+                )
+                bt_repr = str(besluittype)
+            except ObjectDoesNotExist:
+                admin_url = obj.informatieobjecttype
+                bt_repr = f"{obj.besluittype} (EXTERN)"
+
+            return format_html(template, admin_url=admin_url, bt_repr=bt_repr,)
 
         return ""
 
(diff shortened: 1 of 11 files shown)


src/openzaak/sass/admin/components/_autorisatie-form.scss
```diff
@@ -58,4 +58,16 @@ $color-row-background: #f9f9f9 !default;
   & &__va-selection {
   }
 
+  div.external-type ul {
```
sergei-maertens: All the styling so far is done with BEM and the sass operator `&` - please refactor to use proper CSS classnames.

Don't target HTML elements, only target class names please.

src/openzaak/components/autorisaties/admin.py
```diff
@@ -50,13 +51,20 @@ def _get_extra(self, obj) -> str:
                 "<strong>Maximale vertrouwelijkheidaanduiding</strong>: "
                 "{va}"
             )
-            zaaktype = get_related_object(obj)
+            try:
```
sergei-maertens: This try-except is actually wrong if you happen to have UUIDs that are the same, as I just experienced during testing.

I have a catalogue that was imported with the same UUIDs (somehow), but the full URLs are obviously different. So while I'm configuring access on a zaaktype from https://catalogi.vng.cloud/api/v1/, the admin overview is linking me to the internal zaaktype.

This resolution should not be tried, instead, it should use the "check if local object based on URL" mechanism. It's implemented in `django_loose_fk.loaders.BaseLoader.is_local_url`

src/openzaak/components/autorisaties/admin.py
```diff
@@ -50,13 +51,20 @@ def _get_extra(self, obj) -> str:
...
+                zaaktype = get_related_object(obj)
+                admin_url = reverse(
+                    "admin:catalogi_zaaktype_change", kwargs={"object_id": zaaktype.pk}
+                )
+                zt_repr = str(zaaktype)
+            except ObjectDoesNotExist:
+                admin_url = obj.informatieobjecttype
```
sergei-maertens:
```suggestion
                admin_url = obj.zaaktype
```
?

src/openzaak/components/autorisaties/admin.py
```diff
@@ -68,14 +76,21 @@ def _get_extra(self, obj) -> str:
                 "<strong>Maximale vertrouwelijkheidaanduiding</strong>: "
                 "{va}"
             )
-            informatieobjecttype = get_related_object(obj)
-            return format_html(
-                template,
-                admin_url=reverse(
+            try:
```
sergei-maertens: Same remark w/r to local/remote URLs and wrong resolution

src/openzaak/components/autorisaties/admin.py
```diff
@@ -84,15 +99,18 @@ def _get_extra(self, obj) -> str:
                 "<strong>Besluittype</strong>: "
                 '<a href="{admin_url}" target="_blank" rel="noopener">{bt_repr}</a>'
             )
-            besluittype = get_related_object(obj)
-            return format_html(
-                template,
-                admin_url=reverse(
+            try:
```
sergei-maertens: Same remark w/r to local/remote URLs and wrong resolution

src/openzaak/components/autorisaties/admin.py
```diff
@@ -84,15 +99,18 @@ def _get_extra(self, obj) -> str:
...
+                admin_url = reverse(
                     "admin:catalogi_besluittype_change",
                     kwargs={"object_id": besluittype.pk},
-                ),
-                bt_repr=str(besluittype),
-            )
+                )
+                bt_repr = str(besluittype)
+            except ObjectDoesNotExist:
+                admin_url = obj.informatieobjecttype
```
sergei-maertens:
```suggestion
                admin_url = obj.besluittype
```

src/openzaak/components/autorisaties/admin_views.py
```diff
@@ -54,27 +55,40 @@ def get_initial_for_component(
...
     spec: Optional[AutorisatieSpec] = None,
 ) -> List[Dict[str, Any]]:
-    if component not in [ComponentTypes.zrc, ComponentTypes.drc, ComponentTypes.brc]:
-        return []
+    _related_objs = {}
+    _related_objs_external = []
+
+    internal_autorisaties = []
+    for autorisatie in autorisaties:
+        try:
```
sergei-maertens: this is also incorrect resolution if it works just on the UUID value :thinking: 

Probably better to use the loose-fk local-url mechanism here too.

src/openzaak/components/autorisaties/admin_views.py
```diff
@@ -54,27 +55,40 @@ def get_initial_for_component(
...
 
+        _initial = {"externe_typen": _related_objs_external}
+        if _related_objs_external:
+            _initial[
+                "related_type_selection"
+            ] = RelatedTypeSelectionMethods.manual_select
+
         grouped_by_va = defaultdict(list)
-        for autorisatie in autorisaties:
+        for autorisatie in internal_autorisaties:
```
sergei-maertens: what with the external autorisaties that have the same VA, shouldn't those end up in the same group?

src/openzaak/components/autorisaties/admin_views.py
```diff
@@ -54,27 +55,40 @@ def get_initial_for_component(
...
             grouped_by_va[autorisatie.max_vertrouwelijkheidaanduiding].append(
                 autorisatie
             )
 
         for va, _autorisaties in grouped_by_va.items():
-            _initial = {"vertrouwelijkheidaanduiding": va}
+            _initial["vertrouwelijkheidaanduiding"] = va
```
sergei-maertens: This - with the de-indented line 111 is also wrong. The idea is that autorisaties are grouped by `(component, scopes, vertrouwelijkheidaanduiding)` and then you get the list of those particular zaaktypen/external zaaktypen.

As soon as an autorisatie changes VA, it belongs to a different group -> a different entry in the `initial` array. Now, everything is put together even though they have different VA. In fact, this overwrites the VA because of the loop here.

You really need to create a new dict here inside of the loop, and inside of the loop it should be append to the `initial` list.

src/openzaak/components/autorisaties/admin_views.py
```diff
@@ -94,21 +108,27 @@ def get_initial_for_component(
...
                 )
-            initial.append(_initial)
+        initial.append(_initial)
 
     elif component == ComponentTypes.drc:
         informatieobjecttype_ids = set(
             InformatieObjectType.objects.values_list("id", flat=True)
         )
 
+        _initial = {"externe_typen": _related_objs_external}
```
sergei-maertens: The same remarks as above apply

src/openzaak/components/autorisaties/admin_views.py
```diff
@@ -128,13 +148,18 @@ def get_initial_for_component(
...
                 )
-            initial.append(_initial)
+        initial.append(_initial)
 
     elif component == ComponentTypes.brc:
         besluittype_ids = set(BesluitType.objects.values_list("id", flat=True))
         relevant_ids = set(related_objs.values())
 
-        _initial = {}
+        _initial = {"externe_typen": _related_objs_external}
```
sergei-maertens: The same remarks as above apply

stevenbal: Besluittypen do not have `vertrouwelijkheidaanduiding`, so this one should be fine right?

sergei-maertens: Yeah, my copy-paste mania got a bit too much :-)

PR #434, opened 2020-01-22
Fixes #392 -- tweak logging setup to be more container friendly

### LINKED ISSUE #392: As sysadmin, I want to store/see/find logging
... so I can easily debug issues.

**Description**
We might want to describe and/or implement logging to a separate volume mount and aggregate them from several Docker containers.

sergei-maertens:
* Made logging setup so that it writes to stdout on Docker environments
* Re-arranged logging documentation
* Expanded on logging documentation
* Added hints on where-to-go-from-here w/r to logging (ELK, Grafana, fluentd...)


src/openzaak/conf/docker.py
```diff
@@ -5,6 +5,8 @@
 os.environ.setdefault("DB_USER", "postgres")
 os.environ.setdefault("DB_PASSWORD", "")
 
+os.environ.setdefault("LOG_STDOUT", "yes")
```
annashamray: In some places the values of `LOG_STDOUT` are inconsistent:
* here: default value = 'yes'
* in `docs/installation/logging.rst` and in `.env-example `: values = 0 or 1
* in django config: default value = false

Issue #375, opened 2020-01-06
As sysadmin, I want to know the NLX configuration for Open Zaak


joeribekker:
... so I can add these configuration parameters to my NLX inway.

**Description**
We need a simple view in OpenZaak that shows the NLX inway configuration along with some documentation on how to connect OpenZaak to NLX. This view can be accessed from the Open Zaak main page.

- [ ] NLX configuration view
- [ ] Documentation where this can be found

**Background**
For now, we decided we are not deploying an NLX inway as part of OpenZaak. This is not our task plus it seems a bit random to do so since as an NLX inway is completely separate from OpenZaak. We also don't have the proper certificate to do this.

joeribekker:
Initial work done here: https://github.com/open-zaak/open-zaak/commits/feature/nlx-config

TODO:

- [ ] Link from admin page somewhere
- [ ] Protect view
- [ ] Make context creation nicer?

PR #437, opened 2020-01-23
Feature/nlx config

annashamray:
Issue - https://github.com/open-zaak/open-zaak/issues/375


src/openzaak/static/libs/prism/prism.css
```diff
@@ -0,0 +1,142 @@
+/* PrismJS 1.19.0
```
sergei-maertens: Including vendored libraries in version control is not something we do anymore. Please add it via [npm](https://www.npmjs.com/package/prismjs)

Sven or I can help you on how to include it into the frontend toolchain.

src/openzaak/static/libs/prism/prism.js
```diff
@@ -0,0 +1,5 @@
+/* PrismJS 1.19.0
```
sergei-maertens: Use npm please

src/openzaak/templates/admin/nlx_config.html
```diff
@@ -0,0 +1,61 @@
+{% extends "admin/base_site.html" %}
+{% load i18n static %}
+
+{% block extrastyle %}{{ block.super }}
+    <link href="{% static 'libs/prism/prism.css'%}" rel="stylesheet" />
+
+{% endblock %}
+
+
+{% block title %} {% trans "Importeer catalogus" %} {{ block.super }} {% endblock %}
```
sergei-maertens: Incorrect title

src/openzaak/templates/admin/nlx_config.html
```diff
@@ -0,0 +1,61 @@
...
+<a href="{% url 'admin:index' %}">{% trans 'Home' %}</a>
+&rsaquo; <a href="#">{% trans 'Configuratie' %}</a>
+&rsaquo; {% trans 'NLX configuratie' %}
+</div>
+{% endblock %}
+
+{% block content %}
+    <h1>{% trans 'NLX configuratie' %}</h1>
+    <div id="content-main">
+        <form action="" method="POST" enctype="multipart/form-data">
```
sergei-maertens: Why is this a form? There's nothing to submit, no?

src/openzaak/templates/admin/nlx_config.html
```diff
@@ -0,0 +1,61 @@
...
+                <fieldset class="module aligned ">
+
+                    <p>
+                        Gebruik onderstaand voorbeeld voor de configuratie van de NLX-inway.
+                    </p>
+                    <p>
+                        U kunt uw <code>service-config.toml</code> bestand aanpassen met de configuratie onder <code>[services]</code>.
+                    </p>
+
+                    <pre class="language-toml"><code>[services]
```
sergei-maertens: I'm curious if using https://pypi.org/project/django-pygments/ wouldn't be a better option than adding a frontend library? Pygments is used by Sphinx as well, so it'll resemble the documentation more.

src/openzaak/templates/admin/nlx_config.html
```diff
@@ -0,0 +1,61 @@
...
+{% for service in services %}
+    [services.OpenZaak-{{ service.name }}]
```
sergei-maertens: @joeribekker what do we want to do about casing? so far I've always defined everything lower-cased

src/openzaak/templates/admin/nlx_config.html
annashamray: The last update of `django-pygments` was 6 years ago and their site doesn't work. So actually I don't want to use them

src/openzaak/templates/admin/nlx_config.html
joeribekker: That was me, I now remember that was to submit some variables to change the config file, but removed it later. I'm not sure if removing it will break the layout but if not can be removed.

src/openzaak/templates/admin/nlx_config.html
joeribekker: Hmm. I wanted to see a nicer name in the NLX directory but it seems NLX doesn't care too much since I opened a [ticket about adding a pretty name variable 6 months ago](https://gitlab.com/commonground/nlx/nlx/issues/658).

Oh well, it's just for sysadmins/developers I guess, so we can use lower-case to keep it consistent.

src/openzaak/static/libs/prism/prism.css
annashamray: added `prism` via npm

src/openzaak/templates/admin/nlx_config.html
annashamray: changed to "NLX Configuratie".

src/openzaak/templates/admin/nlx_config.html
annashamray: form is removed. Looks ok for me

src/openzaak/templates/admin/nlx_config.html
annashamray: Service name is changed to lower-case

src/openzaak/templates/admin/nlx_config.html
sergei-maertens: good call, I saw the latest release on PyPI from November 2019, but the latest work in the repo is effectively from 6 years ago :+1:

Issue #419, opened 2020-01-15
As a developer, I want to be sure that all admin pages work without errors with read-only permissions


sergei-maertens:
... so I can look at my date instead of seeing 500 error pages

**Description**

There are a bunch of custom admin forms used, where some fields are accessed with the pattern `self.fields[fieldname]`. This works if you can edit the form, but for read-only permissions, those fields are absent from the form because of how the Django admin works.

When that happens, the end-user will see an error page.

We need to go through the admin pages with a read-only account and :

- check if/where this happens
- write an integration test
- fix the custom form

joeribekker:
I find looking at data to be optional... but added it to sprint anyway ;-)

PR #439, opened 2020-01-24
Issue/admin read only 500

stevenbal:
issue #419


stevenbal:
@sergei-maertens I noticed that the custom buttons for actions like publishing zaaktypen and importing catalogi/zaaktypen are visible and clickable for read-only users as well. Should I make custom permissions for this or can I use the existing ones like `catalogi_add_zaaktype`, etc.?

sergei-maertens:
@stevenbal should be the same permissions as the import page (import page -> add permission I guess) or the change page (for publish action)

Issue #404, opened 2020-01-09
As installer, I want to have a checklist on how to check if Open Zaak works as expected


joeribekker:
... so I can leave a municipality with confidence.

**Description**
Document a post-install-and-setup checklist and/or a default simple Postman collection to retrieve a list of Zaken and create a Zaak.

A `manage.py raven test` should also be part of the checklist.

PR #440, opened 2020-01-24
Fixes #404 -- added checklist for open zaak post install

stevenbal:
issue #404 

I will make separate pull requests for the notification management command and view-config changes

PR #442, opened 2020-01-24
Added management command to test notification sending

stevenbal:
related to #404

### DIFF
--- a/src/openzaak/management/commands/send_test_notification.py
+++ b/src/openzaak/management/commands/send_test_notification.py
@@ -1,0 +1,39 @@
+from django.core.management.base import BaseCommand, CommandError
+
+from vng_api_common.notifications.kanalen import Kanaal
+from vng_api_common.notifications.models import NotificationsConfig
+
+
+class Command(BaseCommand):
+    help = "Send a test notification to verify if notifications are properly configured"
+
+    def handle(self, *args, **options):
+        nrc_client = NotificationsConfig.get_client()
+
+        kanaal_name = "test"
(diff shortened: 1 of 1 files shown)


src/openzaak/management/commands/send_test_notification.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+class Command(BaseCommand):
+    help = "Send a test notification to verify if notifications are properly configured"
+
+    def handle(self, *args, **options):
+        nrc_client = NotificationsConfig.get_client()
+
+        data = {
+            "kanaal": "zaken",
```
stevenbal: @sergei-maertens not sure if there is a way to create a `test` kanaal without having do define it in kanalen.py?

sergei-maertens: instantiating a `Kanaal` object should register it :)

src/openzaak/management/commands/send_test_notification.py
```diff
@@ -0,0 +1,39 @@
...
+    help = "Send a test notification to verify if notifications are properly configured"
+
+    def handle(self, *args, **options):
+        nrc_client = NotificationsConfig.get_client()
+
+        kanaal_name = "test"
+        Kanaal(kanaal_name, "test")
+
+        data = {
+            "kanaal": kanaal_name,
```
sergei-maertens: okay, but now it's actually going to fail unless the "test" kanaal has been registered :P

Something to add to the checklist in #440  - make sure that the `test` kanaal exists.

PR #445, opened 2020-01-27
[WIP] Ship deployment tooling as docker image

sergei-maertens:
Extracted from #382 so that I can close that PR, will finish some other day...


sergei-maertens:
The rest of the tooling is sufficient for now.

Issue #197, opened 2019-11-07
As admin if I publish a zaaktype I want the page to be in readonly mode


Hugo-ter-Doest:
... so I can clearly see that the zaaktype is publised and can no longer be modified.

**Description**
Describe the user story in more detail.

PR #447, opened 2020-01-27
Feature/admin readonly zaaktype

annashamray:
Issue - https://github.com/open-zaak/open-zaak/issues/197

### DIFF
--- a/src/openzaak/components/catalogi/tests/admin/test_published_readonly.py
+++ b/src/openzaak/components/catalogi/tests/admin/test_published_readonly.py
@@ -4,7 +4,7 @@ import requests_mock
 from django_webtest import WebTest
 
 from openzaak.accounts.tests.factories import SuperUserFactory
-from openzaak.selectielijst.tests import mock_oas_get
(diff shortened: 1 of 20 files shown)


annashamray:
@sergei-maertens custom formatting of read-only fields was introduced, now all fields from the check-list should be rendered nicely

src/openzaak/components/catalogi/tests/admin/test_published_readonly.py
```diff
@@ -0,0 +1,225 @@
...
+
+        form = response.form
+        form_fields = list(form.fields.keys())
+        informatieobjecttype_fields = [
+            f.name
+            for f in informatieobjecttype._meta.get_fields()
+            if f.name != "datum_einde_geldigheid"
+        ]
+
+        self.assertEqual("datum_einde_geldigheid" in form_fields, True)
```
sergei-maertens: For future reference - you don't have to change it now - there is a `self.assertIn` assertion as well :-)

Issue #356, opened 2020-01-02
Authorizations should check for overlap to prevent fuzzy authorizations


joeribekker:
**Describe the bug**
Authorizations should check for overlap to prevent fuzzy authorizations

See: https://github.com/open-zaak/open-zaak/pull/307/files#diff-f40996d24e342971921c649c3de886a1R377

joeribekker:
This should actually be enforced on API-level (so in the VNG standard) but it's not (new ticket incoming) but for Open Zaak we can enforce this on UI-level.

annashamray:
Implemented in https://github.com/open-zaak/open-zaak/pull/450

PR #450, opened 2020-01-28
check for zaaktype overlap in the auth admin

annashamray:
issue - https://github.com/open-zaak/open-zaak/issues/356

### DIFF
--- a/src/openzaak/components/autorisaties/forms.py
+++ b/src/openzaak/components/autorisaties/forms.py
@@ -2,6 +2,7 @@ from typing import List, Tuple
 
 from django import forms
 from django.conf import settings
+from django.core.exceptions import ValidationError
 from django.db import transaction
 from django.utils.translation import ugettext_lazy as _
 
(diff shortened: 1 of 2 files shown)


src/openzaak/components/autorisaties/forms.py
```diff
@@ -428,8 +437,31 @@ def save(self, commit=True):
...
+                types = set(form.get_types() or [])
+                scopes = form.cleaned_data["scopes"]
+                types_field = COMPONENT_TO_FIELDS_MAP[form.cleaned_data["component"]][
+                    "types_field"
+                ]
+                for scope in scopes:
+                    previous_types = scope_types.get(scope, set())
+                    if previous_types.intersection(types):
+                        raise ValidationError(
+                            _("{} should not overlap scopes in one Applicatie").format(
```
sergei-maertens:
```suggestion
                            _("{field} may not have overlapping scopes.").format(
```

1. Phrasing fixed
2. Always use keyword arguments in translatable strings - different languages have different order of nouns/verbs, having named variables makes this easier to achieve.

PR #451, opened 2020-01-29
Fixes #167 -- point to hosted selectielijst as default

### LINKED ISSUE #167: As municipality, I want to have the ReferentieLijsten API available
... so I can use it for the selectielijst.

... so I can use the GemeenteZaak API's independently from an external API for ReferentieLijsten (that probably will not be available in the near future).

**Description**

See https://referentielijsten-api.vng.cloud/api/v1/schema/

**Tasks**

* [x] Fork
* [x] Production grade (read: performance measurements)
* [ ] Add caching
* [x] Hosting setup in k8s (not dedicated)
* [x] See if VNG can host this anyway

No documentation needed. It needs to be up and running "somewhere".

sergei-maertens:
Changed the default Selectielijst API.

It's available, but the SSL certificate needs to be sorted out first.

- [x] SSL fixed


src/openzaak/tests/schemas/referentielijsten.yaml
```diff
@@ -0,0 +1,1488 @@
+openapi: 3.0.0
```
sergei-maertens: should be deleted in favour of `src/openzaak/selectielijst/tests/files/openapi.yaml`

sergei-maertens:
Just had a status update from Cyso - the SSL certificate is not happening yet. So I'll include a data migration to move from VNG hosted version to this one when it's ready.

joeribekker:
@annashamray Can you fix this heisenbug?

PR #452, opened 2020-01-29
Fixes #396 -- Document the need to restart containers if yo…

### LINKED ISSUE #396: Changing .env file does not trigger a Docker restart/reload
**Describe the bug**
From our findings at gemeente Buren:

Changes made to the .env file are not picked up by the Docker containers. Either we make them pick up or document you need to restart the containers.

sergei-maertens:
…u fidget with the .env file

### DIFF
--- a/docs/installation/deployment/single_server.rst
+++ b/docs/installation/deployment/single_server.rst
@@ -166,6 +166,21 @@ A full example might look like this:
    outcome. If you decide to change configuration parameters, you do not have
    to start from scratch.
 
+**Changing environment variables**
+
+The Open Zaak configuration is templated out to ``/home/openzaak/.env`` on the host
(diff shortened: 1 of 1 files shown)


joeribekker:
Fixes #396 (didn't get picked up from title change).

joeribekker:
@sergei-maertens Is this the recommended way? If not, we should indicate so and refer to the preffered method (ansible?)

sergei-maertens:
Well, the `.env` template is included in the Ansible role. I want to hold off on it for now to not overengineer the roles, but it would require changes to the roles either way.

Option 1: define a variable holding key-value pairs of extra things to add to the .env file, can easily be templated out
Option 2: add the ability to specify a custom `.env` template file

I'm not sure if/how much people will actually be modifying the `.env` file. I'd rather wait until there's a need?

PR #453, opened 2020-01-30
Integrated Open Notificaties deployment

sergei-maertens:
The actual deployment is in https://github.com/open-zaak/open-notificaties/pull/12, but some notable changes:

* nginx template is moved to ansible collection
* nginx vhost is now in `openzaak.conf` instead of `default.conf` so that Open Notificaties can be installed alongside
* fixed some docs here and there


docs/installation/deployment/single_server.rst
```diff
@@ -114,7 +117,9 @@ server.
 
 .. warning:: It's important to use the correct domain name, as the SSL certificate
    will be generated for this domain and only this domain will be whitelisted
-   by Open Zaak!
+   by Open Zaak! If you are using a private DNS name, then no SSL certificate
+   can be created via Letsencrypt - make sure to disable it by setting
+   ``certbot_create_if_missing=false``.
```
joeribekker:
```suggestion
   ``certbot_create_if_missing=false`` or ``openzaak_ssl=false`` if you don't
   plan on using HTTPS at all.
```

Issue #153, opened 2019-11-01
As developer, I want to know where Open Zaak API's differ from the standard


joeribekker:
... so I can take this into account when developing.

**Description**
Some issues are mandatory from our point of view but not (yet) added to the standard. This ticket is a placeholder to make an overview of such issues and document them.

- ~#118~ (this was integrated into the standard by VNG)

sergei-maertens:
## Autorisaties API

no differences found

## Besluiten API

- [ ] Description `besluit_create`
- [ ] Description `besluit_update`
- [ ] Description `besluit_partial_update`
- [ ] Schema `Besluit.besluittype` is missing `minLength: 1`

## Catalogi API

## Documenten API

## Zaken API

sergei-maertens:
Created https://github.com/VNG-Realisatie/gemma-zaken/issues/1542 for the leftovers

PR #454, opened 2020-01-30
Fixes #298 -- block zaak data mutations if the zaak is closed

### LINKED ISSUE #298: De referentieimplementatie laat op dit moment toe dat nieuwe documenten aan een gearchiveerde zaak gekoppeld worden
# Bug

Na het archiveren van een zaak is deze in principe immutable, het zou dus door (reguliere) users niet mogelijk mogen zijn om nieuwe documenten te relateren aan deze zaken.

Check:
- [x] Documenten koppelen
- [x] Nieuwe statussen zetten (mag mits scope heropenen)
- [x] Zaakobjecten
- [x] Zaakeigenschappen
- [x] ZaakInformatieObjecten
- [x] Klantcontacten
- [x] Rollen
- [x] Resultaten
- [x] ZaakBesluiten

---

Collection of what the standard describes:

From [zrc-007](https://github.com/VNG-Realisatie/gemma-zaken/blob/master/docs/_content/standaard/zaken/index.md#afsluiten-zaak-zrc-007):

> Als een Zaak een eindstatus heeft dan is de zaak afgesloten en mogen gegevens van de zaak niet meer aangepast worden (behalve om redenen van correctie). Dit MOET beveiligd worden met de scope zds.scopes.zaken.geforceerd-bijwerken.

From [zrc-021](https://github.com/VNG-Realisatie/gemma-zaken/blob/master/docs/_content/standaard/zaken/index.md#zrc-021):

> Het resultaat van een zaak is bepalend voor het archiefregime. Bij het afsluiten van een zaak MOETEN de attributen Zaak.archiefnominatie en Zaak.archiefactiedatum bepaald worden...

-> "Archivering" vindt plaats bij het afsluiten van de zaak

sergei-maertens:
Changing of Zaak data is not allowed if the Zaak is closed. This extends to related objects:

- [x] ZaakInformatieObject
- [x] ZaakObject
- [x] Resultaat
- [x] KlantContact
- [x] Status (special exception scope)
- [x] ZaakEigenschap
- [x] Rol
- [x] ZaakBesluit

However, the updates are allowed if you have the scope `zaken.geforceerd-bijwerken`.

- [x] ZaakInformatieObject
- [x] ZaakObject
- [x] Resultaat
- [x] KlantContact
- [x] Status (special exception scope)
- [x] ZaakEigenschap
- [x] Rol

### DIFF
--- a/src/openzaak/components/zaken/api/mixins.py
+++ b/src/openzaak/components/zaken/api/mixins.py
@@ -1,0 +1,42 @@
+from django.db import models
+
+from rest_framework import serializers
+
+from .exceptions import ZaakClosed
+
+
+class ClosedZaakMixin:
+    def perform_create(self, serializer: serializers.ModelSerializer) -> None:
+        """
+        Block the create if the related zaak is closed.
+
+        :raises: PermissionDenied if the related Zaak is closed.
+        """
+        zaak = serializer.validated_data.get("zaak")
+        if zaak and zaak.is_closed:
+            raise ZaakClosed()
+
+        super().perform_create(serializer)
+
+    def perform_update(self, serializer: serializers.ModelSerializer) -> None:
+        """
+        Block the update if the related zaak is closed.
+
+        :raises: PermissionDenied if the related Zaak is closed.
+        """
+        zaak = serializer.instance.zaak
+        if zaak and zaak.is_closed:
+            raise ZaakClosed()
+
+        super().perform_update(serializer)
+
+    def perform_destroy(self, instance: models.Model) -> None:
+        """
+        Block the destroy if the related zaak is closed.
+
+        :raises: PermissionDenied if the related Zaak is closed.
+        """
+        if instance.zaak and instance.zaak.is_closed:
+            raise ZaakClosed()
+
+        super().perform_destroy(instance)
--- a/src/openzaak/components/zaken/api/mixins.py
+++ b/src/openzaak/components/zaken/api/mixins.py
@@ -1,11 +1,48 @@
+from typing import Optional
+
 from django.db import models
 
 from rest_framework import serializers
 
+from openzaak.components.zaken.api.scopes import SCOPE_ZAKEN_GEFORCEERD_BIJWERKEN
+from openzaak.components.zaken.models import Zaak
+
 from .exceptions import ZaakClosed
+from .serializers import ZaakSerializer
 
 
 class ClosedZaakMixin:
+    def _has_override(self, zaak: Zaak) -> bool:
+        jwt_auth = self.request.jwt_auth
+        zaak_data = ZaakSerializer(zaak, context={"request": self.request}).data
+        return jwt_auth.has_auth(
+            scopes=SCOPE_ZAKEN_GEFORCEERD_BIJWERKEN,
+            zaaktype=zaak_data["zaaktype"],
+            vertrouwelijkheidaanduiding=zaak_data["vertrouwelijkheidaanduiding"],
+            init_component=self.queryset.model._meta.app_label,
+        )
+
+    def _check_zaak_closed(self, zaak: Optional[Zaak] = None) -> None:
+        """
+        Raise ZaakClosed if an attempt is made to modify a closed zaak.
+
+        Unless you have the scope SCOPE_ZAKEN_GEFORCEERD_BIJWERKEN, an application
+        is not allowed to mutate a closed zaak.
+        """
+        # if there is no zaak involved, nothing to do
+        if not zaak:
+            return
+
+        # zaak open -> don't raise exceptions
+        if not zaak.is_closed:
+            return
+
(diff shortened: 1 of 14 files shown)


src/openzaak/components/zaken/api/mixins.py
```diff
@@ -0,0 +1,74 @@
...
+        raise ZaakClosed()
+
+    def perform_create(self, serializer: serializers.ModelSerializer) -> None:
+        """
+        Block the create if the related zaak is closed.
+
+        :raises: PermissionDenied if the related Zaak is closed.
+        """
+        zaak = serializer.validated_data.get("zaak")
+        self._check_zaak_closed(zaak)
```
annashamray: Why the check `self._check_zaak_closed` is put here, but not into `validate` method? I thought all checks are better placed in `validate`. Or in this case it doesn't matter?

src/openzaak/components/zaken/tests/test_zaak_closed.py
```diff
@@ -124,3 +156,371 @@ def test_reopenzaak_not_allowed(self):
...
+
+        resultaat.delete()
+
+        data = {
+            "zaak": reverse(self.zaak),
+            "resultaattype": f"http://testserver{reverse(resultaat.resultaattype)}",
+        }
+        self.assertCreateBlocked(reverse(Resultaat), data)
+
+    def test_zaakbesluiten(self):
```
annashamray: here the `Besluit` resource is only tested. I think it's nice to have `ZaakBesluit` test also, because in case `besluiten` component is external  the check for closed zaak will be made only in `ZaakBesluit` resource

sergei-maertens: I thought about it too, but saw that we were currently doing it this way in `ZaakViewSet` and wanted to keep the behaviour consistent. It's now raising a 403 (not a 400), since it is also a permission error (you're lacking the overide scope). But I think you can argue for both cases.

src/openzaak/components/zaken/api/mixins.py
sergei-maertens: I thought about it too, but saw that we were currently doing it this way in ZaakViewSet and wanted to keep the behaviour consistent. It's now raising a 403 (not a 400), since it is also a permission error (you're lacking the overide scope). But I think you can argue for both cases.

src/openzaak/components/zaken/tests/test_zaak_closed.py
sergei-maertens: I tried, but it's pretty much impossible with the signals we have.

In the relation side, `Besluit` is leading and `ZaakBesluit` follows. If the `besluiten` component is external, they have to perform the check that the zaak is not closed.

src/openzaak/components/zaken/api/viewsets.py
```diff
@@ -580,8 +588,8 @@ class ZaakEigenschapViewSet(
     required_scopes = {
         "list": SCOPE_ZAKEN_ALLES_LEZEN,
         "retrieve": SCOPE_ZAKEN_ALLES_LEZEN,
-        "create": SCOPE_ZAKEN_BIJWERKEN,
-        "destroy": SCOPE_ZAKEN_BIJWERKEN,
+        "create": SCOPE_ZAKEN_BIJWERKEN | SCOPE_ZAKEN_GEFORCEERD_BIJWERKEN,
```
joeribekker: This will no doubt cause a patch-version change to the API spec.

Shouldn't we generate the API spec as well as part of this PR?

src/openzaak/components/zaken/api/mixins.py
```diff
@@ -0,0 +1,74 @@
...
+        # if there is no zaak involved, nothing to do
+        if not zaak:
+            return
+
+        # zaak open -> don't raise exceptions
+        if not zaak.is_closed:
+            return
+
+        # zaak is closed - do we have the special i-can-do-anything permission?
+        if self._has_override(zaak):
```
joeribekker: I like this pattern.

sergei-maertens:
@joeribekker yep, I'll re-generate, did so too in RI: https://github.com/VNG-Realisatie/gemma-zaakregistratiecomponent/pull/139

src/openzaak/components/zaken/openapi.yaml
```diff
@@ -2023,6 +2023,27 @@ paths:
       operationId: rol_delete
       summary: Verwijder een ROL van een ZAAK.
       description: Verwijder een ROL van een ZAAK.
+      parameters:
```
sergei-maertens: @joeribekker these are missing in the RI, they'll show up with the final check for #153 - it's a bug in the 1.0.x standard to be fixed.

src/openzaak/components/zaken/api/viewsets.py
sergei-maertens: Done

PR #457, opened 2020-01-31
Provide individual API versions for each API

sergei-maertens:
* Each API has a (potentially) different patch version
* Ensure the API version header is correctly returned
* Ensure the API version is correctly reflected in the API spec

### DIFF
--- a/src/openzaak/management/commands/generate_swagger_component.py
+++ b/src/openzaak/management/commands/generate_swagger_component.py
@@ -1,3 +1,5 @@
+from django.conf import settings
+
 from vng_api_common.management.commands import generate_swagger
 
 SCHEMA_MAPPING = {
@@ -30,8 +32,12 @@ class Command(generate_swagger.Command):
         urlconf,
         component=None,
         *args,
-        **options
+        **options,
     ):
+        _version = getattr(settings, f"{component.upper()}_API_VERSION")
+
+        # Setting must exist for vng-api-common, so monkeypatch it in
+        settings.API_VERSION = _version
 
         if not component:
             super().handle(
@@ -45,7 +51,7 @@ class Command(generate_swagger.Command):
                 info,
                 urlconf,
                 *args,
-                **options
+                **options,
             )
 
         # rewrite command arguments based on the component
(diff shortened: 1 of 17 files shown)


src/openzaak/management/commands/generate_swagger_component.py
```diff
@@ -30,8 +32,12 @@ def handle(
         urlconf,
         component=None,
         *args,
-        **options
+        **options,
     ):
+        _version = getattr(settings, f"{component.upper()}_API_VERSION")
+
+        # Setting must exist for vng-api-common, so monkeypatch it in
+        settings.API_VERSION = _version
```
joeribekker: ~In that case we can keep the placeholder setting in [src/openzaak/conf/includes/api.py](https://github.com/open-zaak/open-zaak/pull/457/files#diff-10f23095d64e39bd1017698c8afa68bfR4) that you `del`'d explicitly.~ Oh it's a management command...

src/openzaak/utils/middleware.py
```diff
@@ -15,3 +24,45 @@ def __call__(self, request):
...
+        if version is not None:
+            response[VERSION_HEADER] = version
+
+        return response
+
+    def _get_version(self, path: str) -> Optional[str]:
+        for prefix, version in self.version_mapping.items():
+            if path.startswith(prefix):
+                return version
+        return None
```
joeribekker: Feels a little crufty. You add the version to the response while we already know the version upon entering the API: In the URL-patterns.

Can't we pass to the root API URL-pattern something like: `url('/', zaken_router, kwargs={'version': settings.ZAKEN_API_VERSION})`  so that's its available to the views (for later ^_^) and in the response, we simply pass this value again without the need to look up which URL the response came from?

sergei-maertens: but `version` is only the major versions, which is used in the URLs, while the header contains the full `major.minor.patch` version, or am I missing something?

joeribekker: I meant a new version argument, not the existing one used for major version in the `v<verison>` url path, ie `request_version`=request.REQUESTED_VERSION or settings.API_VERSION

But, can be done later.

PR #461, opened 2020-02-04
:bug: Fixes #460 -- Don't check for type overlap if not needed.

### LINKED ISSUE #460: Cannot create authorisations for all components
**Describe the bug**
When creating read-only authorisations in Open Zaak for all components (audittrail.lezen, \<component\>.lezen, I got a 500-error.

See: https://sentry.maykinmedia.nl/maykin-media/open-zaak/issues/265662/

joeribekker:


src/openzaak/components/autorisaties/forms.py
```diff
@@ -322,12 +322,13 @@ def _validate_external_types(self, component):
                 )
                 self.add_error("externe_typen", error)
 
-    def get_types(self):
-        component = self.cleaned_data["component"]
+    def get_types(self, component):
```
annashamray: I don't quite understand why to change 'get_types' parameters?

src/openzaak/components/autorisaties/forms.py
```diff
@@ -322,12 +322,13 @@ def _validate_external_types(self, component):
...
         related_type_selection = self.cleaned_data.get("related_type_selection")
         types = None
         if related_type_selection:
-            _field_info = COMPONENT_TO_FIELDS_MAP[component]
+            _field_info = COMPONENT_TO_FIELDS_MAP.get(component)
+            if _field_info is None:
```
annashamray: I'm not sure if this check is required. If `related_type_selection` is filled, that means that some types should be chosen. And if not, it's an error.

src/openzaak/components/autorisaties/forms.py
```diff
@@ -445,11 +446,15 @@ def _validate_overlapping_types(self):
...
+                component = form.cleaned_data["component"]
+                types = set(form.get_types(component) or [])
                 scopes = form.cleaned_data["scopes"]
-                types_field = COMPONENT_TO_FIELDS_MAP[form.cleaned_data["component"]][
-                    "types_field"
-                ]
+
+                _field_info = COMPONENT_TO_FIELDS_MAP.get(component)
+                if _field_info is None:
+                    continue
```
annashamray: I think if there are no types we still should validate scopes for overlapping or in this case duplicating each other.

src/openzaak/components/autorisaties/forms.py
joeribekker: well, all your other helpers had that argument and everytime `get_types` was called, the component was already known (via the same line of code that I removed here).

src/openzaak/components/autorisaties/forms.py
joeribekker: i agree, will fix

src/openzaak/components/autorisaties/forms.py
joeribekker: This function check for overlapping types. How can you check for types if there's no types field?

The rest of the for-loop continues right?

PR #464, opened 2020-02-04
fix overlapping check for auth without types

annashamray:
issue - https://github.com/open-zaak/open-zaak/issues/460#event-3005437071


joeribekker:
Merged with #461

PR #465, opened 2020-02-04
set omschrijving_generiek to prevent triggering RolOccurenceValidator

annashamray:
Issue - https://github.com/open-zaak/open-zaak/pull/451#issuecomment-581904630

### DIFF
--- a/src/openzaak/components/zaken/tests/test_zaak_closed.py
+++ b/src/openzaak/components/zaken/tests/test_zaak_closed.py
@@ -273,12 +273,12 @@ class ClosedZaakRelatedDataNotAllowedTests(JWTAuthMixin, APITestCase):
 
     def test_rollen(self):
         roltype = RolTypeFactory.create(
-            zaaktype=self.zaaktype, omschrijving_generiek=RolOmschrijving.initiator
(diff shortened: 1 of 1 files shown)


src/openzaak/components/zaken/tests/test_zaak_closed.py
```diff
@@ -272,8 +272,14 @@ def test_klantcontacten(self):
...
     def test_rollen(self):
-        roltype = RolTypeFactory.create(zaaktype=self.zaaktype)
-        rol = RolFactory.create(zaak=self.zaak, roltype=roltype)
+        roltype = RolTypeFactory.create(
+            zaaktype=self.zaaktype, omschrijving_generiek=RolOmschrijving.initiator
+        )
+        rol = RolFactory.create(
+            zaak=self.zaak,
+            roltype=roltype,
+            omschrijving_generiek=RolOmschrijving.zaakcoordinator,
```
sergei-maertens: this is not allowed to be different from the `roltype` value - maybe set it to `belanghebbende` instead? you can have multiple of those

annashamray: Thanks. I missed it. `omschrijving_generiek` values are replaces with `RolOmschrijving.behandelaar`

Issue #480, opened 2020-02-13
Resultaattype from Selectielijst doesn't seem to respect or see the Zaaktype processtype.


joeribekker:
**Describe the bug**
Selectielijst Zaaktype doesn't seem to show. And the possible resultaattypes dont seem to be filtered correctly (tried with referentielijsten-api.vng.cloud and selectielijst.openzaak.nl)

![image](https://user-images.githubusercontent.com/96970/74415670-15e08600-4e44-11ea-9425-bf2fcc959c35.png)

**To Reproduce**
1. Create Zaaktype
2. Add Resultaattype
3. Check selectielijstklasse

**Expected behavior**
Zaaktype processtype should be shown and ist should be filtered

annashamray:
This case only happens when a new `resultaattype` is being created and it's `zaaktype` field is not save to DB yet. Once you choose the `zaaktype` and submit the form the `Zaaktype processtype` field and filtering of `selectielijstklasse` will appear

joeribekker:
Can we add a message in that case where the "-" (dash) is now as value for Zaaktype:

"Please save this Resultaattype first to get proper filtering of selectielijstklasses" (Kies "Opslaan en opnieuw bewerken" om de bijpassende selectielijstklassen te filteren.")

joeribekker:
Or better, grab the Zaaktype from the query parameters of not on the object (`?zaaktype=X`)

PR #485, opened 2020-02-17
Issue/zaaktype procestype selectielijst

stevenbal:
fixes https://github.com/open-zaak/open-zaak/issues/480

### DIFF
--- a/src/openzaak/components/catalogi/admin/resultaattype.py
+++ b/src/openzaak/components/catalogi/admin/resultaattype.py
@@ -10,7 +10,7 @@ from openzaak.selectielijst.admin_fields import (
 from openzaak.selectielijst.models import ReferentieLijstConfig
 from openzaak.utils.admin import UUIDAdminMixin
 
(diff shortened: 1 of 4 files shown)


src/openzaak/components/catalogi/admin/resultaattype.py
```diff
@@ -41,8 +41,21 @@ class ResultaatTypeAdmin(
         "toelichting",
     )
 
+    def get_extra_context(self, request, *args, **kwargs):
+        context = super().get_extra_context(request, *args, **kwargs)
+        self.zaaktype = context.get("zaaktype")
```
sergei-maertens: I don't like setting random attributes during a method call, is there no better way?

stevenbal: I have tried, but I don't think there is another way to access the request from the `get_zaaktype_procestype` zaaktype method, at least I couldn't find one

PR #497, opened 2020-03-05
Support NLX to fetch objects in other APIs

sergei-maertens:
- [x] Unit test added
- [x] Implementation done
- [x] Docs updated


sergei-maertens:
@annashamray yeah, unless that uses `settings.LINK_FETCHER` somehow but I don't think it does.

PR #509, opened 2020-03-13
Fixes #507 -- added download with original filename to EIO admin

### LINKED ISSUE #507: As admin, I want to download a document, in the admin, by its filename
... so I can open it directly.

**Description**
Currently, the're a "inhoud" field that let's you download the contents of the Document/Informatieobject in the admin. This file is by default a `.bin` file.

Most of the time, a filename (bestandsnaam) is part of the metadata. If that's the case, I want a link that let's me download the file with it's proper filename. This can be a new line/field in the admin.

PS. This wish was a result of the Delft POC.

stevenbal:
issue: https://github.com/open-zaak/open-zaak/issues/507

### DIFF
--- a/src/openzaak/components/documenten/admin.py
+++ b/src/openzaak/components/documenten/admin.py
@@ -181,6 +181,8 @@ class EnkelvoudigInformatieObjectAdmin(
     raw_id_fields = ("canonical", "_informatieobjecttype")
     viewset = viewsets.EnkelvoudigInformatieObjectViewSet
     private_media_fields = ("inhoud",)
+    private_media_view_class = PrivateMediaView
+    private_media_file_widget = PrivateFileWidget
 
     fieldsets = (
         (
@@ -252,29 +254,4 @@ class EnkelvoudigInformatieObjectAdmin(
         return (
             link_to_related_objects(Gebruiksrechten, obj.canonical),
             link_to_related_objects(ObjectInformatieObject, obj.canonical),
-        )
-
-    def formfield_for_dbfield(self, db_field, request, **kwargs):
-        field = super().formfield_for_dbfield(db_field, request, **kwargs)
-        private_media_fields = self.get_private_media_fields()
-        if db_field.name in private_media_fields:
-            view_name = self._get_private_media_view_name(db_field.name)
-            obj = self.get_object(
-                request, request.resolver_match.kwargs.get("object_id")
-            )
-            attrs = {}
-            if obj:
-                display_value = obj.bestandsnaam if obj.bestandsnaam else obj.inhoud.url
-                attrs["display_value"] = display_value
-            field.widget = PrivateFileWidget(
-                url_name="admin:%s" % view_name, attrs=attrs,
-            )
-        return field
-
-    def get_private_media_view(self, field):
-        return PrivateMediaView.as_view(
-            model=self.model,
-            file_field=field,
-            permission_required=self.get_private_media_permission_required(field),
-            sendfile_options=self.get_private_media_view_options(field),
         )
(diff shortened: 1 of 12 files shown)


src/openzaak/components/documenten/admin.py
```diff
@@ -23,6 +26,22 @@
 )
 
 
+class FileDownloadMixin:
```
sergei-maertens: Please don't make _everything_ a mixin. There is only one place where this is used now - this should really just be a method on the `ModelAdmin` where it's used.

src/openzaak/components/documenten/admin.py
```diff
@@ -251,3 +273,22 @@ def get_object_actions(self, obj):
             link_to_related_objects(Gebruiksrechten, obj.canonical),
             link_to_related_objects(ObjectInformatieObject, obj.canonical),
         )
+
+    def get_urls(self):
```
sergei-maertens: why are you not re-using the URL(s) from django-privates that are already provided?

src/openzaak/components/documenten/admin.py
```diff
@@ -251,3 +273,22 @@ def get_object_actions(self, obj):
...
+                "<int:pk>/download/",
+                self.download,
+                name="documenten_enkelvoudiginformatieobject_download",
+            ),
+        ]
+        return my_urls + urls
+
+    def download(self, request, pk):
+        obj = self.get_object(request, pk)
+        response = FileResponse(obj.inhoud)
```
sergei-maertens: You _really_ don't want to have Django serving potentially massive files from disk. This should use `sendfile` on nginx, all of this is implemented in django-privates already. Please look up the documentation/implementation of django-privates on how to customize the attachment filename.

src/openzaak/components/documenten/admin.py
stevenbal: I figured since I added it to both the `EnkelvoudigInformatieObjectInline` and the `EnkelvoudigInformatieObjectAdmin`

sergei-maertens: my bed, I missed that in the diff then :) 😅

src/openzaak/components/documenten/admin.py
```diff
@@ -251,3 +253,23 @@ def get_object_actions(self, obj):
...
+
+    def formfield_for_dbfield(self, db_field, request, **kwargs):
+        field = super().formfield_for_dbfield(db_field, request, **kwargs)
+        private_media_fields = self.get_private_media_fields()
+        if db_field.name in private_media_fields:
+            view_name = self._get_private_media_view_name(db_field.name)
+            obj = self.get_object(request, request.resolver_match.kwargs["object_id"])
+            field.widget = PrivateFileWidget(
+                url_name="admin:%s" % view_name,
+                attrs={"display_value": obj.bestandsnaam},
```
stevenbal: @sergei-maertens there has to be a better way to change the display value for the download link of the file in the admin right? I want it to show the `bestandsnaam` instead of the path where it is stored in private_media

src/openzaak/components/documenten/tests/admin/test_eio_download.py
```diff
@@ -0,0 +1,38 @@
...
+        bestand_field = response.html("div", {"class": "field-inhoud"})[0]
+        download_link = bestand_field("a")[0]
+
+        self.assertEqual(download_link.text, "iets.txt")
+
+        download_response = self.app.get(download_link.attrs["href"])
+        self.assertEqual(download_response.content, b"STUFF")
+        self.assertEqual(
+            download_response.content_disposition, 'attachment; filename="iets.txt"'
+        )
```
joeribekker: Can you add a test what the filename is when "bestandsnaam" is empty?

PR #515, opened 2020-03-17
wizard form - NLX step

annashamray:
Issue - https://github.com/open-zaak/open-zaak/issues/518
gist - https://gist.github.com/sergei-maertens/5e595d28ced3fe900a8d2ae8ba18479e

Step1. NLX


src/openzaak/config/forms.py
```diff
@@ -0,0 +1,31 @@
...
+        outway = self.cleaned_data["outway"]
+
+        if not outway:
+            return outway
+
+        # try to tcp connect to the port
+        parsed = urlparse(outway)
+        with socket.socket() as s:
+            try:
+                s.connect((parsed.hostname, parsed.port))
```
sergei-maertens: I'd like to see a timeout specified, 2s for example should be enough: https://docs.python.org/3.6/library/socket.html#timeouts-and-the-connect-method

src/openzaak/config/models.py
```diff
@@ -0,0 +1,21 @@
...
+from solo.models import SingletonModel
+
+from .constants import NLXDirectories
+
+
+class NLXConfig(SingletonModel):
+    directory = models.CharField(
+        _("NLX directory"), max_length=50, choices=NLXDirectories.choices, blank=True
+    )
+    outway = models.URLField(_("outway address"), blank=True)
```
sergei-maertens:
```suggestion
    outway = models.URLField(_("outway address"), blank=True, help_text=_("Example: http://my-outway.nlx:8080")
```

src/openzaak/config/templates/config/config_wizard.html
```diff
@@ -0,0 +1,36 @@
...
+
+{% block content %}
+
+<div class="px-3 py-3 pt-md-5 pb-md-4 mx-auto text-center">
+    <h1 class="display-4">Service configuratie: {{ wizard.steps.current| upper }}</h1>
+    <p class="lead">Step {{ wizard.steps.step1 }} of {{ wizard.steps.count }}</p>
+</div>
+
+<div class="container">
+    {% load i18n %}
```
sergei-maertens: already loaded at the top of the template :)

src/openzaak/config/urls.py
```diff
@@ -0,0 +1,8 @@
+from django.urls import path
+
+from .views import ConfigDetailView, ConfigWizardView
+
+urlpatterns = [
+    path("", ConfigDetailView.as_view(), name="config-detail"),
+    path("change", ConfigWizardView.as_view(), name="config-change"),
```
sergei-maertens:
```suggestion
    path("change/", ConfigWizardView.as_view(), name="config-change"),
```

src/openzaak/config/forms.py
annashamray: timeout is added

src/openzaak/config/templates/config/config_wizard.html
annashamray: fixed, thanks!

src/openzaak/config/models.py
```diff
@@ -11,7 +11,7 @@ class NLXConfig(SingletonModel):
     directory = models.CharField(
         _("NLX directory"), max_length=50, choices=NLXDirectories.choices, blank=True
     )
-    outway = models.URLField(_("outway address"), blank=True)
+    outway = models.URLField(_("outway address"), blank=True, help_text=_("Example: http://my-outway.nlx:8080")
```
ericherman: What about using an example domain URL here?

https://www.iana.org/domains/reserved

sergei-maertens: Not a bad suggestion - however, since outways are not running on publicly accessible domains, this might give an incorrect idea that the outway address should look like a publicly available domain? Currently we've been deploying NLX inway/outway in the `nlx` namespace in kubernetes clusters, and the port number is also very relevant, so I figured this is closer to the real situation. Thoughts?

ericherman: If it would be more clear to use an "nlx" domain, because the real-life versions will have that, then maybe that is better. If "nlx" becomes a top level domain, perhaps "example.nlx" would become a reserved name in that domain, thus maybe something like "my-outway.example.nlx"? But this is a small thing. Perhaps more important is probably to think about if "nlx" becomes a new TLD, how will that intersect with the kubernetes deployments.

sergei-maertens: in theory, `my-outway.nlx.svc.cluster.local` should be the DNS name that makes clear it's internal in the cluster, but Open Zaak also deploys on VMs without K8s, where the hostname is rather `nlx-outway` without TLD.

In the cluster I've been experimenting with, I also found that `my-outway.nlx` did work and the `my-outway.nlx.svc.cluster.local` did not - might be a version thing.

In the end - we expect the configurator to know where the NLX outway is hosted or at least know who to ask this. I'm not sure we should worry too much about `.nlx` becoming a TLD - even in a k8s the local DNS trumps the external DNS :thinking:

ericherman: I agree, this is not a big deal -- if there were an obvious "better" thing, then great, but if not, we move on.

Issue #318, opened 2019-12-09
ZaakTypen and IOTypen from different Catalogus in BesluitType M2M relations in admin


stevenbal:
**Describe the bug**
I have the following `ZaakType`n, `BesluitType`n and `InformatieObjectType`n in my catalogus:
![Screenshot from 2019-12-09 14-02-17](https://user-images.githubusercontent.com/29249171/70438201-97efaf80-1a8d-11ea-8ecf-daf55cc6d7b2.png)

When I navigate to the edit page for the `BesluitType` and I scroll down to the relations tab, I see the following for zaaktypen:
![Screenshot from 2019-12-09 14-08-58](https://user-images.githubusercontent.com/29249171/70438314-ce2d2f00-1a8d-11ea-96ea-c7ac5f7508dc.png)

The list for M2M relations between Besluittypen-Zaaktypen and Besluittypen-informatieobjecttypen shows the ZaakTypen and InformatieObjectTypen from all `Catalogi`, not just from the `Catalogus` which the `BesluitType` belongs to

**To Reproduce**
1. Create two Catalogi and add a zaaktype and informatieobjecttype to each
2. Create a Besluittype for one of the Catalogi and go to the edit page
3. Scroll down to the relations tab and look at the zaaktypen and informatieobjecttypen tabs

**Expected behavior**
Since there is validation in the API on these M2M relations that checks whether the types belong to the same `Catalogus`, I would expect the boxes for the M2M relations to only show types that belong to the same catalogus as the BesluitType that is being edited

EDIT: the same issue occurs for `zaaktype` on the `ZaakTypeInformatieObjectTypeAdmin`

joeribekker:
This is really annoying I can imagine but it's not a minor fix.

stevenbal:
reopened for IOTypen: https://github.com/open-zaak/open-zaak/pull/526#issuecomment-607617095

PR #526, opened 2020-03-27
Issue/besluittype admin relation filtering

stevenbal:
issue: https://github.com/open-zaak/open-zaak/issues/318


joeribekker:
@stevenbal I imagine we need this for Informatieobjecttype as well? :)

PR #529, opened 2020-03-30
Feature/eio inhoud validation

stevenbal:
I noticed that this validation was implemented in the reference implementation, but not yet in OZ (https://github.com/VNG-Realisatie/gemma-zaken/issues/679)


joeribekker:
Checked Travis-CI for status, it succeeds according to https://travis-ci.org/github/open-zaak/open-zaak/builds/668765781

PR #533, opened 2020-03-31
prevent saving autorizaties.enabled field

annashamray:


annashamray:
> Please add a regression test :-)

@sergei-maertens the test is added

PR #546, opened 2020-04-10
Issue/document add admin errors

### LINKED ISSUE #543: Adding a new document in the admin errors
**Describe the bug**

Going to the admin to add a document/informatieobject shows an error page when I click the "Add informatieobject" button - the form doesn't even render.

Sentry event: https://sentry.maykinmedia.nl/maykin-media/open-zaak-utrecht-test/issues/269068/?query=is%3Aunresolved

Running `latest` version of Open Zaak.

**To Reproduce**
1. Enter the admin
2. Click "Toevoegen" for Documenten
3. Observe the error

**Expected behavior**

The form to fill out the properties should be visible.

stevenbal:
Fixes #543

**Changes**

* use getattr to ensure the PrivateFileWidget for the EIO admin does not crash when creating a new EIO

### DIFF
--- a/src/openzaak/components/documenten/tests/admin/test_eio.py
+++ b/src/openzaak/components/documenten/tests/admin/test_eio.py
@@ -1,0 +1,49 @@
+from django.urls import reverse
+
+from django_webtest import WebTest
+from webtest import Upload
+
+from openzaak.accounts.tests.factories import SuperUserFactory
+from openzaak.components.catalogi.tests.factories import InformatieObjectTypeFactory
+
+from ...models import EnkelvoudigInformatieObject
+from ..factories import EnkelvoudigInformatieObjectCanonicalFactory
+
+
+class EnkelvoudigInformatieObjectAdminTests(WebTest):
+    @classmethod
+    def setUpTestData(cls):
+        cls.user = SuperUserFactory.create()
+
+    def setUp(self):
+        super().setUp()
+
+        self.app.set_user(self.user)
+
+    def test_add_informatieobject_page(self):
+        add_url = reverse("admin:documenten_enkelvoudiginformatieobject_add")
+
+        response = self.app.get(add_url)
+        self.assertEqual(response.status_code, 200)
+
+    def test_create_informatieobject_save(self):
+        informatieobjecttype = InformatieObjectTypeFactory.create(concept=False)
+        canonical = EnkelvoudigInformatieObjectCanonicalFactory.create(
+            latest_version=None
+        )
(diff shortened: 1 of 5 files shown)


src/openzaak/components/documenten/tests/admin/test_eio.py
```diff
@@ -0,0 +1,14 @@
+from django.test import TestCase
+from django.urls import reverse
+
+from openzaak.utils.tests import AdminTestMixin
+
+
+class EnkelvoudigInformatieObjectAdminTests(AdminTestMixin, TestCase):
+    heeft_alle_autorisaties = True
```
sergei-maertens, line 8: this attribute is normally only used for API tests, no?

src/openzaak/components/documenten/widgets.py
```diff
@@ -3,5 +3,7 @@
 
 class PrivateFileWidget(_PrivateFileWidget):
     def get_display_value(self, value):
-        instance = value.instance
-        return instance.bestandsnaam or super().get_display_value(value)
+        instance = getattr(value, "instance", None)
+        return getattr(instance, "bestandsnaam", None) or super().get_display_value(
+            value
+        )
```
sergei-maertens, line 9:
```suggestion
        if not value:
            return super().get_display_value(value)

        instance = value.instance
        return instance.bestandsnaam or super().get_display_value(value)
```

Isn't this cleaner, because `value` is `None` when adding a new document?

src/openzaak/components/documenten/widgets.py
```diff
@@ -3,5 +3,7 @@
 
 class PrivateFileWidget(_PrivateFileWidget):
     def get_display_value(self, value):
```
sergei-maertens, line 5:
```suggestion
from typing import Optional
from django.db.models.files import FieldFile

    def get_display_value(self, value: Optional[FieldFile]):
```

I _think_ it's a `FieldFile` object, you'd have to double check to get the right type hints :)

src/openzaak/components/documenten/widgets.py
stevenbal: yes you're right. separates the two cases more clearly

src/openzaak/components/documenten/tests/admin/test_eio.py
stevenbal: whoops, true, bit too much copypasta

stevenbal:
requires https://github.com/sergei-maertens/django-privates/pull/2 to be merged and version of django_privates to be bumped

PR #547, opened 2020-04-10
Added test for creating Zaaktype without referentieproces.link

### LINKED ISSUE #544: creating zaaktype with the api call without optional link attribue in referentieproces results in 500 error
**Describe the bug**
According to the documentation for creating a zaaktype with the vng api the link attribute is not required for referentieproces. When posting a json without this  attribute this results in a 500 error.

The following is part of the debug message:

> Exception Type: KeyError at /catalogi/api/v1/zaaktypen
Exception Value: 'link'
Request information:
USER: AnonymousUser

When using an empty string for the link the zaaktype is created successfully.

**To Reproduce**
1. post a zaaktype for the catalogi without a link attribute in referentieproces

**Expected behavior**
The link attribute should not be required to be present.

stevenbal:
Fixes #544 

**Changes**

* added a test for bug that occurred when creating a Zaaktype via the API without referentieproces.link (which should be optional)

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -72,5 +72,5 @@ unidecode==1.0.22         # via vng-api-common
 uritemplate==3.0.0        # via coreapi, drf-yasg
 urllib3==1.24.3           # via requests
 uwsgi==2.0.18
-vng-api-common==1.0.37
+vng-api-common==1.0.38
 zgw-consumers==0.8.1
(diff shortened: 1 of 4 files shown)


stevenbal:
requires https://github.com/VNG-Realisatie/vng-api-common/pull/126 to be merged first

PR #563, opened 2020-04-17
Fixes #553 -- made Eigenschap.specificatie required in admin

### LINKED ISSUE #553: Eigenschap specification is required, but doesn't appear to be a required field (not bold)
**Describe the bug**
When adding an Eigenschap, it seems like the specification is not a required field. But it apparently is: 
afbeelding

**To Reproduce**
1. Add an Eigenschap.
2. Save it without choosing a Specification.

**Expected behavior**
I expect the text "Specificatie van eigenschap:" to be bold, or the error not to appear.

stevenbal:
Fixes #553 

**Changes**

* set `blank=False` for Eigenschap.specificatie_van_eigenschap

### DIFF
--- a/src/openzaak/components/catalogi/models/eigenschap.py
+++ b/src/openzaak/components/catalogi/models/eigenschap.py
@@ -178,7 +178,7 @@ class Eigenschap(models.Model):
     specificatie_van_eigenschap = models.ForeignKey(
         "catalogi.EigenschapSpecificatie",
         verbose_name=_("specificatie van eigenschap"),
-        blank=True,
+        blank=False,
         null=True,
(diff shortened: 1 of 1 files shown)


src/openzaak/components/catalogi/models/eigenschap.py
```diff
@@ -178,7 +178,7 @@ class Eigenschap(models.Model):
     specificatie_van_eigenschap = models.ForeignKey(
         "catalogi.EigenschapSpecificatie",
         verbose_name=_("specificatie van eigenschap"),
-        blank=True,
+        blank=False,
```
sergei-maertens, line 181: Can you also clean up the `def clean method`?

PR #597, opened 2020-05-27
Add feature flag to skip Publication validation for *typen

### LINKED ISSUE #596: As a participant of a POC project, I want to be able to test zaaktypen while editing them
... so I don't have to go through the publish and re-configuration phase all the time.

**Description**

In a number of pilots/proof-of-concepts, Open Zaak is installed and configured with a number of zaaktypen. Other developers make use of these, but the zaaktype contents change frequently because of a very-short feedback-cycle (think one-two weeks total project time).

Tweaking zaaktypen, having to publish the new versions and configuring the new zaaktype URLs (for different version) hurts the velocity a lot.

A feature flag to bypass publication-validation fixes this, with a strong warning that you should not enable this in production. Bypassing the validation bypasses standards-compliance.

sergei-maertens:
Fixes #596 

**Changes**

* Added a model for application feature flags
    * added feature flag to allow unpublished zaaktypen/informatieobjecttypen/besluittypen to be used
* Added feature flag config to admin dashboard


joeribekker:
The code addresses the problem by allowing concept zaaktypen for test-purposes.

PR #600, opened 2020-05-28
Optimize zaak-list endpoint

### LINKED ISSUE #599: The zaak-list endpoint is too slow
**Describe the bug**

The endpoint `/zaken/api/v1/zaken` does not perform well enough. In the Gemeente Utrecht environment, the following call took 895ms - that's almost a second!

```http
GET http://localhost:12018/utrecht/openzaak-zaken/zaken?zaaktype=&identificatie=&bronorganisatie=&maximaleVertrouwelijkheidaanduiding= HTTP/1.1
```

**To Reproduce**
1. Create a considerable amount of zaken in the database (couple thousand should do)
2. Ensure that zaken have relations:
    - statussen
    - eigenschappen
    - gerelateerde zaken
    - resultaat
    - kenmerken

**Expected behavior**

The endpoint should return results fast, as this is a project aimed at production and performance.

sergei-maertens:
Fixes #599

**Changes**

* added django-silk to profile queries/requests in DRF (django debug toolbar does not support this)
* added whitenoise to support the django-silk interface with `DEBUG=no` (in DEBUG mode, Django keeps a log of queries and is slower in general)
* added a prefetch related to the zaak viewset
* added a select related in the zaak viewset

---

All testing was done on my local machine. A database dump from Gemeente Utrecht was loaded in my machine to compare the results (same dataset from #599). Absolute numbers between dev/proeftuin environment cannot be compared because of the big hardware differences. However, benchmarks were run to establish a baseline.

**Baseline - no changes**

The `/zaken/api/v1/zaken` endpoint was hit 10 times in total, with no changes in data.

| avg duration | minimal | maximal |
|---|---|---|
| 1477 (ms)             |  1261 (ms)    |  1876 (ms)

**Results after optimizations**

| avg duration | minimal | maximal |
|---|---|---|
| 1105 (ms)             |  969 (ms)    |  1405 (ms)

## Profile data (silk):

Action | Request time (ms) | # queries | query time (ms)
---|---|---|---
baseline | 1991 | 134 | 375
prefetch eigenschappen | 1589 | 35 | 130
`select_related` | 1267 | 14 | 85

Analysis of the queries shows that the main query to fetch the Zaken from the db takes about half of the total query time. The remaining queries are prefetch-queries or supporting/related to auth, with little room for improvement.

## Hardware impact

The main query taking 40ms is just on my machine - the query plan looks like this:

```
                                                                        QUERY PLAN                                                                         
-----------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.42..47.33 rows=100 width=2468) (actual time=0.168..0.718 rows=100 loops=1)
   ->  Nested Loop Left Join  (cost=0.42..1853.88 rows=3951 width=2468) (actual time=0.167..0.700 rows=100 loops=1)
         ->  Index Scan Backward using zaken_zaak_pkey on zaken_zaak  (cost=0.28..200.79 rows=3951 width=224) (actual time=0.035..0.067 rows=100 loops=1)
         ->  Index Scan using catalogi_zaaktype_pkey on catalogi_zaaktype  (cost=0.14..0.16 rows=1 width=2180) (actual time=0.001..0.001 rows=1 loops=100)
               Index Cond: (id = zaken_zaak._zaaktype_id)
 Planning Time: 7.062 ms
 Execution Time: 39.302 ms
(7 rows)
```

Running the same on the actual infrastructure shows

```
                                                                        QUERY PLAN                                                                        
----------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.42..25.27 rows=100 width=807) (actual time=0.035..0.313 rows=100 loops=1)
   ->  Nested Loop Left Join  (cost=0.42..981.47 rows=3948 width=807) (actual time=0.034..0.303 rows=100 loops=1)
         ->  Index Scan Backward using zaken_zaak_pkey on zaken_zaak  (cost=0.28..203.77 rows=3948 width=223) (actual time=0.014..0.028 rows=100 loops=1)
         ->  Index Scan using catalogi_zaaktype_pkey on catalogi_zaaktype  (cost=0.14..0.16 rows=1 width=520) (actual time=0.001..0.001 rows=1 loops=100)
               Index Cond: (zaken_zaak._zaaktype_id = id)
 Planning time: 0.413 ms
 Execution time: 8.201 ms
(7 rows)
```

The actual infrastructure runs queries about 5 times faster.

edit: the previous query times were off because of database caches (probably). Re-running show more realistic results.

### DIFF
--- a/src/openzaak/conf/dev.py
+++ b/src/openzaak/conf/dev.py
@@ -72,6 +72,13 @@ REST_FRAMEWORK["DEFAULT_RENDERER_CLASSES"] += (
     "rest_framework.renderers.BrowsableAPIRenderer",
 )
 
+#
+# DJANGO-SILK
+#
+if config("PROFILE", default=False):
+    INSTALLED_APPS += ["silk"]
+    MIDDLEWARE = ["silk.middleware.SilkyMiddleware"] + MIDDLEWARE
+
 warnings.filterwarnings(
     "error",
     r"DateTimeField .* received a naive datetime",
--- a/src/openzaak/conf/dev.py
+++ b/src/openzaak/conf/dev.py
@@ -78,6 +78,8 @@ REST_FRAMEWORK["DEFAULT_RENDERER_CLASSES"] += (
 if config("PROFILE", default=False):
     INSTALLED_APPS += ["silk"]
     MIDDLEWARE = ["silk.middleware.SilkyMiddleware"] + MIDDLEWARE
+    security_index = MIDDLEWARE.index("django.middleware.security.SecurityMiddleware")
+    MIDDLEWARE.insert(security_index + 1, "whitenoise.middleware.WhiteNoiseMiddleware")
 
 warnings.filterwarnings(
     "error",
--- a/src/openzaak/urls.py
+++ b/src/openzaak/urls.py
@@ -1,3 +1,4 @@
+from django.apps import apps
 from django.conf import settings
 from django.conf.urls.static import static
 from django.contrib import admin
@@ -45,3 +46,7 @@ if settings.DEBUG and "debug_toolbar" in settings.INSTALLED_APPS:
     import debug_toolbar
 
     urlpatterns += [path("__debug__/", include(debug_toolbar.urls))]
+
+
+if apps.is_installed("silk"):
+    urlpatterns += [path(r"silk/", include("silk.urls", namespace="silk"))]
--- a/src/openzaak/urls.py
+++ b/src/openzaak/urls.py
@@ -42,7 +42,7 @@ urlpatterns += static(
     settings.PRIVATE_MEDIA_URL, document_root=settings.PRIVATE_MEDIA_ROOT
 )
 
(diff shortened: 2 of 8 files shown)


src/openzaak/conf/dev.py
```diff
@@ -72,6 +72,15 @@
     "rest_framework.renderers.BrowsableAPIRenderer",
 )
 
+#
+# DJANGO-SILK
+#
+if config("PROFILE", default=False):
```
joeribekker, line 78: Can you add this variable to the relevant docs?

src/openzaak/urls.py
```diff
@@ -45,3 +46,7 @@
     import debug_toolbar
 
     urlpatterns += [path("__debug__/", include(debug_toolbar.urls))]
+
+
+if apps.is_installed("silk"):
```
joeribekker, line 51: Oh, nice. Should we also use this pattern for debug toolbar?

sergei-maertens: yes :-)

PR #632, opened 2020-06-26
Issue/cmis eio inhoud url

stevenbal:
Fixes https://github.com/open-zaak/open-zaak/pull/626#pullrequestreview-436733071

**Changes**

* fixed serialization of EIO.inhoud for CMIS_ENABLED
* changed EIO filtering by version to use the OZ version (e.g. 110) instead of the Alfresco version (e.g. 1.1)

### DIFF
--- a/src/openzaak/components/documenten/api/serializers.py
+++ b/src/openzaak/components/documenten/api/serializers.py
@@ -39,6 +39,7 @@ from ..models import (
     Gebruiksrechten,
     ObjectInformatieObject,
 )
+from ..utils import PrivateMediaStorageWithCMIS
 from .validators import InformatieObjectUniqueValidator, StatusValidator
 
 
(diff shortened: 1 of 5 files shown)


stevenbal:
@sergei-maertens I noticed that in the tests, filtering EIO by version seems to happen with strings (like here: https://github.com/open-zaak/open-zaak/blob/issue/cmis-eio-inhoud-url/src/openzaak/components/documenten/tests/test_enkelvoudiginformatieobject_cmis.py#L607). I assume these are the versions of the document in Alfresco?

Since this differs from the `version` attribute on the EIO instance, which version type should be returned by the serializer?

sergei-maertens:
> Since this differs from the `version` attribute on the EIO instance, which version number type should be returned by the EIO serializer for fields like `version` and `inhoud` (since this url can be filtered by version)?

@stevenbal per the API spec the version number must be an integer. Joeri told me that the version nummer apparently is tracked on the Alfresco metadata as well.

stevenbal:
> > Since this differs from the `version` attribute on the EIO instance, which version number type should be returned by the EIO serializer for fields like `version` and `inhoud` (since this url can be filtered by version)?
> 
> @stevenbal per the API spec the version number must be an integer. Joeri told me that the version nummer apparently is tracked on the Alfresco metadata as well.

Hmm okay, filtering by version now seems to work by using the Alfresco version (like "1.1"), so I'll look into fixing that first

PR #640, opened 2020-07-09
Add Helm files

rjzondervan:
This pull request adds helm files for OpenZaak. They might not be perfect yet, but we managed to install openzaak with them on the begrafenisplanner environment for Westfriesland.

The helm value files can be found in deployment/helm, the helm templates can be found in deployment/helm/templates


sergei-maertens:
Thanks for this contribution, we'll review this as soon as possible!

deployment/helm/templates/configmap.yaml
```diff
@@ -0,0 +1,17 @@
...
+  #namespace: {{ .Values.settings.env }}
+  labels:
+    app.kubernetes.io/name: {{ .Values.settings.name }}
+    app.kubernetes.io/part-of: {{ .Values.settings.name }}
+    helm.sh/chart: {{ include "chart" . }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+data:
+  allowed-hosts: {{ .Values.security.allowedHosts | quote }}
+  module: {{ .Values.settings.module }}
```
sergei-maertens, line 14: I'd rename this to `django_settings_module` to stay in line with common practices (and mimic the environment variable that applies this)

deployment/helm/templates/configmap.yaml
```diff
@@ -0,0 +1,17 @@
...
+  cache-default: {{ .Values.settings.name }}-cache:{{ .Values.redis.port }}/{{ .Values.redis.database }}
+  cache-axes: {{ .Values.settings.name }}-cache:{{ .Values.redis.port }}/{{ .Values.redis.database }}
+  debug: {{ .Values.settings.debug | quote }}
```
sergei-maertens, line 17: this should be left out - we want to make it as hard as possible to run insecure deployments

deployment/helm/templates/ingress.yaml
```diff
@@ -0,0 +1,49 @@
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  annotations:
+    # add an annotation indicating the issuer to use.
+    cert-manager.io/acme-challenge-type: http01
+    cert-manager.io/cluster-issuer: letsencrypt-prod
```
sergei-maertens, line 7: shouldn't that `prod` label be some kind of variable so you can deploy staging/production easily too? or am I misunderstanding how things work?

deployment/helm/templates/ingress.yaml
```diff
@@ -0,0 +1,49 @@
...
+                serviceName: {{ $.Values.settings.name }}
+                servicePort: 80
+    {{- else }}
+    - host: {{ $.Values.settings.subdomain }}.{{ $.Values.settings.env }}.{{ $.Values.settings.domain }}
+      http:
+        paths:
+            - backend:
+                serviceName: {{ $.Values.settings.name }}
+                servicePort: 80
+    {{- end }}
```
sergei-maertens, line 49: I would strongly advise not to have multiple domains pointing to the same instance, because you then get different URLs pointing to the same resource in the API, which in turns can break relation-filtering in other systems that save references via URL (as they should!)

deployment/helm/templates/nginx-service.yaml
```diff
@@ -0,0 +1,19 @@
...
+metadata:
+  name: {{ .Values.settings.name }}
+  labels:
+    app.kubernetes.io/name: {{ .Values.settings.name }}
+    app.kubernetes.io/part-of: {{ .Values.settings.name }}
+    helm.sh/chart: {{ include "chart" . }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+spec:
+  type: NodePort
```
sergei-maertens, line 12: this may conflict with other http servers on those nodes - the default is ClusterIP and intended to sit behind an ingress. Maybe make this a variable?

deployment/helm/templates/pvc.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: {{ .Values.volume.claimName }}
+spec:
+  storageClassName: nfs
```
sergei-maertens, line 6: that should definitely be configurable through a variable :-)

deployment/helm/values.yaml
```diff
@@ -0,0 +1,138 @@
...
+  name: openzaak
+  subdomain: openzaak
+  domain: zuid-drecht.nl
+  env: dev
+  module: openzaak.conf.docker
+  secret: 7y82z1%ef)k@f=u-_%ri71lk4=7jy^z5#@6d4*92=t@p0-yk5k
+  dsn:
+  version: latest
+  pullPolicy: Always
+  debug: 1
```
sergei-maertens, line 15: default values should be generic (e.g. use `gemeente.nl` instead of real domain names) _and_ default to secure values -> `debug: 0`, `version` pinned to a particular version (it will happen that `latest` is broken and if your cluster _then_ pulls a new version, that's going to be annoying)

Definitely don't put realistic secret keys in example values!

deployment/helm/values.yaml
```diff
@@ -0,0 +1,138 @@
...
+
+security:
+  allowedHosts: '*'
```
sergei-maertens, line 18: we advise against insecure defaults like this

deployment/helm/values.yaml
```diff
@@ -0,0 +1,138 @@
...
+
+email:
+  url: localhost
+  port: 867
+  user: email
+  password: '!changeMe!'
+
+nginx:
+  version: stable
+  maxUpload: 4Gi
```
sergei-maertens, line 28: variable defined but not used in the nginx configmap, not sure if that's intentional or forgotten

deployment/helm/values.yaml
```diff
@@ -0,0 +1,138 @@
...
+postgres:
+  url: localhost
+  port: 5432
+  user: openzaak
+  password: '!changeMe!'
+  database: openzaak
+
+redis:
+  port: 6379
+  version: 5.0
```
sergei-maertens, line 44: careful - YAML will treat this as a float, better to explicitly quote?

deployment/helm/values.yaml
```diff
@@ -0,0 +1,138 @@
...
+  database: openzaak
```
sergei-maertens, line 45: didn't know that redis can work with named databases, afaik it's a number from 0-15 by default?

sergei-maertens:
hi @rjzondervan - we discussed this long-standing PR during the technical steering group meeting today.

We noted that we don't have people who can maintain Helm charts once they're added to Open Zaak - would you be up for maintaining this once it gets merged?

Additionally, we haven't heard back since the PR review was done. Is there anything we can help with to process the remarks?

sergei-maertens:
I'm closing this PR in favour of the work done by @bartjkdp in https://github.com/open-zaak/charts

Issue #610, opened 2020-06-03
View Config page does not respect new services in 1.2


sergei-maertens:
Hinted at by Ramon Rockx on Slack, it seems that the refactor into zgw-consumers services breaks the notifications integration:

* view config relies on the singleton config & still fetches credentials from `vng_api_common.APICredential`: https://github.com/open-zaak/open-zaak/blob/dbe7e916ea1b18325c72f48fcd205b2e6f41080a/src/openzaak/utils/views.py#L46
* the API uses the `vng_api_common` mixins (https://github.com/open-zaak/open-zaak/blob/dbe7e916ea1b18325c72f48fcd205b2e6f41080a/src/openzaak/components/zaken/api/viewsets.py#L23) which definitely do not make use of zgw-consumers

sergei-maertens:
There is a workaround by adding the External API credentials. The model is hidden from the dashboard, but you can access it directly by URL: http://localhost:8000/admin/vng_api_common/apicredential/ (replace the host with your own domain of course)

PR #655, opened 2020-07-28
Add database migrations to move APICredential setup to Services

sergei-maertens:
Related to #610 

**Changes**

In 20c654be1b8de8e3e9376213fcd9bf451a6a0a09 the config was changed to
get a client from the zgw_consumers.Service model rather than using
vng_api_common.APICredential. However, existing configuration for
people upgrading was not automatically migrated.

This commit adds migrations to automatically migrate the credentials to
zgw_consumers.Service.


sergei-maertens:
Yeah, good catch!

src/openzaak/selectielijst/migrations/0003_move_config_to_service_model.py
```diff
@@ -0,0 +1,67 @@
...
+    Service.objects.create(
+        label=credentials.label if credentials else "VNG Selectielijst",
+        api_type=APITypes.orc,
+        api_root=config.api_root,
+        oas=f"{config.api_root}schema/openapi.yaml",
+        client_id=credentials.client_id if credentials else "",
+        secret=credentials.secret if credentials else "",
+        user_id=credentials.user_id if credentials else "",
+        user_representation=credentials.user_representation if credentials else "",
+        auth_type=AuthTypes.zgw if credentials else AuthTypes.no_auth,
```
joeribekker, line 30: This is always no-auth I think? Even if a different selectielijst is used theres no way to tell if it uses ZGW auth.

sergei-maertens: True, was trying to accommodate for any edge cases, but it makes more sense to just leave it `no_auth` and let the end-user amend if needed.

sergei-maertens:
@joeribekker tests should be fixed, and migration changes made (`nrc` type and not assuming the auth scheme for selectielijst)

PR #672, opened 2020-08-03
Add SPDX license identifiers, copyright to src/

### LINKED ISSUE #674: All source code files in the codebase SHOULD include a copyright notice and a license header.
**Is your feature request related to a problem? Please describe.**
Not all source code files contain license or copyright information, making it less easy to copy into other works.

This is a requirement in the Standard for Public Code.
See also: issue #673 and https://github.com/open-zaak/open-zaak/blob/master/docs/introduction/open-source/public_code.md

**Describe the solution you'd like**
An SPDX license identifier and copyright line in each file.

**Describe alternatives you've considered**
A single reference in the top level directory is already present and could be considered good enough.

ericherman:
Fixes #674 

**Changes**

Addresses gap in docs/introduction/open-source/public_code.md

"All source code files in the codebase SHOULD include a copyright
notice and a license header."

From the Standard for Public Code:
	https://standard.publiccode.net/criteria/open-licenses.html

SPDX is aligned with README.md for license and copyright holder.
Year range is based upon git history.

Change was generated with a naive script:
https://github.com/ericherman/code-snips/blob/master/shell/spdx.sh
and then adjusted to deal with "sha-bang" files.


sergei-maertens:
Maybe a small "cleanup" is to change the HTML comments `<!-- -->` into `{% comment %} ... {% endcomment %}`, so the comments don't end up in the templated HTML. As far as I know, all `.html` files are actually Django templates.

ericherman:
> I think we should also have a CI job to check that the license header is present (probably best to limit it to files included in the diff).

That would be a very nice addition.

> One other remark that I'm not sure about but that CI will surely catch - the HTML comments in Django template files before `{% extends %}` tags might trip the template rendering because the `{% extends %}` tag should be the first tag in the template. Not sure if Django cares about the HTML comments or not.

I would hope tests to fail if this is the case, if we find something breaks but that CI does not catch, let's find a way to add that to the CI rather than expect any contributor to remember that.

sergei-maertens:
> I would hope tests to fail if this is the case, if we find something breaks but that CI does not catch, let's find a way to add that to the CI rather than expect any contributor to remember that.

and they do so now :grimacing:

ericherman:
> > I would hope tests to fail if this is the case, if we find something breaks but that CI does not catch, let's find a way to add that to the CI rather than expect any contributor to remember that.
> 
> and they do so now grimacing

So, retreat to the html-style comments then?

sergei-maertens:
@ericherman yeah, the HTML style comments is the quickest solution to get this in

Ideally I'd see those replaced by django comments for reasons mentioned before, but then they'd have to come _after_ the `{% extends ... %}` line (which is always a single line if it's present).

sergei-maertens:
I've restarted the build as it's failing because of a Heisenbug somewhere

PR #676, opened 2020-08-03
Support CMIS 1.0 with Web Service binding (SOAP)

SilviaAmAm:
Taiga issue 177

**Changes**

- The binding used by the CMIS client can be configured to either web service (SOAP) or browser.
- The documentation regarding the installation with the CMIS adapter was updated.
- An extra Travis job was added to run the CMIS tests using the web service binding.


docker-compose.travis-cmis.yml
```diff
@@ -11,7 +11,7 @@ version: "2"
 services:
     alfresco:
         image: alfresco/alfresco-content-repository-community:6.1.2-ga
-        mem_limit: 1500m
+        mem_limit: 2500m
```
sergei-maertens, line 14: Looking at the Travis [build environments](https://docs.travis-ci.com/user/reference/overview/#virtualisation-environment-vs-operating-system), it looks like the VMs have 7.5GB of memory.

I think it wouldn't hurt to just make this 4GB for Alfresco, then there's still 3.5GB left over for the rest which should be plenty.

```suggestion
        mem_limit: 4g
```

joeribekker (approved):

PR #728, opened 2020-10-01
Fixes 584 download non existing EIO causes 500 error

### LINKED ISSUE #584: Trying to download a nonexistent EIO gives a 500 error
**Describe the bug**
If I am trying to download the content of an `EnkelvoudigInformatieObject` that no longer exists, I get a 500 error

**To Reproduce**
1. Create an EnkelvoudigInformatieObject and store the URL
2. Delete the EIO
3. Do a GET request on `{{eio_url}}/download`

**Expected behavior**
Since a GET request on the EIO itself results in a 404, I would expect an attempt to download the content of that nonexistent EIO to give a 404 as well

related issue in gemma-zaken: https://github.com/VNG-Realisatie/gemma-zaken/issues/1617

SilviaAmAm:
Fixes #584 

**Changes**
Added a check for when strings are passed to the `BinaryFileRenderer`

In the case where there is an error response and the renderer is `BinaryFileRenderer`, the response content is a string rather than binary data. If a renderer returns a string, the `Response` class attempts to encode it using the specified charset. 
In the `BinaryFileRenderer` the charset is set to `None`, so this caused an error.
Now, inside `BinaryFileRenderer` there is a check for string responses, which are then encoded to UTF-8.

### DIFF
--- a/src/openzaak/components/documenten/api/renderers.py
+++ b/src/openzaak/components/documenten/api/renderers.py
@@ -10,4 +10,9 @@ class BinaryFileRenderer(BaseRenderer):
     render_style = "binary"
 
     def render(self, data, media_type=None, renderer_context=None):
+        # When trying to download a non-existing file, `data` contains a string instead of binary data.
+        # The string needs to be encoded as UTF-8, or it will be coerced into a bytestring using self.charset,
+        # which will cause an error (issue #584)
+        if isinstance(data, str):
+            return data.encode("utf-8")
         return data
--- a/src/openzaak/components/documenten/tests/test_enkelvoudiginformatieobject.py
+++ b/src/openzaak/components/documenten/tests/test_enkelvoudiginformatieobject.py
@@ -411,6 +411,23 @@ class EnkelvoudigInformatieObjectAPITests(JWTAuthMixin, APITestCase):
         # Test response
         self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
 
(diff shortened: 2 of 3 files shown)


stevenbal (approved):

src/openzaak/components/documenten/api/renderers.py
```diff
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: EUPL-1.2
 # Copyright (C) 2019 - 2020 Dimpact
+import six
```
sergei-maertens, line 3: no reason to use `six`, Python 2 is dead an burried

src/openzaak/components/documenten/api/renderers.py
```diff
@@ -10,4 +11,9 @@ class BinaryFileRenderer(BaseRenderer):
     render_style = "binary"
 
     def render(self, data, media_type=None, renderer_context=None):
+        # When trying to download a non-existing file, `data` contains a string instead of binary data.
+        # The string needs to be encoded as UTF-8, or it will be coerced into a bytestring using self.charset,
+        # which will cause an error (issue #584)
+        if isinstance(data, six.text_type):
```
sergei-maertens, line 17:
```suggestion
        if isinstance(data, str):
```

sergei-maertens (changes requested):
Can you also rebase on master and resolve the merge conflict please?

sergei-maertens (approved):

PR #737, opened 2020-10-09
POC of batch requests endpoint

sergei-maertens:


sergei-maertens:
Not pursuing this at this time.

PR #746, opened 2020-10-21
Update members and specify typical roles

joeribekker:
**Changes**

As per previous meeting, the members and desired roles to take place in the product steering team have been refined as proposal.


Hugo-ter-Doest (approved):
Looks good

PR #751, opened 2020-10-22
Fixes #713 Removing a zaaktype removes the related autorisatie

### LINKED ISSUE #713: Deleting a catalogue (with concept zaaktypes) when permissions have been defined for "all current and future zaaktypes" breaks the zaken list endpoint
**Describe the bug**

The application/authorization machinery allows applications to receive permissions for all current and future zaaktypen (or IOT/BT). This installs an `AutorisatieSpec`.

If you then create zaaktypen in a catalogus without publishing them, an Autorisatie is created. Deleting the zaaktype again does not delete the attached Autorisatie, which subsequently is used for permission checks/server-side filtering of Zaken from that zaaktype.

Zaaktypes that are deleted, should also delete related Autorisatie objects.

**To Reproduce**
1. Create an application with authorizations `zaken:lezen` on all current and future zaaktypes
2. Create a catalogus with a zaaktype - unpublished
3. Delete the zaaktype
4. Make an API call with the credentials of the created application to `/zaken/api/v1/zaken`
5. Observe a server error related to django-loose-fk

**Expected behavior**

The API should not return a server error, but appropriately filtered data.

SilviaAmAm:
Fixes #713 

**Changes**
When a `ZaakType` is removed, now there is a receiver for the post delete signal which checks for existing `Autorisaties` and removes them.

A management command was also added to sync existing `Autorisaties` for `ZaakType` instances that have previously been deleted.

### DIFF
--- a/src/openzaak/components/autorisaties/management/commands/sync_autorisaties.py
+++ b/src/openzaak/components/autorisaties/management/commands/sync_autorisaties.py
@@ -1,0 +1,37 @@
+# SPDX-License-Identifier: EUPL-1.2
+# Copyright (C) 2019 - 2020 Dimpact
+from urllib.parse import urlparse
+
+from django.apps import apps
+from django.core.exceptions import ObjectDoesNotExist
+from django.core.management import BaseCommand
+from django.utils.translation import ugettext_lazy as _
+
+from django_loose_fk.loaders import BaseLoader
+from vng_api_common.utils import get_resource_for_path
+
+Autorisatie = apps.get_model("authorizations", "Autorisatie")
+is_local_url = BaseLoader().is_local_url
+
+
+class Command(BaseCommand):
+    help = _(
+        "Remove any existing authorisations related to zaaktypen that have been removed"
+    )
+
+    def handle(self, *args, **options):
+
+        to_delete = []
+
+        for autorisatie in Autorisatie.objects.all():
+            if not is_local_url(autorisatie.zaaktype):
+                continue
+
+            parsed = urlparse(autorisatie.zaaktype)
+
+            try:
+                get_resource_for_path(parsed.path)
+            except ObjectDoesNotExist:
+                to_delete.append(autorisatie)
+
+        Autorisatie.objects.filter(pk__in=[obj.pk for obj in to_delete]).delete()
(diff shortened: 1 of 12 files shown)


src/openzaak/components/autorisaties/management/commands/sync_autorisaties.py
```diff
@@ -0,0 +1,37 @@
...
+    help = _(
+        "Remove any existing authorisations related to zaaktypen that have been removed"
+    )
+
+    def handle(self, *args, **options):
+
+        to_delete = []
+
+        for autorisatie in Autorisatie.objects.all():
+            if not is_local_url(autorisatie.zaaktype):
```
sergei-maertens, line 27: this does not only affect `zaaktype`, but also `informatieobjecttype` and `besluittype`

src/openzaak/components/catalogi/signals.py
```diff
@@ -0,0 +1,31 @@
...
+
+from openzaak.components.catalogi.models.zaaktype import ZaakType
+
+logger = logging.getLogger(__name__)
+
+
+@receiver(post_delete, dispatch_uid="catalogi.sync_autorisaties")
+def sync_autorisaties(
+    sender: ModelBase, signal: ModelSignal, instance: ZaakType, **kwargs
+) -> None:
```
sergei-maertens, line 21: Technically `instance` is a `Model` instance, as it will also fire for other models. You can limit the senders though: https://docs.djangoproject.com/en/2.2/topics/signals/#connecting-to-signals-sent-by-specific-senders

Additionally, this applies not only for `ZaakType`, but also for `InformatieObjectType` and `BesluitType`

src/openzaak/components/catalogi/signals.py
```diff
@@ -0,0 +1,31 @@
...
+    logger.debug("Received signal %r, from sender %r", signal, sender)
+
+    if sender is not ZaakType:
+        return
+
+    instance_path = instance.get_absolute_api_url()
+    site = Site.objects.get_current()
+    protocol = "https" if getattr(settings, "IS_HTTPS", True) else "http"
```
sergei-maertens, line 29:
```suggestion
    protocol = "https" settings.IS_HTTPS else "http"
```

Setting is defined in `base.py`, so the attribute always exists.

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #766, opened 2020-11-06
More performant CMIS operation by using differently generated Document.identificatie

### LINKED ISSUE #762: Change the auto-generated identificaties for documents when CMIS is in use
When a new document is created without a specified `identificatie`, a human readable `identificatie` is created.

What this does is:
- Takes the name of the model (ex: `DOCUMENT`)
- Takes the year of when it was created (ex: `2020`)
- Checks among existing documents if there are any other with identificatie `DOCUMENT-2020-*` and then based on how many are found it assigns a `identificatie` of format ex: `DOCUMENT-2020-0000000004`.

The filtering on `identificatie` uses a SQL query with the `LIKE` word. This is not supported by all DMSs. 

*To change*:
When the CMIS adapter is used, if no `identificatie` is specified, assign a value equal to the document `uuid`.

joeribekker:
Fixes #762

**Changes**

More performant CMIS operation by using differently generated Document.identificatie

### DIFF
--- a/docs/installation/config/cmis.rst
+++ b/docs/installation/config/cmis.rst
@@ -4,23 +4,23 @@ CMIS adapter
 ============
 
 In a default installation of Open Zaak, any document created through the
-`Documenten API`_ are stored on disk and their metadata is stored in the 
-database. However, it is also possible to store these documents in a Document 
+`Documenten API`_ are stored on disk and their metadata is stored in the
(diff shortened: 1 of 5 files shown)


joeribekker:
@SilviaAmAm Can you document that when using the CMIS adapter, identificatie is done different.

sergei-maertens (approved):

PR #779, opened 2020-11-24
Fixing duplicate ObjectInformatieObject creation with CMIS

### LINKED ISSUE #778: Duplicate ObjectInformatieObjects created
**Describe the bug**
If the CMIS-adapter is enabled, when a POST request is done to `/zaken/api/v1/zaakinformatieobjecten` to create a `ZaakInformatieObject` (ZIO), this function is called:
https://github.com/open-zaak/open-zaak/blob/master/src/openzaak/components/zaken/api/serializers/zaken.py#L552

What it does (in the CMIS case) is:

1. Create the ZIO. A `post_save` signal is received by https://github.com/open-zaak/open-zaak/blob/master/src/openzaak/components/documenten/signals.py#L26 which creates an `ObjectInformatieObject` (OIO).
2.  Checks whether the informatieobject related to the zaak by the new ZIO is internal or external. Since the informatieobject is in the DMS, it is considered external.  
3.  Proceeds to create the remote OIO. Since the OIO exists already in Alfresco, a duplicate is created (and a spurious copy of the informatieobject along with it).

However, if the CMIS-adapter is disabled and a ZIO is created with a 'real' external informatieobject, this is what happens:

1. Create the ZIO. A `post_save` signal is received by https://github.com/open-zaak/open-zaak/blob/master/src/openzaak/components/documenten/signals.py#L26. Here the OIO is NOT created. This is because the `zio.informatieobject` is of class 'VirtualEnkelvoudigInformatieObject' instead of 'EnkelvoudigInformatieObject' (https://github.com/open-zaak/open-zaak/blob/master/src/openzaak/components/documenten/query/django.py#L100).
2. Checks whether the informatieobject related to the zaak by the new ZIO is internal or external. It is external.  
3. Creates the remote OIO. 

The problem that is that in the CMIS case, in step 1 the document is considered local, but in step 2 it is considered external, so an extra OIO is created.

**To Reproduce**
Take test https://github.com/open-zaak/open-zaak/blob/master/src/openzaak/components/zaken/tests/test_zaakinformatieobjecten_cmis.py#L41 and add the following 2 lines at the end of the test:

```
self.assertEqual(ObjectInformatieObject.objects.count(), 1)
self.assertEqual(EnkelvoudigInformatieObject.objects.count(), 1)
```

**Expected behavior**
There should be only one OIO created when a ZIO is created.

SilviaAmAm:
Fixes #778 

**Changes**
When CMIS is enabled, the `ZaakInformatieObjectSerializer` and `BesluitInformatieObjectSerializer` method `create()` returns after creating a new object in the same way as for local `ZaakInformatieObject` and `BesluitInformatieObject`.

### DIFF
--- a/src/openzaak/components/besluiten/api/serializers.py
+++ b/src/openzaak/components/besluiten/api/serializers.py
@@ -194,8 +194,8 @@ class BesluitInformatieObjectSerializer(serializers.HyperlinkedModelSerializer):
         with transaction.atomic():
             bio = super().create(validated_data)
 
-        # local FK - nothing to do -> our signals create the OIO
-        if bio.informatieobject.pk:
+        # local FK or CMIS - nothing to do -> our signals create the OIO
+        if bio.informatieobject.pk or settings.CMIS_ENABLED:
             return bio
 
         # we know that we got valid URLs in the initial data
--- a/src/openzaak/components/zaken/api/serializers/zaken.py
+++ b/src/openzaak/components/zaken/api/serializers/zaken.py
@@ -553,8 +553,8 @@ class ZaakInformatieObjectSerializer(serializers.HyperlinkedModelSerializer):
         with transaction.atomic():
             zio = super().create(validated_data)
 
-        # local FK - nothing to do -> our signals create the OIO
-        if zio.informatieobject.pk:
+        # local FK or CMIS - nothing to do -> our signals create the OIO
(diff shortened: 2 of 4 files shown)


joeribekker (approved):
Looks good.

sergei-maertens (approved):

PR #786, opened 2020-12-04
Documentation on how to disable SSL/TLS certificate verification

SilviaAmAm:

### DIFF
--- a/docs/installation/config/cmis.rst
+++ b/docs/installation/config/cmis.rst
@@ -159,3 +159,4 @@ If the CMIS adapter is disabled, the procedure to automatically generate the ide
 3. The new document is given an identification field with a unique number that is different from those of all the other documents. This would for example be ``DOCUMENT-2020-0000000004``.
 
 The search done in point 2. requires an SQL LIKE clause, which is not supported by all DMSs. For this reason, if the CMIS adapter is in use, the automatically generated identification field will be equal to the document UUID.
+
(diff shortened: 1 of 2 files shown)


docs/installation/config/cmis.rst
```diff
@@ -159,3 +159,11 @@ If the CMIS adapter is disabled, the procedure to automatically generate the ide
 3. The new document is given an identification field with a unique number that is different from those of all the other documents. This would for example be ``DOCUMENT-2020-0000000004``.
 
 The search done in point 2. requires an SQL LIKE clause, which is not supported by all DMSs. For this reason, if the CMIS adapter is in use, the automatically generated identification field will be equal to the document UUID.
+
+Additional notes on TLS
+-----------------------
+
+In order to disable SSL/TLS certificate verification, the environment variable ``CURL_CA_BUNDLE`` needs to be set to an empty string.
```
sergei-maertens, line 166: This sounds like a terrible idea that will disable _all_ TLS certificate validation, even for other services such as a catalogi API. If this should only be restricted to CMIS things, then it should be monkeypatched instead with a CMIS-specific configuration variable.

sergei-maertens (changes requested):

docs/installation/config/cmis.rst
joeribekker: Agree, @SilviaAmAm can you move this to the env config: https://open-zaak.readthedocs.io/en/latest/installation/config/env_config.html in similar format as other env vars with aproppriate warning that this is a bad idea.

joeribekker (changes requested):

sergei-maertens (approved):

PR #808, opened 2021-01-06
CMIS URL mapping

### LINKED ISSUE #828: CMIS-adapter URL mapping
See issue https://github.com/open-zaak/cmis-adapter/issues/37 in the cmis-adapter

SilviaAmAm:
Fixes #828 
Related to issue 37 in cmis-adapter (https://github.com/open-zaak/cmis-adapter/issues/37)

**Changes**
- Added the setting to turn on/off the URL shrinking
- Added step to the CI matrix
- Updated the tests
- Added exception catching to handle errors returned from the CMIS-adapter
- Added some documentation about the new settings

(The PR is currently marked as WIP because the version of the CMIS-adapter is currently pinned as a git commit)


joeribekker (approved):
Ok.

If we get any more of these DMS specific things, we should really consider containing the CMIS-adapter more (tests with openzaak and docs) in its own repo.

sergei-maertens (approved):
The view mixin *may* affect the schema generation in the event that not a full ModelViewSet is used. It's possible that now methods are defined that weren't before (`list`, `retrieve`, `destroy` etc....)

One possible solution would be to define a meta class that inspects the other base classes to check if the methods exist, and only then add the decorator to them. Not considering this a merge blocker at the time, but for future reference :-)

PR #821, opened 2021-01-14
Fixes #820 - Filtering ZIOs on informatieobject with CMIS enabled

### LINKED ISSUE #820: Filtering ZIO on informatieobject with CMIS enabled doesnt work
**Describe the bug**
Filtering `ZaakInformatieObject` on `informatieobject` when CMIS is enabled returns `ZaakInformatieObject` that are not related to the requested document.

The tests for filtering have the following pattern (both in the case of CMIS enabled and disabled):

1. Create a document
2. Create a zaak
3. Relate the document and the zaak by creating a ZaakInformatieObject
4. List the `ZaakInformatieObject` in the API filtering on the document created in 1.

Since there is only 1 `ZaakInformatieObject`, the test passes.

**To Reproduce**
[This](https://github.com/open-zaak/open-zaak/blob/master/src/openzaak/components/zaken/tests/test_zaakinformatieobjecten_cmis.py#L209) test should be modified as follows:

```python
def test_filter_by_informatieobject(self):
    self.create_zaak_besluit_services()
    zaak = self.create_zaak()

    # Create two ZIOs
    eio1 = EnkelvoudigInformatieObjectFactory.create()
    eio1_url = f"http://example.com{reverse(eio1)}"
    self.adapter.get(eio1_url, json=serialise_eio(eio1, eio1_url))

    ZaakInformatieObjectFactory.create(informatieobject=eio1_url, zaak=zaak)

    eio2 = EnkelvoudigInformatieObjectFactory.create()
    eio2_url = f"http://example.com{reverse(eio2)}"
    self.adapter.get(eio2_url, json=serialise_eio(eio2, eio2_url))

    ZaakInformatieObjectFactory.create(informatieobject=eio2_url, zaak=zaak)

    zio_list_url = reverse("zaakinformatieobject-list")

    # Test that only 1 of the 2 ZIOs is returned
    response = self.client.get(
        zio_list_url, {"informatieobject": eio1_url}, HTTP_HOST="example.com",
    )

    self.assertEqual(response.status_code, 200)
    self.assertEqual(len(response.data), 1)
    self.assertEqual(response.data[0]["informatieobject"], eio1_url)
```

**Expected behavior**
When filtering the `ZaakInformatieObject` on the `informatieobject` only the objects that relate to the requested document should be returned.

SilviaAmAm:
Fixes #820

**Cause of the problem**

The `ZaakInformatieObject` have a filterset class `ZaakInformatieObjectFilter` which specifies how to filter on the FKorURLField `informatieobject`. 
When the dictionary of filters is generated (here: https://github.com/maykinmedia/django-loose-fk/blob/develop/django_loose_fk/filters.py#L58) the document is considered 'local' even though it is in the DMS. Because of this, the ZIOs are filtered using the canonical object, but this doesn't work if CMIS is enabled, so the wrong ZIOs are returned.

**Changes**

I overwrote the `filter()` method of the `ZaakInformatieObjectFilter` class to take into account the CMIS case.

### DIFF
--- a/src/openzaak/components/zaken/api/filters.py
+++ b/src/openzaak/components/zaken/api/filters.py
@@ -1,7 +1,12 @@
 # SPDX-License-Identifier: EUPL-1.2
 # Copyright (C) 2019 - 2020 Dimpact
+from urllib.parse import urlparse
+
+from django.conf import settings
+
 from django_filters import filters
 from django_loose_fk.filters import FkOrUrlFieldFilter
+from django_loose_fk.utils import get_resource_for_path
 from vng_api_common.filtersets import FilterSet
 from vng_api_common.utils import get_help_text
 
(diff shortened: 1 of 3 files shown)


joeribekker (commented):
@sergei-maertens Altho this works, shouldn't we allow for some configuration here, that decides what is local and what not?

Maybe a setting `FK_OR_URL_IS_LOCAL_FUNC` where you can defined a callable, or so. I see similar functionality in `BaseLoader.is_local_url` and the comparison `local = parsed.netloc == host` in 2 places.

sergei-maertens:
> Maybe a setting FK_OR_URL_IS_LOCAL_FUNC where you can defined a callable, or so. I see similar functionality in BaseLoader.is_local_url and the comparison local = parsed.netloc == host in 2 places.

it's tricky, because the `BaseLoader.is_local_url` doesn't have access to the request object, so it falls back to the `settings.ALLOWED_HOSTS`. If you can compare `parsed.netloc` with `host`, that's always going to be the better option because you're looking at a live request.

@SilviaAmAm can you rebase this PR on master?

sergei-maertens (approved):

sergei-maertens:
I've created https://github.com/maykinmedia/django-loose-fk/issues/20 for a possible refactor

Issue #819, opened 2021-01-13
Transaction management for zaakinformatieobject create is broken when using an external DRC


sergei-maertens:
**Describe the bug**

The dance to relate documents to zaken requires the remote DRC to be able to see that zaak & documented are 'already' related in Zaken API side of things. However, this currently doesn't work, because the transaction that persists that to database is still pending. The upper `create` call in the viewset is marked as `@transaction.atomic` because of the `NotificationViewSetMixin` which requires the transactions to be atomic.

**To Reproduce**
1. Create a zaak in Open Zaak
2. Create a document in an external DRC, for example, the reference implementation from VNG
3. Make the `POST` call to Open Zaak to relate zaak & document
4. Observe an unexpected validation error

If you put a debugger breakpoint in the ZIO serializer as part of the `except:` block, you can pause the execution and make the API call that the DRC makes to verify that the ZIO is not yet committed to the database, and therefor invisible.

**Expected behavior**

The call completes correctly with an HTTP 201.

Possible solution is to store the pending ZIO in a cache that's outside of the database transaction, and also check that cache in the ZIO list implementation. This cache could have a short expiry time, such as 10 seconds.

Refactoring to remove the outer transaction blocks seems like a massive undertaking that requires changes to `vng_api_common`. It might be worth it, as the notification integration shouldn't really mark the `@transaction.atomic`, but rather hook into `transaction.on_commit` to send the notification. That would open us up to manual transaction management in this particular viewset/action.

This is affecting the ZGW standard Postman test-suite and effectively makes Open Zaak unusable with an external DRC that performs the strict validations as the standard prescribes them.

PR #827, opened 2021-01-18
Fix Open Zaak handling of remote DRCs

sergei-maertens:
Relates to #819

While testing the remote DRC integration with the DRC reference implementation service, it turns out that there were more defects in Open Zaak that prevented this from working correctly.

All the issues were uncovered by the test suite on api-test.nl, in combination with the `vngr/gemma-drc:1.0.1.post0` reference implementation of a DRC.

**Changes**

* added regression test for (partial) update validation of `ZaakInformatieObject` where the remote Documents API url does not change
* fixed validation bug where instance comparison happened rather than URL comparison if CMIS is not enabled
* added regression test and fix for remote OIO `DELETE` call on cascading zaak-destroy operations
* added regression test and fix for `Besluit.zaak` being output as `null` in the API, which should be empty string instead according to the API spec

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -24,7 +24,7 @@ django-extra-fields==1.2.4  # via -r requirements/base.in
 django-extra-views==0.13.0  # via -r requirements/base.in
 django-filter==2.0        # via -r requirements/base.in, django-loose-fk, vng-api-common
 django-ipware==2.1.0      # via django-axes
-django-loose-fk==0.7.1    # via -r requirements/base.in
+django-loose-fk==0.7.3    # via -r requirements/base.in
 django-markup==1.3        # via -r requirements/base.in
 django-ordered-model==2.1.0  # via django-admin-index
 django-privates==1.2.1    # via -r requirements/base.in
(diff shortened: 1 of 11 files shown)


SilviaAmAm (approved):
The changes look good to me. 

I wonder if it might be good to have a comment outlining the different scenarios:
1. CMIS case: check on url
2. External document: check on url
3. Internal document: retrieve the canonical object and check its the same as before.

Just because I had to run the code in the different scenarios to remind myself what happens in each case. But maybe it's obvious enough from the code if you know it well!

sergei-maertens:
> The changes look good to me.
> 
> I wonder if it might be good to have a comment outlining the different scenarios:
> 
> 1. CMIS case: check on url
> 2. External document: check on url
> 3. Internal document: retrieve the canonical object and check its the same as before.
> 
> Just because I had to run the code in the different scenarios to remind myself what happens in each case. But maybe it's obvious enough from the code if you know it well!

The code is non-obvious and pretty complicated. It would definitely benefit from some more comments yeah, I'll make sure to add these before checking in the PR.

PR #836, opened 2021-02-01
Add Helm chart for Open Zaak

bartjkdp:
**Changes**

This PR contains a Helm chart for Open Zaak. It follows the structure of `helm create` as closely as possible.

- PostgreSQL and Redis are dependent charts and can be deactivated with `tags.*`
- Contains a smoke-test with wget

This could live in a separate repository as well (e.g. open-zaak/charts). Would also be a great place to configure the [chart-releaser-action](https://github.com/helm/chart-releaser-action).

Tested this chart with multiple end-users, currently hosted [here](https://github.com/bartjkdp/open-zaak-charts).

### DIFF
--- a/deployment/helm/.helmignore
+++ b/deployment/helm/.helmignore
@@ -1,0 +1,23 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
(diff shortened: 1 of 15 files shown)


bartjkdp:
Moved to: https://github.com/open-zaak/charts

PR #837, opened 2021-02-02
Build correct STATIC_URL and MEDIA_URL

### LINKED ISSUE #741: Statics don't show correctly on a default Open Notificaties docker
**Describe the bug**
Seems the URL path is resolved to //static or so. This is caused by SUB_PATH in Docker-compose. If you remove it, it works fine.
However, it does work in Open Zaak with that variable and the value that is also in ON.

**To Reproduce**
1. Start clean ON docker compose
2. Go to /admin/

**Expected behavior**
Layout presents itself as it should

sergei-maertens:
Fixes #741

**Changes**

* Only prefix `subpath` if it's longer than `/`

### DIFF
--- a/src/openzaak/conf/production.py
+++ b/src/openzaak/conf/production.py
@@ -52,7 +52,7 @@ SECURE_CONTENT_TYPE_NOSNIFF = True  # Sets X-Content-Type-Options: nosniff
 SECURE_BROWSER_XSS_FILTER = True  # Sets X-XSS-Protection: 1; mode=block
 
 # Deal with being hosted on a subpath
-if subpath:
+if subpath and subpath != "/":
     STATIC_URL = f"{subpath}{STATIC_URL}"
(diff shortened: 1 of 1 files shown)


stevenbal:
Failing docs build seems unrelated to the change?

stevenbal (approved):

sergei-maertens:
> Failing docs build seems unrelated to the change?

indeed - via via the bug has been reported to Pleio (issue number #6218). Maybe for the time being we'll need to ignore that URL in the linkcheck.

stevenbal:
> > Failing docs build seems unrelated to the change?
> 
> indeed - via via the bug has been reported to Pleio (issue number #6218). Maybe for the time being we'll need to ignore that URL in the linkcheck.

I added it in https://github.com/open-zaak/open-zaak/pull/840/commits/94e37cb0a8f8bfdd5afcd4b3e4844ded8ec4ed77

PR #841, opened 2021-02-03
:wrench: Increase DATA_UPLOAD_MAX_NUMBER_FIELDS for large auth operations

### LINKED ISSUE #807: Authorizations on many zaaktypen fails due to maximum number of fields in POST body
**Describe the bug**
When setting authorizations on many zaaktypen, the POST body can contain too many fields. See here: https://docs.djangoproject.com/en/2.2/ref/settings/#data-upload-max-number-fields This leads to a 500 error.

**Expected behavior**
Should save :)

stevenbal:
Fixes #807

**Changes**
* Increase `DATA_UPLOAD_MAX_NUMBER_FIELDS` to 10000 to allow large auth operations (for many zaaktypen for example)

### DIFF
--- a/src/openzaak/conf/includes/base.py
+++ b/src/openzaak/conf/includes/base.py
@@ -384,6 +384,11 @@ X_FRAME_OPTIONS = "DENY"
 SILENCED_SYSTEM_CHECKS = ["rest_framework.W001"]
 
 #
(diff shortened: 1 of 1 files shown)


sergei-maertens (approved):
The Django default of 1,000 is increased to 10,000 (10x as many)

PR #842, opened 2021-02-03
Issue/806 producten diensten required

### LINKED ISSUE #806: The field "Producten of Diensten" in the Catalogi/Zaaktypen admin interface is required
**Describe the bug**
The field "product_of_diensten" is marked as required in the API. Unfortunately, the API and the UI use 2 different meanings:

* In the API, "required" means: You must provide the field in the body and/or it must be present in the response (as far as I can find).
* In the Admin, "required" means: The field cannot be empty.

In OAS, you can mimic the same behaviour as the admin by setting: `minItems=1`. See also: https://swagger.io/docs/specification/data-models/data-types/ Or, we can remove required from the API.

**To Reproduce**
1. Create Zaaktype in API with `product_of_diensten=[]` (provided, but with empty array)
2. This should succeed
3. Open Zaaktype in Admin and try to save it.
4. An error shows that the field `product_of_diensten` is required.

**Expected behavior**
Behaviour should be consistent.

stevenbal:
Fixes #806

**Changes**

* Added `blank=True` to `ZaakType.producten_of_diensten` to let the admin conform to the API spec


src/openzaak/components/catalogi/migrations/0008_auto_20210203_0947.py
```diff
@@ -0,0 +1,29 @@
+# SPDX-License-Identifier: EUPL-1.2
+# Copyright (C) 2020 Dimpact
```
stevenbal, line 2: @sergei-maertens should this be 2021?

sergei-maertens: for new files, yes, the year that they've been added.

sergei-maertens (approved):

PR #867, opened 2021-02-12
Fix invalid applications showing up in the Autorisaties API

### LINKED ISSUE #835: Openzaak management portal applicatie creation
Hi. I am looking for a confirmation whether this is a bug or not.

**Describe the bug**
When creating an application (API autorisaties -> Applicaties), it is possible to create an application with _heeftAlleAutorisaties_ set to false + without any authorizations defined. 
This seems to conflict with the following business rule in the standard:

_Indien heeftAlleAutorisaties gezet is, dan moet je autorisaties leeg (of weg) laten.
Indien je autorisaties meegeeft, dan moet heeftAlleAutorisaties de waarde false hebben of weggelaten worden_

**To Reproduce**
1. Go to API autorisaties -> Applicaties module, click on "APPLICATIE TOEVOEGEN"
2. Create an application with heeftAlleAutorisaties: false and do not define any authorizations.

**Expected behavior**
Creating an application without any authorizations should not be allowed, as it is said in the standard?

sergei-maertens:
Fixes #835

**Changes**

* Prevent invalid/not-ready applications from showing up in the API (missing authorizations / superuser flag, or having both)
* Display readiness in admin list + hints on how to resolve un-readiness
* Added filter in admin so that invalid apps can quickly be retrieved

Preview:

image

### DIFF
--- a/src/openzaak/components/autorisaties/api/viewsets.py
+++ b/src/openzaak/components/autorisaties/api/viewsets.py
@@ -95,7 +95,13 @@ class ApplicatieViewSet(
     Na het verwijderen wordt een notificatie verstuurd.
     """
 
-    queryset = Applicatie.objects.prefetch_related("autorisaties").order_by("-pk")
+    queryset = (
+        Applicatie.objects.exclude(
+            heeft_alle_autorisaties=False, autorisaties__isnull=True
+        )
+        .prefetch_related("autorisaties")
+        .order_by("-pk")
+    )
     serializer_class = ApplicatieSerializer
     _filterset_class = ApplicatieFilter
     pagination_class = PageNumberPagination
--- a/src/openzaak/components/autorisaties/api/viewsets.py
+++ b/src/openzaak/components/autorisaties/api/viewsets.py
@@ -99,6 +99,7 @@ class ApplicatieViewSet(
         Applicatie.objects.exclude(
             heeft_alle_autorisaties=False, autorisaties__isnull=True
         )
+        .exclude(heeft_alle_autorisaties=True, autorisaties__isnull=False)
         .prefetch_related("autorisaties")
         .order_by("-pk")
     )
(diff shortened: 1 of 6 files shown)


src/openzaak/components/autorisaties/api/viewsets.py
```diff
@@ -95,7 +95,14 @@ class ApplicatieViewSet(
     Na het verwijderen wordt een notificatie verstuurd.
     """
 
-    queryset = Applicatie.objects.prefetch_related("autorisaties").order_by("-pk")
+    queryset = (
+        Applicatie.objects.exclude(
+            heeft_alle_autorisaties=False, autorisaties__isnull=True
+        )
+        .exclude(heeft_alle_autorisaties=True, autorisaties__isnull=False)
```
annashamray, line 102: I'm not sure that invalid applicaties should be hidden for UPDATE or DELETE. It could cover cases, where the user creates invalid applicaties in the admin but then wants to change or delete them through API. 

Perhaps it's the reason why Postman tests are failing for this branch

annashamray (commented):
I think

src/openzaak/components/autorisaties/api/viewsets.py
sergei-maertens: > I'm not sure that invalid applicaties should be hidden for UPDATE or DELETE. It could cover cases, where the user creates invalid applicaties in the admin but then wants to change or delete them through API.

I think you should either use the admin, or the API, not a mixture. The API rules prescribe that it's not allowed to have applications with superuser flag not set but also no authorizations. This expresses the expectation that you shouldn't see objects like that, ever. They are simply not "done" w/r to configuration.

> Perhaps it's the reason why Postman tests are failing for this branch
I'll have to look into that, thanks for the heads-up

sergei-maertens: > Perhaps it's the reason why Postman tests are failing for this branch

This was indeed the case, but the root issue was a bit more nuanced. I believe commit 9e799673e186713face82724a18e59d0eff02c44 should fix it.

annashamray (approved):

PR #875, opened 2021-02-25
Prevent exception when doing PUT and ZaakTypenRelatie already exists

### LINKED ISSUE #851: Updating a zaaktype (PUT) raises an IntegrityError on gerelateerdeZaken
**Describe the bug**
When we've already got a zaaktype on concept and we want to publish it (publicatieIndicatie set on true) we get an internal server error. The message in Open-Zaak log:

_django.db.utils.IntegrityError: duplicate key value violates unique constraint "catalogi_zaaktypenrelati_zaaktype_id_gerelateerd__215fe641_uniq"
DETAIL:  Key (zaaktype_id, gerelateerd_zaaktype)=(379, http://open-zaak.gemeentenet.local/catalogi/api/v1/zaaktypen/81256657-d7b1-4ddb-88de-054fccb0edfc) already exists._

It seems that we also need to update the attribute 'gerelateerdeZaaktypen'. Is this truly mandatory??? Can't we just keep the zaaktype the same and just publish it? 

**To Reproduce**
1. Insert a zaaktype and let it stay on concept
2. Create a PUT call where you set the ' publicatieIndicatie' of the zaaktype of step 1 on true
3. PUT the call to http://open-zaak.gemeentenet.local/catalogi/api/v1/zaaktypen/{uuid} (UUID of the zaaktype at step 1)

**Expected behavior**
A http status 200 with a message that the zaaktype is now published or een error where it says that you can't have the same value at the attribute 'gerelateerdeZaaktypen'

shea-maykinmedia:
Fixes #851 

**Changes**

When doing a `PUT` on the `zaaktypen` endpoint we want to delete `ZaakTypenRelatie` that were previously created so that they can be recreated in this `PUT` request.

### DIFF
--- a/src/openzaak/components/catalogi/api/viewsets/zaaktype.py
+++ b/src/openzaak/components/catalogi/api/viewsets/zaaktype.py
@@ -125,12 +125,13 @@ class ZaakTypeViewSet(
 
     def perform_update(self, serializer):
 
-        for zaaktypenrelaties in serializer.validated_data["zaaktypenrelaties"]:
-            # Delete any ZaakTypenRelatie so they can be recreated
-            ZaakTypenRelatie.objects.filter(
-                zaaktype=serializer.instance,
-                gerelateerd_zaaktype=zaaktypenrelaties["gerelateerd_zaaktype"],
-            ).delete()
+        if not serializer.partial:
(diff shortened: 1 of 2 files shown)


sergei-maertens (changes requested):
Can you add a test where two existing relations are replaced by just a single one as well?

sergei-maertens (approved):

PR #878, opened 2021-03-03
Change CharFilter to a ChoiceFilter 

### LINKED ISSUE #870: Using an invalid value for the 'status' query_param of catalog list operation returns Http error 500
**General information**
Open Zaak version: current on test.openzaak.nl

**Describe the bug**
Server returns error 500 instead of an error message.

**To Reproduce**
Do a list operation on zaaktype and use `alle` instead of `alles` as `status` query (I used ZGWClient from consumers).

For example: http://localhost:9000/catalogi/api/v1/zaaktypen?identificatie=B1796&catalogus=http%3A%2F%2Flocalhost%3A9000%2Fcatalogi%2Fapi%2Fv1%2Fcatalogussen%2Fcce9db14-7acb-41d9-b841-ee10b24d5384&status=alle

**Expected behavior**
I expected a well structured Http 400 error message

There is an error report for test.openzaak.nl here: https://sentry.maykinmedia.nl/maykin-media/open-zaak-public-test/issues/282106/events/?query=is%3Aunresolved

shea-maykinmedia:
Fixes #870 

**Changes**

By changing the CharFilter to a ChoiceFilter the user will receive a 400 error response and the following data back rather than a 500 error

```
{
    "type": "http://localhost:8000/ref/fouten/ValidationError/",
    "code": "invalid",
    "title": "Invalid input.",
    "status": 400,
    "detail": "",
    "instance": "urn:uuid:48c051dc-0064-4529-b899-e3b0bd83367b",
    "invalidParams": [
        {
            "name": "status",
            "code": "invalid_choice",
            "reason": "Selecteer een geldige keuze. %(value)s is geen beschikbare keuze."
        }
    ]
}
```

The only thing I don't like is that `%(value)s` is returned instead of the value they passed in.  This happens because `django-filter`  here https://github.com/carltongibson/django-filter/blob/2.0.0/django_filters/utils.py#L320 uses `e.message` without passing in  the proper `e.params`.  

This is fixed in 2.1.0 https://github.com/carltongibson/django-filter/blob/2.1.0/django_filters/utils.py#L320 so we could upgrade to this version but I wasn't sure if that should be done as part of this pull request.  If we do want to do this as part of this PR let me know.

### DIFF
--- a/src/openzaak/components/catalogi/api/filters.py
+++ b/src/openzaak/components/catalogi/api/filters.py
@@ -30,13 +30,23 @@ STATUS_HELP_TEXT = """filter objects depending on their concept status:
 * `definitief`: Toon objecten waarvan het attribuut `concept` false is (standaard).
 """
 
+ALLES = "alles"
+DEFINITIEF = "definitief"
+CONCEPT = "concept"
+
+STATUS_CHOICES = (
+    (ALLES, "Alles"),
+    (DEFINITIEF, "Definitief"),
+    (CONCEPT, "Concept"),
+)
+
 
 def status_filter(queryset, name, value):
-    if value == "concept":
+    if value == CONCEPT:
         return queryset.filter(**{name: True})
-    elif value == "definitief":
+    elif value == DEFINITIEF:
         return queryset.filter(**{name: False})
-    elif value == "alles":
+    elif value == ALLES:
         return queryset
 
 
@@ -55,8 +65,11 @@ class CharArrayFilter(filters.BaseInFilter, filters.CharFilter):
 
 
 class RolTypeFilter(FilterSet):
-    status = filters.CharFilter(
-        field_name="zaaktype__concept", method=status_filter, help_text=STATUS_HELP_TEXT
+    status = filters.ChoiceFilter(
+        field_name="zaaktype__concept",
+        method=status_filter,
+        help_text=STATUS_HELP_TEXT,
+        choices=STATUS_CHOICES,
     )
 
     class Meta:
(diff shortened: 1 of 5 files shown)


src/openzaak/components/catalogi/api/filters.py
```diff
@@ -30,13 +30,23 @@
...
+ALLES = "alles"
+DEFINITIEF = "definitief"
+CONCEPT = "concept"
+
+STATUS_CHOICES = (
+    (ALLES, "Alles"),
+    (DEFINITIEF, "Definitief"),
+    (CONCEPT, "Concept"),
+)
+
```
sergei-maertens, line 42: Please use `DjangoChoices` for this

sergei-maertens (changes requested):
You can do the django-filter update in this PR!

shea-maykinmedia:
Updated the django-filters library and I can confirm that the bad query param passed in is also returned in the response.

Eg.

```
{
    "type": "http://localhost:8000/ref/fouten/ValidationError/",
    "code": "invalid",
    "title": "Invalid input.",
    "status": 400,
    "detail": "",
    "instance": "urn:uuid:9a8a92ce-49f7-4cf3-b53a-1a58344f0235",
    "invalidParams": [
        {
            "name": "status",
            "code": "invalid_choice",
            "reason": "Selecteer een geldige keuze. alle is geen beschikbare keuze."
        }
    ]
}
```

sergei-maertens (changes requested):
Can you fix the requirements here too?

shea-maykinmedia:
Fixed now! :)

sergei-maertens (approved):

PR #880, opened 2021-03-03
Send notifications for Catalogus admin

### LINKED ISSUE #805: Managing catalogi in the admin interface does not seem to trigger notifications
**Describe the bug**

Catalogus management should send notifications on the CRUD actions as the API does, which does not seem to be the case.

The admin interface is currently the only known viable catalogus management interface. While the API does send notifications, the admin interface operates directly on the database and therefor, notifications are not sent.

I couldn't find any evidence in the codebase or tests that the admin interface does send notifications, and I'm observing behaviour on a client that's subscribed to the *type channels that matches with a lack of notifications being sent.

shea-maykinmedia:
Fixes #805 

**Changes**

Send notifications for changes in Admin.

### DIFF
--- a/src/openzaak/components/catalogi/admin/forms.py
+++ b/src/openzaak/components/catalogi/admin/forms.py
@@ -51,6 +51,35 @@ class ZaakTypeForm(forms.ModelForm):
             return
         self.fields[field].widget.required = True
 
+    def has_changed(self):
+        # Since Django thinks the below fields are always changed we
+        #   need to manually check them ourselves
+
+        for field in [
+            "initial-producten_of_diensten",
+            "initial-verantwoordingsrelatie",
+            "initial-trefwoorden",
+            "selectielijst_procestype_jaar",
+        ]:
+            if self.data[field] != str(
+                getattr(self.instance, field.replace("initial-", ""))
+            ):
+                return True
+
+        changed_data = [
+            data
+            for data in self.changed_data
+            if data
+            not in (
+                "trefwoorden",
+                "producten_of_diensten",
+                "verantwoordingsrelatie",
+                "selectielijst_procestype_jaar",
+            )
+        ]
+
+        return bool(changed_data)
+
     def clean(self):
         super().clean()
 
--- a/src/openzaak/components/catalogi/admin/mixins.py
+++ b/src/openzaak/components/catalogi/admin/mixins.py
@@ -431,10 +431,10 @@ class NotificationMixin:
         viewset = viewset_cls()
 
         send_notification = False
-        if not obj.pk:
+        if not obj.pk or "_addversion" in request.POST:
             send_notification = True
             viewset.action = "create"
-        elif form.has_changed():
+        elif form.has_changed() or "_publish" in request.POST:
             send_notification = True
             viewset.action = "update"
 
--- a/src/openzaak/components/catalogi/admin/mixins.py
+++ b/src/openzaak/components/catalogi/admin/mixins.py
@@ -15,9 +15,16 @@ from django.utils.html import format_html
 from django.utils.text import slugify
 from django.utils.translation import ugettext_lazy as _, ungettext_lazy
 
+from rest_framework.settings import api_settings
+
 from openzaak.utils.admin import ExtraContextAdminMixin
 
-from ..models import Catalogus, InformatieObjectType, ZaakType
+from ..api.viewsets import (
+    BesluitTypeViewSet,
+    InformatieObjectTypeViewSet,
+    ZaakTypeViewSet,
+)
+from ..models import BesluitType, Catalogus, InformatieObjectType, ZaakType
 from .forms import CatalogusImportForm
 from .helpers import AdminForm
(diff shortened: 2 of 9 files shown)


src/openzaak/components/catalogi/admin/mixins.py
```diff
@@ -410,3 +417,25 @@ def get_concept(self, obj):
...
+        elif isinstance(obj, InformatieObjectType):
+            viewset = InformatieObjectTypeViewSet()
+        elif isinstance(obj, BesluitType):
+            viewset = BesluitTypeViewSet()
+
+        scheme = api_settings.DEFAULT_VERSIONING_CLASS()
+        request.version, request.versioning_scheme = (
+            scheme.determine_version(request, version=api_settings.DEFAULT_VERSION),
+            scheme,
+        )
```
sergei-maertens, line 435: I'd prefer an approach similar to https://github.com/encode/django-rest-framework/blob/master/rest_framework/views.py#L385 where the `request` is wrapped in `rest_framework.request.Request`, rather than mutating the original `request` object.

Side-effects are not cool, try to avoid them!

sergei-maertens: additionally, looking at https://github.com/encode/django-rest-framework/blob/master/rest_framework/views.py#L373 you should be able to rely on the viewset for the versioning scheme stuff:

```py
version, scheme = viewset.determine_version(request)
```

src/openzaak/components/catalogi/admin/mixins.py
```diff
@@ -410,3 +417,25 @@ def get_concept(self, obj):
...
+
+        viewset.action = "update"
+        data = viewset.serializer_class(obj, context={"request": request}).data
+        viewset.notify(status_code=200, data=data, instance=obj)
```
sergei-maertens, line 439: You shouldn't always send a notification, only when something has changed. Any chance you can figure that out? Seems like it should be possible given that you have the `form` object and the `obj` instance.

Also - this doesn't account for `create` notifications rather than `update`.

src/openzaak/components/catalogi/tests/admin/test_notifications.py
```diff
@@ -0,0 +1,151 @@
...
+        mock_resource_get(m, "procestypen", procestype_url)
+
+        besluit = BesluitTypeFactory.create(concept=True, omschrijving="test")
+        url = reverse("admin:catalogi_besluittype_change", args=(besluit.pk,))
+
+        response = self.app.get(url)
+        form = response.forms["besluittype_form"]
+        form.submit("_save")
+
+        notify_mock.assert_called_once()
```
sergei-maertens, line 90: I strongly dislike using `unittest.mock` for this kind of stuff. Please use `requests_mock` and check the actual body of the notification, and make the right assertions on that body.

sergei-maertens (changes requested):
I'm also missing tests where a zaaktype/informatieobjecttype/besluittype is created in the admin, and the right notification is being sent, and the case where a *type is published and/or a new version is being created.

src/openzaak/components/catalogi/admin/forms.py
```diff
@@ -51,6 +51,35 @@ def _make_required(self, field: str):
             return
         self.fields[field].widget.required = True
 
+    def has_changed(self):
+        # Since Django thinks the below fields are always changed we
```
shea-maykinmedia, line 55: I think this is the best way to do this.  This was always evaluating to `True` because the `ArrayField`s and `PositiveIntegerField` send their data as a string and since it's saved differently in the model `has_changed` was always returning True.  (Maybe worth filing a bug in Django?).

Since this would mean always sending a notification when a ZaakType was saved I'm manually checking these fields to see if they've actually changed.

src/openzaak/components/catalogi/tests/admin/test_admin_publish.py
```diff
@@ -87,6 +95,12 @@ def test_publish_zaaktype(self, m):
         )
         self.assertIsNotNone(publish_button)
 
+        # Verify notification is sent
+        called_urls = [item.url for item in m.request_history]
+        self.assertIn(
+            "https://notificaties-api.vng.cloud/api/v1/notificaties", called_urls
+        )
```
shea-maykinmedia, line 102: For testing the notification is send when a new version is created or is published I just added to the current tests but if you prefer a new unit test I can do that. :)

src/openzaak/components/catalogi/admin/forms.py
sergei-maertens: sounds okay, that setup with `initial-` hidden form fields recently bit me too, there's something odd going on with it.

src/openzaak/components/catalogi/tests/admin/test_admin_publish.py
sergei-maertens: it's fine, good integration test

sergei-maertens (approved):

PR #882, opened 2021-03-08
added create super script for open-zaak#873

shazada:
Fixes create user on startup.
This is hardcoded admin and admin as a password.

This will be needed to be more variable in the feature.

[Describe the changes here]

### DIFF
--- a/bin/create_superuser.sh
+++ b/bin/create_superuser.sh
@@ -1,0 +1,11 @@
+# Create superuser
+cat <<EOF | python /app/src/manage.py shell
+from django.contrib.auth import get_user_model
+
+User = get_user_model()
+
+if not User.objects.filter(username="$1").exists():
+    User.objects.create_superuser("$1", "$2", "$3")
+else:
+    print('User "{}" exists already, not created'.format("$1"))
+EOF
--- a/bin/docker_start.sh
+++ b/bin/docker_start.sh
@@ -37,6 +37,9 @@ if [ -d $fixtures_dir ]; then
     done
 fi
 
+# Create superuser
+/create_superuser.sh admin admin@admin.org admin
+
 # Start server
 >&2 echo "Starting server"
 uwsgi \
--- a/Dockerfile
+++ b/Dockerfile
@@ -46,6 +46,7 @@ COPY --from=build /usr/local/bin/uwsgi /usr/local/bin/uwsgi
 # Stage 3.2 - Copy source code
 WORKDIR /app
 COPY ./bin/docker_start.sh /start.sh
+COPY ./bin/create_superuser.sh /create_superuser.sh
 RUN mkdir /app/log /app/config
 
 COPY ./config /app/config


bin/create_superuser.sh
```diff
@@ -0,0 +1,11 @@
...
+cat <<EOF | python /app/src/manage.py shell
+from django.contrib.auth import get_user_model
+
+User = get_user_model()
+
+if not User.objects.filter(username="$1").exists():
+    User.objects.create_superuser("$1", "$2", "$3")
+else:
+    print('User "{}" exists already, not created'.format("$1"))
+EOF
```
sergei-maertens, line 11: We should probably replace this with a built-in management command in Open Zaak in [this folder](https://github.com/open-zaak/open-zaak/tree/master/src/openzaak/management/commands) and name it something like `create_initial_superuser` (in the same spirit as django's built in `createsuperuser`).

That way, we can do strict CLI arg parsing, now I'm a bit worried with this approach if you have a `"` character in your password for example :-)

The command should check if the user already exists or not, and exit if it does. If it doesn't, it should call the [built in](https://github.com/django/django/blob/master/django/contrib/auth/management/commands/createsuperuser.py) `createsuperuser` command.

Code should be something like:

```py
import os

from django.contrib.auth.management.commands.createsueruser import Command as BaseCommand


class Command(BaseCommand):

    def add_arguments(self, parser):
        super().add_arguments(parser)

        parser.add_argument(
            "--password",
            help="Set the password when the superuser is initially created.",
        )

    def handle(self, **options):
        username = options[self.UserModel.USERNAME_FIELD]
        database = options['database']
        qs = (
            self.UserModel
            ._default_manager.db_manager(database)
            .filter(**{self.UserModel.USERNAME_FIELD: username})
        )
        if qs.exists():
            self.stdout.write("Superuser account already exists, exiting")
            return

        # perform user creation from core Django
        super().handle(**options)

        password = options.get("password", os.environ.get("DJANGO_SUPERUSER_PASSWORD"))
        if password:
            self.stdout.write("Setting user password...")
            user = qs.get()
            user.set_password(password)
            user.save()
```

1. Exits early if the user account already exists -> makes it idempotent
2. adds the ability to set the password (which `createsuperuser` doesn't have yet in this version, while using `--no-input`)
3. Support for envvar `DJANGO_SUPERUSER_PASSWORD`, which newer versions of Django support - we'll already be on the way then for Django 3.2 support without having to update configs. When we update, we can drop the code that sets the password

Building this as a management command gives us the ability to write unit tests as well.

bin/docker_start.sh
```diff
@@ -37,6 +37,9 @@ if [ -d $fixtures_dir ]; then
     done
 fi
 
+# Create superuser
+/create_superuser.sh admin admin@admin.org admin
+
```
sergei-maertens, line 42: I'd make this opt-in via envvars here:

```bash
# specify password by setting DJANGO_SUPERUSER_PASSWORD in the env
# specify username by setting OPENZAAK_SUPERUSER_USERNAME in the env
# specify email by setting OPENZAAK_SUPERUSER_EMAIL in the env

if [ -z "${OPENZAAK_SUPERUSER_USERNAME}" ]; then
    python src/manage.py create_initial_superuser \
        --no-input \
        --username "${OPENZAAK_SUPERUSER_USERNAME}" \
        --email "${OPENZAAK_SUPERUSER_EMAIL:-admin\@admin.org}"
fi
```

These three environment variables can then be documented in the environment variables reference: https://github.com/open-zaak/open-zaak/blob/master/docs/installation/config/env_config.md

sergei-maertens (changes requested):

sergei-maertens:
@shazada any progress on this PR? are you people still working on this, or should this be picked up under community support time by me?

shazada:
@sergei-maertens let's see what they've done the past few days. I guess not much update regarding optimizing the code. I'll comment back asap.

sergei-maertens:
changes have been taken and improved upon in #952, so closing this one. Thanks for the initial work!

PR #889, opened 2021-03-12
Configure Notificaties API service in setup_configuration script

shea-maykinmedia:
**Changes**

This configures the script to update the Notificaties API Service so it's setup correctly by the script.  It updates the `auth_type` of the service because the check the endpoint `view-config` does assumes a certain `auth_type`  and creates a `client_id` and `secret` that can then be used to configure your Open Notificaties instance.

The client_id and secret will be written to the console when running the script. Eg.

```bash
Initial configuration for Open Zaak was setup successfully
Notificaties API Service configured with client_id open-zaak-1O4OzrcxfrFk and secret cnlFSqx5B9vI.  Use this to configure your Open Notifications instance.
```

I tested this by going to the `view-config` endpoint on my localhost.

Screenshot

Screenshot 2021-03-12 at 09 20 26

### DIFF
--- a/src/openzaak/management/commands/setup_configuration.py
+++ b/src/openzaak/management/commands/setup_configuration.py
@@ -1,5 +1,7 @@
 # SPDX-License-Identifier: EUPL-1.2
 # Copyright (C) 2020 Dimpact
+import string
+import random
 from argparse import RawTextHelpFormatter
 
 from django.contrib.sites.models import Site
@@ -14,6 +16,9 @@ from vng_api_common.notifications.constants import (
     SCOPE_NOTIFICATIES_PUBLICEREN_LABEL,
 )
 from vng_api_common.notifications.models import NotificationsConfig
+
+from zgw_consumers.models import Service
+from zgw_consumers.constants import AuthTypes
 
 from openzaak.components.autorisaties.api.scopes import SCOPE_AUTORISATIES_LEZEN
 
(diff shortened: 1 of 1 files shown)


src/openzaak/management/commands/setup_configuration.py
```diff
@@ -80,6 +84,22 @@ def handle(self, *args, **options):
...
 
             # Step 2
+            # This service should be created in migration
+            #   0002_move_config_to_service_model
+            service = Service.objects.get(label="Notificaties API")
+            service.api_root = notifications_api_root
+            service.auth_type = AuthTypes.zgw
+            random_client_id = (
+                f'open-zaak-{"".join(random.choices(string.ascii_letters, k=10))}'
+            )
```
sergei-maertens, line 94: rather use `from django.utils.crypto import get_random_string` instead for this kind of stuff. It essentially does the same, but I prefer using framework functions.

src/openzaak/management/commands/setup_configuration.py
```diff
@@ -80,6 +84,22 @@ def handle(self, *args, **options):
             notif_config.save()
 
             # Step 2
+            # This service should be created in migration
+            #   0002_move_config_to_service_model
+            service = Service.objects.get(label="Notificaties API")
```
sergei-maertens, line 89: I'd rather not rely on a hardcoded name, but on the service type:

```suggestion
            service = Service.objects.filter(api_type=APITypes.nrc).first()
            assert service is not None, "Service should have been created by migrations"
```

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #894, opened 2021-03-16
Removing disabled newrelic APM monitoring

alextreme:
**Changes**

In the past we were using New Relic actively and had this commented out in the wsgi.py file to easily enable APM monitoring.

However this was long ago and we now internally use self-hosted elastic-apm for most production systems. Enabling this for open-zaak could be interesting, but as I'm cleaning up newrelic from older projects I've done this too for open-zaak.


sergei-maertens (approved):
Great, thanks! This has been a bit of an eye-sore but it didn't hurt anything, so thanks for the cleanup.

PR #897, opened 2021-03-17
Fixes #881 - Create document without identificatie

### LINKED ISSUE #881: Creating a document without Identificatie fails
**General information**
When creating a document without identificatie, the final document updated in the DMS doesn't have any content.
This seems to have to do with the fact that an extra update operation happens to set the identificatie:
https://github.com/open-zaak/open-zaak/blob/master/src/openzaak/components/documenten/query/cmis.py#L544 

This should also be made more efficient by moving this code to the `create_document` function of the CMIS-adapter. Before the document is created, we should check if there is an identificatie field filled. If not, it should be set to the UUID *before* the document is created.

**To Reproduce**
This test can be added [here](https://github.com/open-zaak/open-zaak/blob/master/src/openzaak/components/documenten/tests/test_eio_file_cmis.py#L96):

```python
    def test_create_enkelvoudiginformatieobject_without_identificatie(self):

        informatieobjecttype = InformatieObjectTypeFactory.create(concept=False)
        informatieobjecttype_url = reverse(informatieobjecttype)
        url = get_operation_url("enkelvoudiginformatieobject_create")
        data = {
            "bronorganisatie": "159351741",
            "creatiedatum": "2018-07-01",
            "titel": "text_extra.txt",
            "auteur": "ANONIEM",
            "formaat": "text/plain",
            "taal": "dut",
            "inhoud": base64.b64encode(b"Extra tekst in bijlage").decode("utf-8"),
            "informatieobjecttype": f"http://testserver{informatieobjecttype_url}",
            "vertrouwelijkheidaanduiding": VertrouwelijkheidsAanduiding.openbaar,
        }

        response = self.client.post(url, data)

        self.assertEqual(response.status_code, status.HTTP_201_CREATED, response.data)

        eio = EnkelvoudigInformatieObject.objects.get()

        self.assertEqual(eio.identificatie, eio.uuid)

        download_url = urlparse(response.data["inhoud"])

        response = self.client.get(download_url)

        self.assertEqual(200, response.status_code)
        self.assertGreater(0, len(response.content))
```

SilviaAmAm:
Fixes #881 

The dependency on the cmis-adapter needs to be updated when https://github.com/open-zaak/cmis-adapter/pull/46 is released.

### DIFF
--- a/src/openzaak/components/documenten/query/cmis.py
+++ b/src/openzaak/components/documenten/query/cmis.py
@@ -11,7 +11,7 @@ from typing import List, Optional, Tuple
 from django.db import IntegrityError
 from django.db.models import fields
 from django.db.models.query import BaseIterable
-from django.forms.models import model_to_dict
+from django.forms import model_to_dict
 from django.utils import timezone
 
 from django_loose_fk.virtual_models import ProxyMixin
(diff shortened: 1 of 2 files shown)


sergei-maertens (approved):

sergei-maertens:
@SilviaAmAm can you rebase on `main`? cmis-adapter 1.2.3 has been released

sergei-maertens (approved):

PR #904, opened 2021-03-18
Bump uwsgi version

### LINKED ISSUE #902: Using sendfile in uWSGI when using the CMIS-adapter can go wrong since the dowloaded file is not actually a file
Solution: Update to uWSGI 2.0.19+

See: https://github.com/unbit/uwsgi/issues/1126#issuecomment-665981654

sergei-maertens:
Fixes #902

**Changes**

* Updated to pip-tools 6 (because of updated pip version, and that unfortunately pollutes this PR)
* Updated to uwsgi 2.0.19.1


sergei-maertens:
#903 had an oops, but was approved by Joeri, so merging this in order to finish #905

PR #912, opened 2021-03-23
Bump DRF, drf-yasg, drf-camelcase

### LINKED ISSUE #935: Update the dependencies to newer versions
Open Zaak has a lot of dependencies, which we need to monitor to stay up to date otherwise we might run into issues when some versions are no longer maintained.

Most notably, this concerns:

* Django, currently on 2.2. Extended support ends April 2022
* django-rest-framework, currently on 3.9, latest is 3.12.x series
* drf-yasg, used under the hood by vng-api-common for API schema generation

Often upgrading smaller packages requires upgrades to one of the big ones to stay compatible. We do not want to end up in a place where we can't upgrade a package (security updates are an important aspect here!) because our "main" dependencies are too far behind.

The current state of deps can be seen here: https://requires.io/github/open-zaak/open-zaak/requirements/?branch=main

To coordinate this:

- [x] Update vng-api-common to allow newer versions and stay compatible (https://github.com/VNG-Realisatie/vng-api-common/commit/bdf4fb905ad9233599ea95579da92230e8453a84)
- [x] Update reference implementation and standard with fixes produced by the upgrades (https://github.com/VNG-Realisatie/zaken-api/releases/tag/1.0.3)
- [ ] Update own dependencies & verify tests/build

sergei-maertens:
In preparation for newer versions of Django (3.2 is around the corner) and DRF

Fixes #935

**Changes**

* Bumped to vng-api-common 1.0.63
    * this now allows newer versions of `drf-yasg`, `djangorestframework` and `django`
    * implies some schema changes, which have been released as Zaken API 1.0.3
* Updated to latest Django 2.2.x
* Updated to djangorestframework 3.12 (from 3.9)
* Bumped drf-camelcase (to 1.2) & applied some monkeypatches to not break the API spec
* Updated drf-nested-routers (0.90.2 -> 0.93.3)
* Updated drf-yasg (1.16 -> 1.20)
* Updated drf-writable-nested (0.4.3 -> 0.6.2)
* Fixed some tests that weren't properly isolated
* Added regression test for Autorisaties API empty scopes (+ applied fix)
* Dropped fork of django-better-admin-arrayfield in favour of the original package which is now (better) maintained
* Moved gitpython dependency from production to dev dependencies
* Dropped nlx-url-rewriter and drf-flex-fields dependencies (both unused)


src/openzaak/components/besluiten/openapi.yaml
```diff
@@ -1136,7 +1136,6 @@ components:
           items:
             $ref: '#/components/schemas/FieldValidationError'
     Wijzigingen:
-      title: Wijzigingen
```
sergei-maertens, line 1139: I don't really consider this a problem, but we should check if we can bring this back.

src/openzaak/components/zaken/openapi.yaml
```diff
@@ -3576,8 +3575,8 @@ components:
           title: Inp locatiebeschrijving
           type: string
           maxLength: 1000
+      nullable: true
```
sergei-maertens, line 3578: this is diverting from the standard spec -> investigate why this is happening

src/openzaak/components/zaken/openapi.yaml
```diff
@@ -3597,20 +3596,20 @@ components:
           type: string
           maxLength: 40
           minLength: 1
-        subAdresBuitenland_1:
+        subAdresBuitenland1:
```
sergei-maertens, line 3599: the underscore-to-camelCase thing is doing something odd here, this *may* be configurable.

src/openzaak/components/zaken/openapi.yaml
```diff
@@ -3625,7 +3624,7 @@ components:
             NATUURLIJK PERSOON
           type: string
           maxLength: 17
-        inpA_nummer:
+        inpANummer:
```
sergei-maertens, line 3627: another breaking changes from the upstream standard -> check what's causing this and if we can prevent it

src/openzaak/components/zaken/openapi.yaml
```diff
@@ -4319,8 +4309,8 @@ components:
           title: Ogo locatie aanduiding
           type: string
           maxLength: 100
+      nullable: true
```
sergei-maertens, line 4312: investigate

src/openzaak/components/zaken/openapi.yaml
```diff
@@ -4333,8 +4323,8 @@ components:
           minLength: 1
         adresAanduidingGrp:
           $ref: '#/components/schemas/TerreinGebouwdObjectAdres'
+      nullable: true
```
sergei-maertens, line 4326: investigate

src/openzaak/components/zaken/openapi.yaml
```diff
@@ -4949,8 +4926,8 @@ components:
           title: Locatie omschrijving
           type: string
           maxLength: 1000
+      nullable: true
```
sergei-maertens, line 4929: investigate

src/openzaak/components/zaken/openapi.yaml
```diff
@@ -5676,7 +5645,7 @@ components:
           enum:
           - blijvend_bewaren
           - vernietigen
-        archiefnominatie__in:
+        archiefnominatie_In:
```
sergei-maertens, line 5648: breaking change from the standard, should not happen

src/openzaak/components/zaken/openapi.yaml
```diff
@@ -5689,15 +5658,15 @@ components:
             aan deze ZAAK indien nog leeg.
           type: string
           minLength: 1
-        archiefactiedatum__lt:
+        archiefactiedatum_Lt:
```
sergei-maertens, line 5661: breaking change from the standard, should not happen

src/openzaak/components/zaken/openapi.yaml
```diff
@@ -5689,15 +5658,15 @@ components:
...
           title: Archiefactiedatum  lt
           description: De datum waarop het gearchiveerde zaakdossier vernietigd moet
             worden dan wel overgebracht moet worden naar een archiefbewaarplaats.
             Wordt automatisch berekend bij het aanmaken of wijzigen van een RESULTAAT
             aan deze ZAAK indien nog leeg.
           type: string
           minLength: 1
-        archiefactiedatum__gt:
+        archiefactiedatum_Gt:
```
sergei-maertens, line 5669: breaking change from the standard, should not happen

src/openzaak/components/zaken/openapi.yaml
```diff
@@ -5732,7 +5701,7 @@ components:
           - gearchiveerd
           - gearchiveerd_procestermijn_onbekend
           - overgedragen
-        archiefstatus__in:
+        archiefstatus_In:
```
sergei-maertens, line 5704: breaking change from the standard, should not happen

src/openzaak/components/zaken/openapi.yaml
```diff
@@ -5742,27 +5711,27 @@ components:
           description: De datum waarop met de uitvoering van de zaak is gestart
           type: string
           minLength: 1
-        startdatum__gt:
+        startdatum_Gt:
```
sergei-maertens, line 5714: breaking change from the standard, should not happen

src/openzaak/components/zaken/openapi.yaml
```diff
@@ -5742,27 +5711,27 @@ components:
...
           title: Startdatum  gt
           description: De datum waarop met de uitvoering van de zaak is gestart
           type: string
           minLength: 1
-        startdatum__gte:
+        startdatum_Gte:
```
sergei-maertens, line 5719: breaking change from the standard, should not happen

src/openzaak/components/zaken/openapi.yaml
```diff
@@ -5742,27 +5711,27 @@ components:
...
           title: Startdatum  gte
           description: De datum waarop met de uitvoering van de zaak is gestart
           type: string
           minLength: 1
-        startdatum__lt:
+        startdatum_Lt:
```
sergei-maertens, line 5724: breaking change from the standard, should not happen

src/openzaak/components/zaken/openapi.yaml
```diff
@@ -5742,27 +5711,27 @@ components:
...
           title: Startdatum  lt
           description: De datum waarop met de uitvoering van de zaak is gestart
           type: string
           minLength: 1
-        startdatum__lte:
+        startdatum_Lte:
```
sergei-maertens, line 5729: breaking change from the standard, should not happen

src/openzaak/components/zaken/openapi.yaml
```diff
@@ -5742,27 +5711,27 @@ components:
...
           title: Startdatum  lte
           description: De datum waarop met de uitvoering van de zaak is gestart
           type: string
           minLength: 1
-        rol__betrokkeneType:
+        rol_BetrokkeneType:
```
sergei-maertens, line 5734: breaking change from the standard, should not happen

src/openzaak/components/zaken/openapi.yaml
```diff
@@ -5786,12 +5755,12 @@ components:
           - vestiging
           - organisatorische_eenheid
           - medewerker
-        rol__betrokkene:
+        rol_Betrokkene:
```
sergei-maertens, line 5758: breaking change from the standard, should not happen

src/openzaak/components/zaken/openapi.yaml
```diff
@@ -5786,12 +5755,12 @@ components:
...
           title: Rol  betrokkene
           description: URL-referentie naar een betrokkene gerelateerd aan de ZAAK.
           type: string
           minLength: 1
-        rol__omschrijvingGeneriek:
+        rol_OmschrijvingGeneriek:
```
sergei-maertens, line 5763: breaking change from the standard, should not happen

src/openzaak/components/zaken/openapi.yaml
```diff
@@ -5843,18 +5812,18 @@ components:
           - confidentieel
           - geheim
           - zeer_geheim
-        rol__betrokkeneIdentificatie__natuurlijkPersoon__inpBsn:
+        rol_BetrokkeneIdentificatie_NatuurlijkPersoon_InpBsn:
```
sergei-maertens, line 5815: breaking change from the standard, should not happen

src/openzaak/components/zaken/openapi.yaml
```diff
@@ -5843,18 +5812,18 @@ components:
...
           title: Rol  betrokkeneidentificatie  natuurlijkpersoon  inpbsn
           description: Het burgerservicenummer, bedoeld in artikel 1.1 van de Wet
             algemene bepalingen burgerservicenummer.
           type: string
           minLength: 1
-        rol__betrokkeneIdentificatie__medewerker__identificatie:
+        rol_BetrokkeneIdentificatie_Medewerker_Identificatie:
```
sergei-maertens, line 5821: breaking change from the standard, should not happen

src/openzaak/components/zaken/openapi.yaml
```diff
@@ -5843,18 +5812,18 @@ components:
...
           title: Rol  betrokkeneidentificatie  medewerker  identificatie
           description: Een korte unieke aanduiding van de MEDEWERKER.
           type: string
           minLength: 1
-        rol__betrokkeneIdentificatie__organisatorischeEenheid__identificatie:
+        rol_BetrokkeneIdentificatie_OrganisatorischeEenheid_Identificatie:
```
sergei-maertens, line 5826: breaking change from the standard, should not happen

sergei-maertens:
https://github.com/axnsan12/drf-yasg/commit/293aac91da11b8beaef9bbc70f01ccde543af237# caused the `title` properties to be dropped. They do not affect the resulting schema, so I think we'll have to just accept this change. It's not easy to revert back to the old behaviour for this.

src/openzaak/components/besluiten/openapi.yaml
sergei-maertens: it's not worth the effort

src/openzaak/components/zaken/openapi.yaml
sergei-maertens: these are missing in the standard, but the reference implementation _does_ allow null

sergei-maertens: Updated reference implementation here as well: https://github.com/VNG-Realisatie/zaken-api/pull/196

src/openzaak/components/zaken/openapi.yaml
sergei-maertens: also present in https://github.com/VNG-Realisatie/zaken-api/pull/196

src/openzaak/components/zaken/openapi.yaml
sergei-maertens: see https://github.com/VNG-Realisatie/zaken-api/pull/196

sergei-maertens:
Made a similar PR to the reference implementation, essentially the `nullable` properties are defects in the standard. They are fixed with https://github.com/VNG-Realisatie/zaken-api/pull/196

sergei-maertens:
The upstream standard PR has been merged. I'll take care of updating that for 1.1 as well and then this can be finished soon

src/openzaak/components/autorisaties/migrations/0005_fill_nlx.py
```diff
@@ -1,44 +1,14 @@
 # SPDX-License-Identifier: EUPL-1.2
 # Copyright (C) 2020 Dimpact
 from django.db import migrations
```
joeribekker, line 3: Isn't it better if we just detect the Open Zaak version, or see if nlx-url-rewriter is installed, or fail gracefully, in order to prevent the manual intervention? If we can avoid it, we should.

src/openzaak/components/catalogi/admin/forms.py
```diff
@@ -83,25 +82,6 @@ def has_changed(self):
     def clean(self):
         super().clean()
```
joeribekker, line 84: Is this all fixed in upstream django-better-admin-arrayfield?

src/openzaak/components/catalogi/openapi.yaml
```diff
@@ -3737,7 +3736,6 @@ components:
           type: boolean
           readOnly: true
     BrondatumArchiefprocedure:
-      title: Brondatum archiefprocedure
```
joeribekker, line 3740: Agree that the missing title impact is very very low.

joeribekker (changes requested):

src/openzaak/components/autorisaties/migrations/0005_fill_nlx.py
sergei-maertens: We can't really because of the migration dependency which is no longer installed. This graceful approach has been there for a year already.

The point is that I want to remove nlx-url-rewriter, so detecting the installation doesn't make any sense

src/openzaak/components/catalogi/admin/forms.py
sergei-maertens: Yep! Test coverage dropped because these lines were no longer hit, and I verified it manually + made the tests more explicit

src/openzaak/components/autorisaties/migrations/0005_fill_nlx.py
joeribekker: OK.

joeribekker (approved):

Issue #871, opened 2021-02-18
Installation signup and vendor list and contact info


Ainali:
**Is your feature request related to a problem? Please describe.**
In case of security incidents or other urgent matters, it would be handy to know who is using OpenZaak and how to reach them.

**Describe the solution you'd like**
Some way to securely gather this information that also respects privacy. The information needs to be available to everyone in the Technical steering team. In the last technical steering meeting the task was assigned to @ericherman and me.

Ainali:
This was discussed in the last [technical steering team meeting](https://lists.publiccode.net/hyperkitty/hyperkitty/list/openzaak-discuss@lists.publiccode.net/thread/LN2YL6MQAQ4NR4L6T6UATQVX7DCJBUIS/) where a draft was reviewed.

PR #915, opened 2021-03-25
Document advice to service providers

ericherman:
Part of issue #871, Service providers should be on the [OpenZaak Release Early Notice List](https://odoo.publiccode.net/survey/start/086e0627-8bc0-4b65-8aa9-f6872aba89d0).


ericherman:
We may also wish this to be elsewhere, perhaps README or elsewhere in Support ... regardless, this step in SECURITY seems like a place it must be.

sergei-maertens (approved):

PR #944, opened 2021-04-09
Issue/890 alphanumeric validation

### LINKED ISSUE #890: Zaak identificatie validation in admin differs from validation in API
**General information**
Open Zaak version: 1.3.4 (latest)

Bug reported to me by Gemeente Amsterdam

**Describe the bug**
When creating/editing a `ZaakType`/`Zaak` via the admin (probably applies to all resources that have `identificatie`), values with spaces are not allowed and the following validation error is returned:
image

However, when creating/editing a `ZaakType` via the API, values with spaces are allowed.

**To Reproduce**
1. Create a `ZaakType` via the API with `identificatie="test 1 2"` and observe that the POST succeeds
2. Try to create a `ZaakType` via the admin with `identificatie="test 1 2"` and observe that it doesn't succeed

**Expected behavior**
I would expect the validation to be consistent: so either spaces should be allowed in the admin as well, or spaces should not be allowed in the API.

stevenbal:
Fixes #890 

**Changes**

* Add the `AlphanumericExcludingDiacritic` validator to the Besluit and Zaak resources, so that this validation is also applied via the API

### DIFF
--- a/src/openzaak/components/besluiten/api/serializers.py
+++ b/src/openzaak/components/besluiten/api/serializers.py
@@ -12,7 +12,11 @@ from rest_framework import serializers
 from rest_framework.reverse import reverse
 from vng_api_common.serializers import add_choice_values_help_text
 from vng_api_common.utils import get_help_text
-from vng_api_common.validators import IsImmutableValidator, validate_rsin
+from vng_api_common.validators import (
+    AlphanumericExcludingDiacritic,
+    IsImmutableValidator,
+    validate_rsin,
+)
 
 from openzaak.components.documenten.api.fields import EnkelvoudigInformatieObjectField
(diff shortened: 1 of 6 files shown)


sergei-maertens (changes requested):
imo the fix should be the other way around - relax admin validation and keep API as is. see also https://github.com/VNG-Realisatie/gemma-zaken/issues/1790#issuecomment-820278954

this fix introduces more annoying new problems than it fixes existing problems. it also deviates from the current reference implementation, so that would have to go through a bugfix release in the standard first.

Please swap the approach and make the admin accept spaces instead of restricting them in the API.

sergei-maertens:
Superseded by #996

PR #948, opened 2021-04-12
Feature/910 related zaaktype select

### LINKED ISSUE #910: As admin, I want to be able to easily select internal zaaktypen when creating a Zaaktypenrelatie
As admin, I want to be able to easily select internal zaaktypen when creating a Zaaktypenrelatie (suggested by Gemeente Amsterdam)

**Description**
In the admin UI, it is possible to create `Zaaktypenrelatie`s. In the `Gerelateerd zaaktype` field, it is possible to enter both internal and external zaaktypen. However, the only way to enter these, is by URL. I think it would be good if you could select a `Gerelateerd zaaktype` from a list (similar to how this works for `Zaaktype van`) for internal zaaktypen, while still allowing the option of entering a URL for external zaaktypen.

image

stevenbal:
Fixes #910 

**Changes**

* Split the `ZaakTypenRelatie.gerelateerd_zaaktype` field into two: a selection from local Zaaktypen (like a regular foreign key) and a field in which a raw URL can be entered. After saving the form, the value is always saved as a URL, it's just a visual change in the admin

### DIFF
--- a/src/openzaak/templates/widgets/multiwidget.html
+++ b/src/openzaak/templates/widgets/multiwidget.html
@@ -3,7 +3,7 @@
 {% spaceless %}
 {% for subwidget in widget.subwidgets %}
     {% include subwidget.template_name with widget=subwidget %}
-    {% if forloop.counter != widget.subwidgets|length %}
+    {% if forloop.last %}
         <strong> of </strong>
     {% endif %}
 {% endfor %}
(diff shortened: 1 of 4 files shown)


stevenbal:
image

src/openzaak/templates/widgets/multiwidget.html
```diff
@@ -0,0 +1,10 @@
+{% comment %} SPDX-License-Identifier: EUPL-1.2 {% endcomment %}
+{% comment %} Copyright (C) 2021 Dimpact {% endcomment %}
+{% spaceless %}
+{% for subwidget in widget.subwidgets %}
+    {% include subwidget.template_name with widget=subwidget %}
+    {% if forloop.counter != widget.subwidgets|length %}
```
sergei-maertens, line 6: could use `forloop.last` instead?

sergei-maertens (changes requested):
That's neat!

Very small template code remark, but otherwise good to go.

sergei-maertens (approved):

PR #957, opened 2021-04-21
Log CMIS calls to file

SilviaAmAm:
**Changes**
Before, the CMIS logs were only shown in the console. Now they are logged to a file `cmis.log` by default.

### DIFF
--- a/src/openzaak/conf/includes/base.py
+++ b/src/openzaak/conf/includes/base.py
@@ -248,6 +248,7 @@ LOGGING = {
         "timestamped": {"format": "%(asctime)s %(levelname)s %(name)s  %(message)s"},
         "simple": {"format": "%(levelname)s  %(message)s"},
         "performance": {"format": "%(asctime)s %(process)d | %(thread)d | %(message)s"},
+        "drc": {"format": "%(asctime)s %(name)s %(message)s"},
     },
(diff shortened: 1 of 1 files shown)


sergei-maertens:
In deployed environments, this implies that people need to mount a volume on `/app/log` to have persistent logs. This may cause problems with multiple replicas as well.

sergei-maertens (changes requested):
I would like to see an envvar option to opt in to logging to file rather than stdout, as it's bad practice to write to files in docker containers that are not in a volume. Now the default behaviour is that we get writes to file, which goes against best-practices-by-default.

joeribekker:
I added this to the discussion for the technical steering group agenda.

sergei-maertens:
During the technical steering group call it became clear that different logging solutions exist (Loki, Prometheus, Docker Swarm...). Open Zaak should continue to log to stdout, and your logging solution is responsible for the aggregation and filtering options. Since we emit which modules generate the logs, this should be sufficient to filter logs to get the relevant output.

Setting up file-based logging in container contexts is a non-trivial challenge that mostly brings maintenance overhead for the Open Zaak team. Everyone ultimately benefits from using a mature logging solution.

With these arguments, I'm closing this PR without merging it.

PR #982, opened 2021-05-12
WIP Feature/931 alfresco model packaging

### LINKED ISSUE #931: Alfresco content model packaging
Namens: OPEN.satisfaction
Contact persoon: Silvion Moesan

Op het moment word het content model als XML in een folder in Alfresco gezet. Dit is niet de juiste manier om een Alfresco extensie te packagen. Daarbij mist er Share configuratie die controle geeft over welke velden wel en niet bewerkbaar zijn in Alfresco Share.

# Development
- Alfresco SDK4 project aanmaken van het Archetype
- Bestaande model op de juiste plek in de structuur plaatsen
- Test instructies toevoegen
- Productie deployment instructies toevoegen
- Ongebruikte example onderdelen verwijderen uit de structuur

timijntema:
Fixes #931

**Changes**

[Describe the changes here]


.github/workflows/ci.yml
```diff
@@ -60,7 +60,7 @@ jobs:
 
       - name: Bring up mock endpoints for SSL verification
         run: docker-compose -f docker-compose.travis.yml up -d mock-endpoints.local
-
+
```
sergei-maertens: Code style nitpicking: can you remove this whitespace cfr. the `.editorconfig`?

docker-compose.alfresco.yml
```diff
@@ -0,0 +1,91 @@
...
+      - db
+
+  nginx:
+    image: nginx
+    volumes:
+      - ./default.conf:/etc/nginx/conf.d/default.conf
+    ports:
+      - "9001:80"
+    depends_on:
+      - web
```
sergei-maertens, line 39: Can we leave these services out (or only include the overrides from `docker-compose.yml`) and change the documentation to use `docker-compose -f docker-compose.yml -f docker-compose.alfresco.yml up` if you want to use the Aflresco-backed CMIS store?

This reduces the maintenance overhead in the docker-compose files because of the duplication of services/configuration.

docker-compose.alfresco.yml
```diff
@@ -0,0 +1,91 @@
...
+    image: postgres:9.6
+    environment:
+      POSTGRES_DB: alfresco
+      POSTGRES_USER: alfresco
+      POSTGRES_PASSWORD: alfresco
+    command: postgres -c max_connections=300 -c log_min_messages=LOG
+    ports:
+      - "11234:5432"
+    volumes:
+      - openzaak-alfresco-db-volume:/var/lib/postgresql/data
```
sergei-maertens, line 73: this should just re-use the `db` service, no? not sure why we're spinning up two postgres containers

sergei-maertens (changes requested):

.github/workflows/ci.yml
timijntema: Already removed the newline change

docker-compose.alfresco.yml
timijntema: I won't advise putting two different datamodels in the same database. This would mean we need to create two separate databases in the same container if we only want one database container. If I want to do that, I think we would need to build a custom postgres image that allows multiple databases to be created in a single container. The default postgresql image is not setup for that.

The way it is done now, one postgresql going down due to an error does not kill database access for the other services that use the other postgresql.

docker-compose.alfresco.yml
timijntema: I will check if this works.

timijntema: It should be possible. The host to use for connections from Openzaak to Alfresco would be `host.docker.internal`. The two stacks would not be in the same docker network . Thats why that host is required. The port would be `11232` since that is the external port. It would not work with the internal port using this host.

The other method would be to manually create a network and make the docker compose files use this network. But the downside of that would be the manual creation of the network.

timijntema: @sergei-maertens Is [this](https://open-zaak.readthedocs.io/en/latest/development/getting_started.html?highlight=docker#testing-with-cmis-adapter-enabled) the correct documentation to adjust?

docker-compose.alfresco.yml
sergei-maertens: I agree, therefor we should use the initialization scripts support documented on https://hub.docker.com/_/postgres/, which can be plain SQL scripts executing the `CREATE DATABASE alfresco;` statements.

docker-compose.alfresco.yml
sergei-maertens: `host.docker.internal` is a macOS thing (and possibly Windows) because Docker runs in a VM there, it doesn't work on Linux. Applying both docker files should use the same docker network, and you can extend the services by re-using the same service name (which can be used for the internal DNS in the bridge network still).

w/r to the referenced documentation, that's not the one - that should _only_ spin up Alfresco without Open Zaak services (because we run Open Zaak outside of Docker while actively developing/running tests).

The `docker-compose.yml` might just not be documented at all - or it would probably be in the README. I'm okay with documenting it with some comments at the top in the `docker-compose.yml` file - it's aimed at developers anyway.

docker-compose.alfresco.yml
timijntema: Since I removed the Openzaak part of this docker-compose file, reusing the db service is no longer an option.

sergei-maertens: It's definitely an option when using the `-f` flags for docker-compose - the `db` service can just be added to the `depends_on` array.

I've seen the comment about starting just Alfresco without Open Zaak and in that case, yes, you would need a DB service. This can still be achieved through the multiple compose files, but makes it a bit annoying. However, that whole set-up is unneccesary - this is the Open Zaak project which has optional support for Alfresco, not the Alfresco project with optional support for Open Zaak. Meaning - if you're starting docker-compose from Open Zaak, you are intending to run Open Zaak, always. Running Alfresco standalone is possible, but that's not something we should actively support.

sergei-maertens (changes requested):
Let's park this for today - after all my meeting today I'll follow up with a separate PR incorporating these changes. There's some broader re-structuring in order for the docker-compose files anyway (they're still referring to Travis while we run on Github Actions now, for example).

Thanks for the sample set-up, that's a very good starting point! :-)

sergei-maertens:
@timijntema I've made a PR here: https://github.com/open-zaak/open-zaak/pull/987 can you check that I haven't missed anything on the Alfresco side of things? I've tested it all locally and Share + Alfresco came up + I could configure Open Zaak to connect to that Alfresco instance, so from my end it looks good.

PR #987, opened 2021-05-25
Integrating custom Alfresco image for testing + clean up repo structure

sergei-maertens:
* Restructured all the docker(-compose) stuff so that it's bundled in the `docker` subdirectory rather than spread out over the root of the repo
* Included the custom docker image for the custom content model, from #982 
* Updated the authors list
* Renamed files with Travis references to the vendor-neutral "CI" reference


sergei-maertens:
Haven't heard back from reviewers in a week, so I'm assuming this is fine. We can revisit/patch/tweak if needed.

PR #988, opened 2021-06-01
Fix alignment with DRC reference implementation (1.0.x)

sergei-maertens:
These came up during testing by Contezza, many thanks for that!

**Changes**

* the `RemoveRelationValidator` now emits the correct error code `remote-relation-exists` instead of `inconsistent-relation`
* the `RemoveRelationValidator` now validates external URLs as well instead of letting those through (when deleting ObjectInformatieObject)

### DIFF
--- a/src/openzaak/components/documenten/api/validators.py
+++ b/src/openzaak/components/documenten/api/validators.py
@@ -89,7 +89,7 @@ class RemoteRelationValidator:
 
     def __call__(self, oio: ObjectInformatieObject):
         # external object
-        if isinstance(oio.object, ProxyMixin):
+        if isinstance(oio.object, ProxyMixin) and not settings.CMIS_ENABLED:
             invalid = self._check_remote(oio)
         else:
             invalid = self._check_local(oio)
@@ -112,16 +112,13 @@ class RemoteRelationValidator:
     def _check_remote(self, oio: ObjectInformatieObject) -> bool:
         object_url = oio.object._loose_fk_data["url"]
 
-        if settings.CMIS_ENABLED:
-            document_url = oio.get_informatieobject_url()
-        else:
-            default_version = settings.REST_FRAMEWORK["DEFAULT_VERSION"]
-            document_url = build_absolute_url(
-                oio.informatieobject.latest_version.get_absolute_api_url(
-                    version=default_version
-                ),
-                request=self.request,
-            )
+        default_version = settings.REST_FRAMEWORK["DEFAULT_VERSION"]
+        document_url = build_absolute_url(
+            oio.informatieobject.latest_version.get_absolute_api_url(
+                version=default_version
+            ),
+            request=self.request,
+        )
 
         # obtain a client for the remote API. this should exist, otherwise loose-fk
         # would not have been able to load this resource :-)
--- a/src/openzaak/components/documenten/api/validators.py
+++ b/src/openzaak/components/documenten/api/validators.py
@@ -112,13 +112,16 @@ class RemoteRelationValidator:
     def _check_remote(self, oio: ObjectInformatieObject) -> bool:
         object_url = oio.object._loose_fk_data["url"]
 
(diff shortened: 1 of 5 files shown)


src/openzaak/components/documenten/api/validators.py
```diff
@@ -66,3 +76,68 @@ class UniekeIdentificatieValidator(_UniekeIdentificatieValidator):
...
+    def _check_remote(self, oio: ObjectInformatieObject) -> bool:
+        object_url = oio.object._loose_fk_data["url"]
+
+        default_version = settings.REST_FRAMEWORK["DEFAULT_VERSION"]
+        document_url = build_absolute_url(
+            oio.informatieobject.latest_version.get_absolute_api_url(
+                version=default_version
+            ),
+            request=self.request,
+        )
```
sergei-maertens, line 121: @SilviaAmAm this works for local documents (without using CMIS), but I remember the `Canonical` construction being difficult w/r to CMIS.

What I'm not sure is:

- given an OIO instance, using CMIS
- what's the best way to get the fully qualified document URL of the EIO

I need this URL for the query parameters used in the `besluitinformatieobject` or `zaakinformatieobject` list call.

src/openzaak/components/documenten/tests/test_objectinformatieobject.py
```diff
@@ -519,36 +555,83 @@ def test_list_filter_by_external_besluit(self):
         self.assertEqual(len(response.data), 1)
         self.assertEqual(response.data[0]["object"], besluit2)
 
-    def test_destroy_external_zaak(self):
-        zaak = "https://externe.catalogus.nl/api/v1/zaken/1c8e36be-338c-4c07-ac5e-1adf55bec04a"
+    @requests_mock.Mocker()
+    def test_destroy_external_zaak(self, m):
```
SilviaAmAm, line 559: Really minor remark, but I find that the name of the test doesn't to match well what happens in the test

src/openzaak/components/documenten/api/validators.py
```diff
@@ -66,3 +76,71 @@ class UniekeIdentificatieValidator(_UniekeIdentificatieValidator):
...
+
+    def __init__(self, request=None):
+        self.request = request
+
+    def __call__(self, oio: ObjectInformatieObject):
+        # external object
+        if isinstance(oio.object, ProxyMixin):
+            invalid = self._check_remote(oio)
+        else:
+            invalid = self._check_local(oio)
```
SilviaAmAm, line 95: This works, but I've been trying lately to handle CMIS objects as 'local' as much as possible. Here in the CMIS case, the OIO is also a `ProxyMixin`, and the `does_zaakinformatieobject_exist` and `does_besluitinformatieobject_exist` can be used. So maybe a possibility would be:
```python
        if isinstance(oio.object, ProxyMixin) and not settings.CMIS_ENABLED:
            invalid = self._check_remote(oio)
        else:
            invalid = self._check_local(oio)
```
And then remove the CMIS case in the `_check_remote()` method?

src/openzaak/components/documenten/tests/test_objectinformatieobject.py
```diff
@@ -519,36 +555,83 @@ def test_list_filter_by_external_besluit(self):
...
-            response = self.client.delete(url)
+        response = self.client.delete(url)
 
         self.assertEqual(response.status_code, status.HTTP_204_NO_CONTENT)
         self.assertEqual(ObjectInformatieObject.objects.count(), 0)
 
-    def test_destroy_external_besluit(self):
-        besluit = "https://externe.catalogus.nl/api/v1/besluiten/1c8e36be-338c-4c07-ac5e-1adf55bec04a"
+    @requests_mock.Mocker()
+    def test_destroy_external_besluit(self, m):
```
SilviaAmAm, line 582: Same minor remark as for the other test

SilviaAmAm (approved):
Looks good to me!

sergei-maertens:
Thanks for the feedback, good suggestions! I'll process them

Issue #869, opened 2021-02-14
Better security: don't run as root user in container


sergei-maertens:
The Dockerfile does not drop privileges to a non-privileged user, while it's perfectly possible.

sergei-maertens:
Note: could be useful/needed for Podman support

ericherman:
From Tech Steering: This has an impact on deployed containers, we can limit impact by doing it sooner.

sergei-maertens:
Fix was added, but to be able to run in Kubernetes with `podSecurityContext.fsGroup`, we need to assign a non-root group as well.

PR #990, opened 2021-06-18
:whale: Drop superuser priviliges in Dockerfile

sjoerdie:
to ensure that containers can be run as non-root users

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,7 +6,11 @@ Changelog
 
 .. warning::
 
-   No longer run as root user, fixing issue: https://github.com/open-zaak/open-zaak/issues/869
+   Manual intervention required
+   
+  k8s the correct pod securityGroupContext needs to be specified and on plain Docker 
+  deploys the permissions of the volumes will probably have to be fixed. 
+
 
 1.4.0 (2021-04-30)
 ------------------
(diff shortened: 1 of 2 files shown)


sergei-maertens (changes requested):
Thanks, the change itself is fine!

Couple things (as discussed on Slack):

* please refer to the issue #869 in the commit message
* please update the changelog, as this change requires manual intervention - on k8s the correct pod `securityGroupContext` needs to be specified (I think `fsGroup: 1000`) and on plain Docker deploys the permissions of the volumes will probably have to be fixed. Just having a placeholder with some words as above is fine, I'll take care of the full text when the 1.5.0 release is being prepared.

Finally - this change should also happen for Open Notificaties

CHANGELOG.rst
```diff
@@ -1,6 +1,17 @@
...
 
+1.5.0 (2021-??-??)
+------------------
+
+.. warning::
+
+   Manual intervention required
+   
+  k8s the correct pod securityGroupContext needs to be specified and on plain Docker 
+  deploys the permissions of the volumes will probably have to be fixed.
```
sergei-maertens, line 12: Deliberately keeping this vague-ish for now, will be cleaned up when the 1.5.0 release is prepared

sergei-maertens (approved):

sergei-maertens (approved):

PR #993, opened 2021-06-21
:wrench: Set the uWSGI timeout to 5 minutes.

joeribekker:
Fixes timeout settings not matching Nginx timeout.

**Changes**

Increases the (undocumented) default timeout of uWSGI from 60s to 300s. This is in line with the Nginx timeout which is also set to 300s.

### DIFF
--- a/bin/docker_start.sh
+++ b/bin/docker_start.sh
@@ -62,4 +62,5 @@ exec uwsgi \
     --enable-threads \
     --processes $uwsgi_processes \
     --threads $uwsgi_threads \
-    --buffer-size=65535
+    --buffer-size=65535 \
+    --http-timeout=300


bin/docker_start.sh
```diff
@@ -62,4 +62,5 @@ exec uwsgi \
     --enable-threads \
     --processes $uwsgi_processes \
     --threads $uwsgi_threads \
-    --buffer-size=65535
+    --buffer-size=65535 \
+    --http-timeout=300
```
sergei-maertens, line 66: can we make this an environment variable?

It should really be an edge case that an HTTP request takes _this_ long and not be the norm.

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #999, opened 2021-06-29
Feature/oidc integration

stevenbal:
fixes: https://github.com/open-zaak/open-zaak/issues/1002

For Den Haag, to see if we can use Keycloak (or other OpenID connect providers for that matter) for authentication

Uses a package I built on top of https://github.com/mozilla/mozilla-django-oidc (basically the same idea as https://github.com/isprojects/django-auth-adfs-db, though it also changes the default behaviour of the backend)

See https://github.com/maykinmedia/mozilla-django-oidc-db for the changes I made to ``mozilla-django-oidc``

TODO:
- [x] Change user creation (use info from identity provider to fill `first_name`, etc.)
- [x] It appears that users are identified by email (which is against the OIDC spec https://openid.net/specs/openid-connect-core-1_0.html#ClaimStability), so maybe the user creation/filtering should be changed -> changed to use `sub` (subject) as unique ID
- [x] Use django-solo for configuration
- [x] Add solo model to admin index fixture
- [x] Add documentation on configuration
- [x] Add tests

### DIFF
--- a/docs/manual/oidc.rst
+++ b/docs/manual/oidc.rst
@@ -1,0 +1,80 @@
+.. _manual_oidc:
+
+===========================
+OpenID Connect configureren
+===========================
+
+Open Zaak ondersteunt Single Sign On (SSO) via het OpenID Connect protocol (OIDC) voor de beheerinterface.
+
+Gebruikers kunnen op die manier inloggen op Open Zaak met hun account bij de OpenID Connect provider. In deze
+flow:
+
+1. Klikt een gebruiker op het inlogscherm op *Inloggen met OIDC*
+2. De gebruiker wordt naar de omgeving van de OpenID Connect provider geleid (bijv. Keycloak) waar ze inloggen met gebruikersnaam
+   en wachtwoord (en eventuele Multi Factor Authorization)
+3. De OIDC omgeving stuurt de gebruiker terug naar Open Zaak (waar de account aangemaakt
+   wordt indien die nog niet bestaat)
+4. Een beheerder in Open Zaak kent de juiste groepen toe aan deze gebruiker als deze
+   voor het eerst inlogt.
+
+.. note:: Standaard krijgen deze gebruikers **geen** toegang tot de beheerinterface. Deze
+   rechten moeten door een (andere) beheerder :ref:`ingesteld <manual_users>` worden. De
+   account is wel aangemaakt.
+
+.. _manual_oidc_appgroup:
+
+Configureren van OIDC zelf
+==========================
+
+Contacteer de IAM beheerders in je organisatie om een *Client* aan te
+maken in de omgeving van de OpenID Connect provider.
+
+Voor de **Redirect URI** vul je ``https://open-zaak.gemeente.nl/oidc/callback`` in,
+waarbij je ``open-zaak.gemeente.nl`` vervangt door het relevante domein.
+
+Aan het eind van dit proces moet je de volgende gegevens hebben (on premise):
+
+* Server adres, bijvoorbeeld ``login.gemeente.nl``
+* Client ID, bijvoorbeeld ``a7d14516-8b20-418f-b34e-25f53c930948``
+* Client secret, bijvoorbeeld ``97d663a9-3624-4930-90c7-2b90635bd990``
+
+Configureren van OIDC in Open Zaak
+==================================
+
+Zorg dat je de volgende :ref:`gegevens <manual_oidc_appgroup>` hebt:
+
+* Server adres
+* Client ID
+* Client secret
+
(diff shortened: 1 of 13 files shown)


alextreme:
Wat mij betreft een mooie uitbreiding en fijn dat we OpenID Connect integratie kunnen toevoegen aan OpenZaak (mits Sergei ermee eens is).

Het is een trivialiteit, maar ik zit een beetje met 'Login using OIDC'. We kunnen het voor nu hernoemen naar 'Login using OpenID Connect' aangezien dat wellicht men iets meer zegt, maar eigenlijk wil je ook de naam beheerbaar hebben om aan te sluiten op wat binnen de organisatie gebruikelijk is. Dit kan Keycloak zijn maar kan net zo goed een andere naam hebben. Wellicht de moeite waard om later mee te pakken in combinatie met andere verbeteringen.

sergei-maertens:
@stevenbal what's the status now? I see my review is requested, but the PR is still in draft. Is this still being actively worked on?

docs/manual/oidc.rst
```diff
@@ -0,0 +1,80 @@
...
+3. Vul bij **OpenID Connect client ID** het Client ID in, bijvoorbeeld
+   ``a7d14516-8b20-418f-b34e-25f53c930948``.
+4. Vul bij **OpenID Connect secret** het Client secret in, bijvoobeeld
+   ``97d663a9-3624-4930-90c7-2b90635bd990``.
+5. Vul bij **OpenID sign algorithm** ``RS256`` in.
+6. Laat bij **OpenID Connect scopes** de standaardwaarden staan.
+
+Vervolgens moeten er een aantal endpoints van de OIDC provider ingesteld worden,
+deze zijn af te leiden uit het OIDC Discovery Endpoint
+(``https://login.gemeente.nl/auth/realms/{realm}/.well-known/openid-configuration``)
```
sergei-maertens, line 64: Can't we configure this discovery URL instead and have the library figure out the correct URLs? That's how the whole OpenID spec works :)

src/openzaak/templates/admin/login.html
```diff
@@ -13,4 +13,12 @@
     <a href="{% url 'django_auth_adfs:login' %}">{% trans "Login with ADFS" %}</a>
 </div>
 {% endif %}
+
+{% get_solo 'mozilla_django_oidc_db.OpenIDConnectConfig' as oidc_config %}
+{% if oidc_config.enabled %}
+<div class="submit-row">
+    <a href="{% url 'oidc_authentication_init' %}">{% trans "Login with OIDC" %}</a>
```
sergei-maertens, line 20: Agree with Alex - this should be configurable :)

sergei-maertens (changes requested):

stevenbal:
> @stevenbal what's the status now? I see my review is requested, but the PR is still in draft. Is this still being actively worked on?

ah forgot to mark as ready, got used to how that works on gitlab :stuck_out_tongue:. It's ready for review

src/openzaak/templates/admin/login.html
sergei-maertens: NLX uses this, and I think that's the simplest:

```suggestion
    <a href="{% url 'oidc_authentication_init' %}">{% trans "Login with organization account" %}</a>
```

sergei-maertens (changes requested):

docs/manual/oidc.rst
stevenbal: Good idea, I could add this to the config model as an optional field and if it is filled in, it could call the endpoint to derive the other endpoints? To still allow manually configuring the other endpoints (if this is necessary for whatever reason)

src/openzaak/accounts/tests/test_oidc.py
```diff
@@ -0,0 +1,40 @@
...
+
+class OIDCLoginButtonTestCase(WebTest):
+    def test_oidc_button_disabled(self):
+        config = OpenIDConnectConfig.get_solo()
+        config.enabled = False
+        config.save()
+
+        response = self.app.get(reverse("admin:login"))
+
+        oidc_login_link = response.html.find("a", string=_("Login with OIDC"))
```
sergei-maertens, line 18:
```suggestion
        oidc_login_link = response.html.find("a", string=_("Login with organization account"))
```

sergei-maertens (changes requested):

docs/manual/oidc.rst
sergei-maertens: yes, that would be okay. You can do that in a separate PR, since I want to merge this asap and release 1.5.0 this month

PR #1011, opened 2021-07-19
Tests: running offline, first pass

sergei-maertens:
This should remove real HTTP calls from the test suite and eventually make
it possible to run the tests offline.

* removes real calls to example.com
* remove real calls to notifications API


sergei-maertens:
Merged this to improve robustness of the test suite. More work will follow, especially dealing with the API spec references is a challenge.

PR #1056, opened 2021-10-28
:test_tube: add regression test for iotype admin notification

annashamray:
regression test to make sure the notification message is correct when the informatiobjecttype is created in the admin


sergei-maertens (changes requested):

sergei-maertens (commented):
This test is now failing for me locally too, so clearly there's an issue :-)

PR #1073, opened 2021-11-24
Pin setuptools for RTD builds

sergei-maertens:
RTD also runs 'pip install -U setuptools' which breaks stuff.


sergei-maertens:
Build is passing now on RTD :tada:

PR #1084, opened 2021-12-22
Problem with error formatting

SilviaAmAm:
I think that the comma in the dictionary was meant to be a colon.


sergei-maertens (approved):

stevenbal:
Is this PR ready to be merged? I switched back to the previous version of the ATP tests for the time being, to prevent CI on master from failing

sergei-maertens:
@stevenbal yeah, I think we were holding off because of the ATP tests. Thanks for fixing those! I just merged this PR, so if you could swap to the new version, that would be great :pray:

PR #1095, opened 2022-02-01
:lock: Check JWT expiry for audittrails and nested Zaak resources

stevenbal:
Fixes: https://github.com/open-zaak/security-issues/issues/3

**Changes**

* Use the proper permission_classes for audittrail viewsets, to enable checking for expired JWTs
* Check JWT expiry for nested Zaak resources


src/openzaak/components/besluiten/api/viewsets.py
```diff
@@ -1,5 +1,5 @@
 # SPDX-License-Identifier: EUPL-1.2
-# Copyright (C) 2019 - 2020 Dimpact
+# Copyright (C) 2019 - 2022 Dimpact
```
sergei-maertens, line 2: FYI - these do not have to be updated, and you can also use just the start year

sergei-maertens (approved):

PR #1103, opened 2022-02-17
Clean up Kubernetes deployment docs

### LINKED ISSUE #854: Improve the documentation on how to deploy on Kubernetes
The primers on Kubernetes/Ansible should be a separate document so that these can easily be skipped for people familiar with any/both.

The documentation should focus on preparing your variables files/configuration and actually deploying the application.

Additionally, it might make sense to wrap all the deployment tooling in a docker container image that can be ran locally, so that users don't have to install all the dependencies.

sergei-maertens:
Fixes #854

**Changes**

* Rewrite the K8s deployment documentation 
* Remove K8s clustering infrastructure code


bartjkdp:
In my opinion we should rethink the way we are supporting the Helm charts. We can either support them or not, but somewhere in the middle is not very helpful to our users and confusing.

If we decide to support the Helm charts, then we can add documentation about them and refer to them. If we are not supporting them, then I would propose to move the charts to a separate organization and only make a small reference from the documentation.

My two cents: Helm is the de-facto standard for deployments on Kubernetes, so supporting Helm as the primary deployment mechanism makes a lot of sense. I think therefore we should go for full support and make sure that we are updating the Helm charts when we are releasing a new version of Open Zaak.

bartjkdp:
@sergei-maertens in that case I think we need to talk to the current production users of Open Zaak and make sure we reserve sufficient funding. A streamlined release process (that includes bumping versions in Helm charts) is vital for software that is being used in production.

docs/installation/deployment/kubernetes.rst
```diff
@@ -204,194 +74,82 @@ You can either clone the https://github.com/open-zaak/open-zaak repository,
...
+.. note::
 
-.. _Kubernetes: https://kubernetes.io/
-.. _NGINX: https://www.nginx.com/
-.. _Ansible: https://www.ansible.com/
+   These charts are contributed by the community on a best-effort basis. The Technical
+   Steering Group takes no responsibility for the quality or up-to-date being of these
+   charts. This may lead to
+   `scary reports <https://github.com/open-zaak/open-zaak/issues/1097>`_ that are
+   easily misinterpreted.
```
bartjkdp, line 152: > This may lead to
   `scary reports <https://github.com/open-zaak/open-zaak/issues/1097>`_ that are
   easily misinterpreted.

I would leave this sentence out as I think it is rather confusing and not directly related to the Helm charts.

sergei-maertens: wrong target audience indeed, removed.

sergei-maertens:
> @sergei-maertens in that case I think we need to talk to the current production users of Open Zaak and make sure we reserve sufficient funding for software maintenance. A streamlined release process is vital for software that is being used in production. I think adding a step in the release process to bump versions in the Helm charts would not be too much of work.
> 
> For now I think we can merge this in with a small change.

I've created #1104 for that and sent an e-mail to the mailing list that should hopefully reach the people with that kind of decision power.

PR #1112, opened 2022-03-10
Upgrade to Django 3.2

### LINKED ISSUE #1098: Update Django to 3.2
The currently used Django version is 2.2 and needs to be migrated to 3.2 (LTS).

**Tasks**

* [x] Follow the guide to [Migrate to Django 3.2](https://docs.djangoproject.com/en/4.0/howto/upgrade-version/)
* [x] Make atomic commits for each chunk of changes (for example: "Updated library X to support Django 3.2", or "Changed import everywhere to use Django 3.2", or "Resolved deprecation issue X in Django 3.2")
* [x] Create a PR that shows that all tests are still succeeding
* [x] Perform visual inspection on the homepage and admin to see if it still looks ok

Known packages to upgrade:

- [x] django-appconf
- [x] django-axes
- [x] django-redis
- [x] django-choices
- [x] drf-extra-fields
- [x] freezegun
- [x] ordered-model
- [x] django-admin-index
- [x] Update django admin overrides stylesheets (see maykin default-project)
- [x] rest_framework_gis

sergei-maertens:
Fixes #1098

**Changes**

I didn't really have to make many changes to the code itself - a couple settings here and there, one workaround in the custom OIOCMISQuerySet and then some test changes here and there. The bulk of the python file modifications are caused by the updated `black` version.

Most of the dependencies (90%?) are upgraded to their latest versions. Some I haven't touched because we've seen finicky behaviour with them before in terms of changing the API behaviour (drf-camel-case, drf-yags... etc.) and I'd like to keep that out of this PR. Testsuite is passing. We should have another issue to check if the documented patch versions of the API actually match up with the VNG versions too, and then check on test.openzaak.nl after merging this to main what the the api-test.nl results are. There are some known failures at the moment, so we should look into fixing those as well, but that's again another ticket.

- Bumped Django
- Bumped vng-api-common
- Bumped cmis-adapter
- Updated django-admin-index to 2.0.0 and fixed/cleaned up the style overrides
- Updated many more dependencies to newer versions to support Django 3.2
- Updated test/CI dependencies
- ~~Updated to latest isort and black and reformatted the code~~
- Updated changelog for upcoming 1.6 release

**TODO**

- [x] Click through all the pages and check for issues


CHANGELOG.rst
```diff
@@ -1,6 +1,73 @@
...
 =========
 
+1.6.0 (2022-03-??)
+------------------
+
+**New features**
+
+* Upgraded to Django 3.2 LTS version (#1098)
+* Upgraded to zgw-consumers 0.18.0, which supports client/server certificate
+  configuration for external services. This effectively adds Mutual TLS capabilities to
```
joeribekker, line 11: Is this true? I don't think this allows you to do certificate checking by the Zaken, Documenten, etc. API's themselves, right?

sergei-maertens: it should work for connections to BAG, BRP etc. that can be defined under the "external services":

image

joeribekker: Yes but the last sentence implies Zaken API etc have it as well

sergei-maertens: I'll make sure to reword it!

sergei-maertens:
@joeribekker release notes have been updated and I've removed the black & isort upgrades

joeribekker (approved):

Issue #1109, opened 2022-03-08
As developer, I want Open Zaak to support ZGW APIs 1.1


stevenbal:
**Changes per API**

Zaken API:
- [ ] ZaakContactMomenten resource added
- [ ] KlantContact resource deprecated
- [ ] ZaakVerzoek resource added
- [ ] ETag caching and HEAD operation added for:
  - [ ] Zaak
  - [ ] ZaakEigenschap
  - [ ] Resultaat
  - [ ] Rol
  - [ ] Status
  - [ ] ZaakInformatieObject
  - [ ] ZaakContactMoment
  - [ ] ZaakVerzoek
- [ ] `zaken.geforceerd-bijwerken` removed?
  - [ ] klantcontact
  - [ ] resultaten
  - [ ] rollen
  - [ ] ZIO POST
  - [ ] zaakobjecten
  - [ ] zaakeigenschap POST
- [ ] rol__betrokkenetype filter removed?
- [ ] zaak POST/PUT description changed?
- [ ] archiefnominatie/archiefstatus enum removed? max_length 1?

Catalogi API:
- [ ] ETag caching added

Documenten API:
- [ ] ETag caching added

Besluiten API:
- [ ] ETag caching added

Autorisaties API:
- [ ] ...

stevenbal:
This PR will combine all the 1.1 changes: https://github.com/open-zaak/open-zaak/pull/1111

stevenbal:
@sergei-maertens when checking the diff for the API specs of Zaken API 1.0 compared to 1.1, I noticed several deletions/changes that might not be correct/intentional?

- `zaken.geforceerd-bijwerken` scope was deleted
- `Zaak.rol__betrokkenetype` filter was deleted (does seem to appear in 1.2.x https://github.com/VNG-Realisatie/gemma-zaken/blob/878c1fa16986f4424c50760d7a5269a763925d01/api-specificatie/zrc/1.2.x/openapi.yaml#L3001)
- In `zaak_create` description the following: `geldigheid zaaktype URL - de resource moet opgevraagd kunnen worden uit de Catalogi API en de vorm van een ZAAKTYPE hebben.` to  `zaaktype moet een geldige URL zijn.`
[...]

PR #1126, opened 2022-03-29
:sparkles: [#1109] Add caching to Besluiten API resources

stevenbal:
Issue #1109

**Changes**
* Implements caching for Besluiten API resources


stevenbal:
`ZaakListPerformanceTests` seems to be a bit flaky sometimes?

https://github.com/open-zaak/open-zaak/runs/5735551021?check_suite_focus=true

sergei-maertens:
This is nice, but I don't see this implemented in the upstream standard anywhere, or am I missing something?

sergei-maertens:
Checked with Steven, looks like Besluiten API 1.1.0 has not been formally released yet, the reference implementation is still at 1.0.1. So, closing this

PR #1140, opened 2022-04-12
Migrate auth-adfs-db config to mozilla-oidc-db

### LINKED ISSUE #1139: Migrate from django-auth-adfs-db to mozilla-django-oidc-db
Looks like django-auth-adfs-db is no longer maintained by isprojects, and we already have the generic mozilla-django-oidc-db library in our dependencies which works with Azure AD/ADFS as well.

This is related to system check `fields.W904` - with the EOL of Django 2.2, we can remove support for django.contrib.postgres.fields.JSONField and instead use django.db.models.JSONField

**Task**

1.7.0 release:

- [x] #1140 
- [x] Remove database table(s) #1142 
- [x] Update documentation #1142 

1.8.0 release:

- [x] Make data migration no-op
- [x] Remove dependency
- [x] Document that upgrade from 1.6.x to 1.8.x directly is not possible, and 1.7.0 needs to be executed if OIDC is relevant. Otherwise manual cleanup is required.

sergei-maertens:
Related to #1139

**Changes**

This copies over the OIDC configuration using the config form
from the mozilla-oidc-db library to derive the correct endpoints.

Tested against Azure AD configuration, unfortunately I don't have
an on-prem installation available for further tests.


src/openzaak/accounts/migrations/0006_migrate_from_auth_adfs_db.py
```diff
@@ -0,0 +1,67 @@
...
+    provider_config.load_config()
+
+    # copy configuration over
+    form = OpenIDConnectConfigForm(
+        instance=oidc_config,
+        data={
+            "enabled": adfs_config.enabled,
+            "oidc_rp_client_id": adfs_config.client_id,
+            "oidc_rp_client_secret": adfs_config.client_secret,
+            "oidc_rp_sign_algo": "RS256",
```
stevenbal, line 38: Just checking, is this the only supported algorithm for ADFS?

stevenbal (approved):

src/openzaak/accounts/migrations/0006_migrate_from_auth_adfs_db.py
sergei-maertens: not sure, but I was checking the documentation of Open Zaak where it describes that you should enter `RS256`. In django-auth-adfs I saw a couple more, see https://github.com/snok/django-auth-adfs/blob/master/django_auth_adfs/backend.py#L67

PR #1141, opened 2022-04-12
Update to Python 3.10

sergei-maertens:
**Changes**

Changed the base python version from 3.7 to 3.10. 3.7 is EOL in a about a year.


.github/workflows/ci.yml
```diff
@@ -272,7 +272,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v2-beta
         with:
-          node-version: '12'
+          node-version: '16'
```
stevenbal, line 275: Why does this specific step need node 16?

stevenbal (approved):

.github/workflows/ci.yml
sergei-maertens: I noticed it was not synced anymore with the node version used in the Dockerfile, so that's the primary reason. Secondary, newer node versions come with a newer npm version (npm 8) which uses a newer format for package-lock.json, which is used during `npm ci`.

PR #1142, opened 2022-04-12
Uninstall django auth adfs + document changes

sergei-maertens:
Relates to #1139

**Changes**

* Included script to remove the django-auth-adfs-db table & migration history
* Document the changes
    * Update ADFS docs with warning and point to generic OIDC docs
    * Add manual intervention to the release notes
* Added redirect from old callback URL to OIDC library callback URL. This *should* work and is intended to give people some time to deal with the administrative stuff to apply changes in the IAM team.
* Uninstall the auth backend, URLs... Only `INSTALLED_APPS` is kept for now so the migrations can run. In 1.8.0, we can fully remove the library.


Dockerfile
```diff
@@ -57,6 +57,8 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 WORKDIR /app
 COPY ./bin/docker_start.sh /start.sh
 COPY ./bin/reset_migrations.sh /app/bin/reset_migrations.sh
+COPY ./bin/uninstall_adfs.sh ./bin/uninstall_django_auth_adfs_db.sql /app/bin/
```
stevenbal, line 60: Should this already be included in the docker image for 1.7.0, if it should only be used for version 1.8.x and onward?

stevenbal (approved):

Dockerfile
sergei-maertens: I wanted to add it so we don't forget about it for 1.8 - so now it's basically undocumented.

If it's executed by accident, then migrations can run again and will re-create the tables (I presume), so running the script again then should be sufficient.

stevenbal: Fair enough, it won't do any harm I think :+1:

PR #1143, opened 2022-04-12
Feature/1109 zaken api etags

sergei-maertens:
Related to #1109

**Changes**

* Added ETag DB model field through mixin to models
* Added `@conditional_retrieve()` decorator to viewsets

TODO:

- [x] tests
- [x] sort out missing schema/model changes that may be wrong in reference implementation

Some things are unclear at the moment - the reference implementation code does not seem to match the API schema?

-> had contact with Michiel from VNG, and the reference implementations/docs will receive a bugfix to remove the API spec mistakes. The behaviour of the reference implementation is as intended.


sergei-maertens:
@stevenbal I need to investigate the failure cause

sergei-maertens:
Failing build is caused by bug in vng-api-common, with the pending fix https://github.com/VNG-Realisatie/vng-api-common/pull/200

src/openzaak/components/zaken/migrations/0009_remove_zaakobject__etag.py
```diff
@@ -0,0 +1,16 @@
+# SPDX-License-Identifier: EUPL-1.2
+# Copyright (C) 2022 Dimpact
+# Generated by Django 3.2.13 on 2022-04-22 13:34
+
+from django.db import migrations
```
stevenbal, line 5: Can't this migration be squashed with `0008`?

stevenbal (approved):

src/openzaak/components/zaken/migrations/0009_remove_zaakobject__etag.py
sergei-maertens: sure

PR #1150, opened 2022-04-25
:construction_worker: Pull ZGW test suite from github in CI

stevenbal:
Issue: #1109 

**Changes**
* Retrieve the ZGW postman tests directly from github, instead of using ATP

Manual run of CI to check if it succeeds: https://github.com/open-zaak/open-zaak/actions/runs/2219210454

### DIFF
--- a/bin/postman_tests.sh
+++ b/bin/postman_tests.sh
@@ -29,7 +29,7 @@ until [ "$status" = "200" ]; do
 done
 
 # Download and execute the ZGW postman tests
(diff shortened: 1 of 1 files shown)


stevenbal:
@sergei-maertens I was checking whether there already was a testsuite for ZGW 1.1 on the test platform. But because it's such a mess, I think it might be better to just take the tests directly from github, because then we're sure it's the right version. What do you think?

sergei-maertens:
@stevenbal seems good to me

sergei-maertens (approved):

PR #1160, opened 2022-05-06
Protect fk relations (zaaktype -> don't delete zaken)

### LINKED ISSUE #979: Zaken should not be deleted when deleting the related zaaktype
**General information**
Open Zaak version: 1.4.0

**Describe the bug**
Resources between API's should be PROTECTED or SET_NULL*. Thus, we should only be allowed to delete a Zaaktypen if no cases are related.

\* SET_NULL is the most accurate because if Zaken en Catalogi were different components, the Catalogi API zaaktype would not have a db relation to a zaak in the Zaken API.

**To Reproduce**
1. Create Zaaktype
2. Create Zaak of above Zaaktype
3. Delete Zaaktype
4. Observe deleted Zaak

**Expected behavior**
Hmm, not sure :(

sergei-maertens:
Fixes #979

**Changes**

* Changed `on_delete=models.CASCADE` to `on_delete=models.PROTECT` for FK references to catalogi app models


stevenbal (approved):
Just checking: FK relations in the same API should still cascade right? E.g. deleting a `Zaak` should also delete related `deelzaken`, `zaakobjecten` etc.

sergei-maertens:
> Just checking: FK relations in the same API should still cascade right? E.g. deleting a `Zaak` should also delete related `deelzaken`, `zaakobjecten` etc.

Correct - and we have tests for that too :)

joeribekker (approved):

PR #1162, opened 2022-05-17
:zap: [#1105] Cache requests when running imports

### LINKED ISSUE #1105: Catalogus imports time out
For sufficiently large catalogus/zaaktypen imports through the admin interface, the import process takes longer than the (default) UWSGI timeout (? to confirm). There may also be an nginx timeout applicable. This makes the import unusable for any large enough real data.

edit: need to re-check this, might have been because of validation errors but no response data was sent

stevenbal:
Fixes #1105

**Changes**

* Use requests-cache to improve catalogus import performance


stevenbal:
I noticed that saving `ResultaatType` caused very slow imports, this might be a solution

src/openzaak/components/catalogi/management/commands/import.py
```diff
@@ -39,6 +40,8 @@ def add_arguments(self, parser):
 
     @transaction.atomic
     def handle(self, *args, **options):
+        backend = requests_cache.BaseCache()
+        requests_cache.install_cache("import", backend=backend)
```
sergei-maertens, line 44: Does this cache to filesystem or...? I'm going to need some more information :P

I'd also rather see this as a decorator or context manager that guarantees the cache is uninstalled again even if something unexpectedly throws exceptions. it seems this is globally installing the cache by monkeypatching, which lives across multiple requests in the uwsgi process. That also raises questions for me whether this is thread safe or not?

I think I'd also like to see this tied in with Django's cache framework so we can point to a cache backend from the django settings (which opens up the option to store stuff in memory / redis etc. down the road).

sergei-maertens (changes requested):

src/openzaak/components/catalogi/management/commands/import.py
stevenbal: I think the default is in-memory (https://requests-cache.readthedocs.io/en/stable/modules/requests_cache.backends.base.html), I'll look into tieing this into the existing Django cache framework

sergei-maertens: yes, redis is (by design) not thread-safe. We'd opt for a local memory cache in this case, or if a request object is available, on the request instance even (I believe we already have such a dirty hack)

src/openzaak/utils/cache.py
```diff
@@ -0,0 +1,107 @@
...
+
+from django.conf import settings
+from django.core.cache import caches
+
+import requests_cache
+from requests_cache import BaseCache
+from requests_cache.backends.base import KEY_FN
+from requests_cache.cache_keys import create_key
+
+DESERIALIZE_ERRORS = (
```
sergei-maertens, line 15: is this used?

stevenbal: nope, removing it

src/openzaak/components/catalogi/management/commands/import.py
stevenbal: I switched to LocMemCache

sergei-maertens (approved):
:zap: gotta go fast

PR #1163, opened 2022-05-18
Allow updating a zaaktype with relations to published documenttypes

### LINKED ISSUE #983: Not possible to update concept case type which is related to published information object types
**General information**
Open Zaak version: 1.4.0

**Describe the bug**
It is not possible to update concept case type using PUT/PATCH methods when case type is connected to already published information object types.
[Here ](https://vng-realisatie.github.io/gemma-zaken/standaard/catalogi/index#ztc-010) are the rules describing this behavior and while it is true that objects related to non concept objects should not be modified, the exception is in case type and object information type relation.

**To Reproduce**
1. Create new information object type (informatieobjecttypen) and publish it
2. Create new case type (zaakType)
3. Create new case type and information object type relation (zaaktype-informatieobjecttypen) with newly created objects
4. Try update case type using PATCH/PUT endpoints, error is returned: 'Objects related to non-concept informatieobjecttypen can't be updated'

**Expected behavior**
Case type gets updated, even if it has connections to already published information object types.

sergei-maertens:
Fixes #983

**Changes**

Updated validator to not check for related published documenttypes (and modified the test), as this falls under the exception for ZTC-10 and ZTC-11 rules.

I've also checked the behaviour for creating those relations and think we cover it correctly with the tests/validators.

I am however curious if this bug is also present in the VNG test suite, so I'm expecting potential build failures there. If that's the case, we need to get that test-suite fixed as well.

### DIFF
--- a/src/openzaak/components/catalogi/api/serializers/zaaktype.py
+++ b/src/openzaak/components/catalogi/api/serializers/zaaktype.py
@@ -193,7 +193,7 @@ class ZaakTypeSerializer(
             RelationCatalogValidator("besluittypen"),
             ConceptUpdateValidator(),
             M2MConceptCreateValidator(["besluittypen", "informatieobjecttypen"]),
-            M2MConceptUpdateValidator(["besluittypen", "informatieobjecttypen"]),
+            M2MConceptUpdateValidator(["besluittypen"]),
             DeelzaaktypeCatalogusValidator(),
             VerlengingsValidator(),
         ]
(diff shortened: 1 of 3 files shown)


joeribekker (approved):
> I am however curious if this bug is also present in the VNG test suite, so I'm expecting potential build failures there. If that's the case, we need to get that test-suite fixed as well.

I guess you're right.

sergei-maertens:
> > I am however curious if this bug is also present in the VNG test suite, so I'm expecting potential build failures there. If that's the case, we need to get that test-suite fixed as well.
> 
> I guess you're right.

I hate it when I'm right :smile:

stevenbal (approved):

PR #1164, opened 2022-05-18
Improve handling of selectielijst in the catalogi admin

### LINKED ISSUE #1030: Resultaattype selectielijst is "sometimes" incorrect in what it lists.
**General information**
Open Zaak version: 1.4.1

**Describe the bug**
Create a Resultaattype from scratch shows ALL resultaten from the Selectielijst (including from all years since there's no attached zaaktype yet, where the year is configured)

image

**To Reproduce**
1. Create Resultaattype
2. Scroll to Selectielijstklasse
3. Inspect the list of resultaten showing duplicates sometimes (names are to be expected but not the URL)

This is, I think, an underlying issue to the error "This selectielijstklasse is not a valid choice" which happens when a Zaaktype IS actually connected. I suspect cache or cache keys being set incorrectly, or the admin-voodoo-magic (comment in code) to get the zaaktype is going wrong.

image

I think what helps alot is some Javascript that refreshes this list based on the selected zaaktype

**Expected behavior**
The list of selectieklasses should be correct.

sergei-maertens:
~~Depends on #1161~~

Fixes #1030

**Changes**

* The list of possible selectielijstklassen is now only shown if the zaaktype has a selectielijst procestype specified. If it hasn't (or the zaaktype is not known yet), then the field is disabled and an empty value can be stored
* Publishing a zaaktype now validates that all resultaattypen have a selectielijstklasse specified. Combined with the previous change, this allows a more iterative configuration without losing any work.
* The zaaktype in the "admin voodoo" is now also grabbed from the POST data instead of only GET parameter
* The possible options for selectielijstklasse are now refreshed/rendered based on the chosen zaaktype. Changing the zaaktype causes the relevant list to be loaded
* Refresh list of procestypen in dropdown when the year changes
* Ensure the selectielijst configuration is respected w/r to defaults if the zaaktype has no year specified

### DIFF
--- a/src/openzaak/components/catalogi/api/admin/views.py
+++ b/src/openzaak/components/catalogi/api/admin/views.py
@@ -1,0 +1,37 @@
+# SPDX-License-Identifier: EUPL-1.2
+# Copyright (C) 2022 Dimpact
+from django.utils.translation import gettext_lazy as _
+
+from rest_framework import authentication, permissions
+from rest_framework.exceptions import ValidationError
+from rest_framework.response import Response
+from rest_framework.views import APIView
+
+from openzaak.selectielijst.admin_fields import get_selectielijst_resultaat_choices
+
+from ...admin.forms import EMPTY_SELECTIELIJSTKLASSE_CHOICES
+from ...models import ZaakType
+
+
+class SelectielijstResultatenListView(APIView):
+    authentication_classes = (authentication.SessionAuthentication,)
+    permission_classes = (permissions.IsAdminUser,)
+    schema = None  # keep it undocumented
+
+    def get(self, request, *args, **kwargs):
+        try:
+            zaaktype_id = int(request.GET.get("zaaktype", ""))
+        except ValueError:
+            raise ValidationError(
+                {"zaaktype": _("Provide a valid zaaktype ID to filter for")}
+            )
+
+        zaaktype = ZaakType.objects.filter(pk=zaaktype_id).first()
+        if zaaktype is None:
+            raise ValidationError({"zaaktype": _("Could not find the zaaktype")})
+
+        if not (url := zaaktype.selectielijst_procestype):
+            return Response(list(EMPTY_SELECTIELIJSTKLASSE_CHOICES))
+
+        choices = get_selectielijst_resultaat_choices(url)
+        return Response(choices)
--- a/src/openzaak/components/catalogi/api/admin/views.py
+++ b/src/openzaak/components/catalogi/api/admin/views.py
@@ -8,6 +8,8 @@ from rest_framework.response import Response
 from rest_framework.views import APIView
 
 from openzaak.selectielijst.admin_fields import get_selectielijst_resultaat_choices
+from openzaak.selectielijst.api import get_procestypen
+from openzaak.selectielijst.models import ReferentieLijstConfig
 
 from ...admin.forms import EMPTY_SELECTIELIJSTKLASSE_CHOICES
 from ...models import ZaakType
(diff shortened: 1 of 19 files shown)


src/openzaak/components/catalogi/api/admin/views.py
```diff
@@ -0,0 +1,37 @@
...
+from rest_framework.response import Response
+from rest_framework.views import APIView
+
+from openzaak.selectielijst.admin_fields import get_selectielijst_resultaat_choices
+
+from ...admin.forms import EMPTY_SELECTIELIJSTKLASSE_CHOICES
+from ...models import ZaakType
+
+
+class SelectielijstResultatenListView(APIView):
```
stevenbal, line 16: Not sure what causes the issue, but it seems that the active values of `ZaakType.selectielijst_procestype_jaar` and `ZaakType.selectielijst_procestype` are not properly displayed, they always seem to show the defaults from the config instead of the actual values on the `ZaakType`

src/openzaak/js/components/admin/zaaktypen/selectielijst.js
```diff
@@ -0,0 +1,93 @@
...
+    if (error) {
+        console.error(error);
+        return `Error: ${error}`;
+    }
+
+    let options;
+    if (loading) {
+        options = <option value="">Loading...</option>;
+    } else {
+        // TODO: the selected attribute should work via defaultValue, except it doesn't and I can't figure out why not
```
stevenbal, line 34: I tried this locally, it seems that `defaultValue` expects an object of the same shape as the elements in `options`. Something like this should work

```javascript
let initialOption = {};
if(options.constructor === Array) initialOption = options.find(option => option.key === initialValue)
...
<select defaultValue={initialOption} name="selectielijst_procestype" id="id_selectielijst_procestype">
```

src/openzaak/js/components/admin/zaaktypen/selectielijst.js
```diff
@@ -0,0 +1,93 @@
...
+// Copyright (C) 2020 Dimpact
+import React from 'react';
+import ReactDOM from 'react-dom';
+import PropTypes from 'prop-types';
+import useAsync from 'react-use/esm/useAsync';
+
+const PROCESTYPEN_ENDPOINT = '/admin/api/v1/catalogi/selectielijst/procestypen';
+
+
+const ProcestypeOptions = ({ year, initialValue='' }) => {
```
stevenbal, line 11: I noticed locally that this initial value doesn't seem to change when changing the `year`, seems like a bug?

stevenbal (changes requested):

src/openzaak/js/components/admin/zaaktypen/selectielijst.js
sergei-maertens: as discussed - this was on purpose to "freeze" the current value in case you reset the year again to what was stored

src/openzaak/components/catalogi/api/admin/views.py
```diff
@@ -0,0 +1,66 @@
...
+        choices = get_selectielijst_resultaat_choices(url)
+        return Response(choices)
+
+
+class SelectielijstProcestypenListView(APIView):
+    authentication_classes = (authentication.SessionAuthentication,)
+    permission_classes = (permissions.IsAdminUser,)
+    schema = None  # keep it undocumented
+
+    def get(self, request, *args, **kwargs):
```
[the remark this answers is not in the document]
sergei-maertens: discussed in the office that the current situation is the least bad of all options - the fields are required and depend on each other, so allowing blank options just increases complexity.

src/openzaak/js/components/admin/zaaktypen/selectielijst.js
sergei-maertens: nasty react state/props bug, now fixed by forcing a `key` that changes with the state.

stevenbal (approved):

PR #1165, opened 2022-05-19
Allow null eindegeldigheid in catalogi api

### LINKED ISSUE #981: Non concept object type's, property 'eindeGeldigheid' can't be set to null
**General information**
Open Zaak version: 1.4.0

**Describe the bug**
'eindeGeldigheid' field for published object types (case type, information object type and etc) can't be set to 'null' value when doing patch object type request.
[Here](https://vng-realisatie.github.io/gemma-zaken/standaard/catalogi/index#ztc-009) it is specified that this field is allowed to be updated for published object types, and I would assume it should accept all values supported by it including 'null'.

**To Reproduce**
1. Have a published object type
2. Do a patch request on a object type with body:
`{
    "eindeGeldigheid": null
}`
3. Error (sometimes errors, depending on object type) returned indicating that concept object can't be updated.

**Expected behavior**
'eindeGeldigheid' field set to null value.

sergei-maertens:
Fixes #981

**Changes**

* Added test to check the value can be set to null
* Added test to check that changes introducing overlap are still blocked
* Made validation error field context-dependent (if we're patching just the field eindeGeldigheid -> then errors are attached there)
* Fixed bug that prevented patches setting eindeGeldigheid to `None` from succeeding

TODO: backport to reference implementation (probably)

### DIFF
--- a/src/openzaak/components/catalogi/api/validators.py
+++ b/src/openzaak/components/catalogi/api/validators.py
@@ -10,6 +10,7 @@ from vng_api_common.constants import (
 )
 
 from openzaak.client import fetch_object
+from openzaak.utils.serializers import get_from_serializer_data_or_instance
 
 from ..constants import SelectielijstKlasseProcestermijn as Procestermijn
 from ..utils import has_overlapping_objects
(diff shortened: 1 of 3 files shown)


sergei-maertens:
Backport fix submit in https://github.com/VNG-Realisatie/catalogi-api/pull/134

PR #1176, opened 2022-05-31
 :bug: [#1169] Derive selectielijst_procestype_jaar when saving ZaakTypen 

### LINKED ISSUE #1169: ZaakType import does not set `selectielijst_procestype_jaar`
**General information**
Open Zaak version: 1.6.0

**Describe the bug**
`ZaakType` has an internal field `selectielijst_procestype_jaar` to determine the possible choices for `selectielijst_procestype`. Because this field is not exposed via the API, it is not included when exporting a `ZaakType`, which causes this field to be `None` when importing a `ZaakType`.

This leads to incorrect information in the admin, because the `selectielijst_procestype` from the import could be from the `2020` selectielijst, but the `selectielijst_procestype_jaar` defaults to `2017`, causing a mismatch:

Example (the procestype is from 2020: https://selectielijst.openzaak.nl/api/v1/procestypen/0f86ece3-f863-4c67-928e-cc71ad0e2318):
Screenshot from 2022-05-23 15-32-31

**To Reproduce**
1. Create a `ZaakType` with some `selectielijst_procestype` from `2020`
2. Export this `ZaakType` and delete it
3. Import this `ZaakType` and view it in the admin
4. Scroll down to the `Gemeentelijke selectielijst` and observe the selected `selectielijst_procestype_jaar`

**Expected behavior**
The import functionality should derive the value of `selectielijst_procestype_jaar` by retrieving the `selectielijst_procestype` and taking the `jaar` from its response.

Additionally, there should probably be a validation in the admin that checks whether `selectielijst_procestype` and `selectielijst_procestype_jaar` actually match.

stevenbal:
Fixes #1169

**Changes**

* Derive `ZaakType.selectielijst_procestype_jaar` when saving
* Add data migration to do this derivation for existing data


src/openzaak/components/catalogi/models/zaaktype.py
```diff
@@ -297,6 +298,10 @@ def save(self, *args, **kwargs):
         if not self.pk:
             transaction.on_commit(AutorisatieSpec.sync)
 
+        if self.selectielijst_procestype and not self.selectielijst_procestype_jaar:
+            response = requests.get(self.selectielijst_procestype).json()
```
sergei-maertens, line 302: Please use the `Client` abstraction like in https://github.com/open-zaak/open-zaak/blob/main/src/openzaak/components/catalogi/admin/forms.py#L150 to do the actual retrieval - this future-proofs us for proxies like NLX and allows us to configure authorization if it's relevant (for example, because the API gateway requires it).

Raw `requests` usage is pretty much never needed.

sergei-maertens (changes requested):

sergei-maertens (approved):
We really need to clear up the mocking interfaces used, ~~there's also `openzaak.selectielijst.tests.mixins.SelectieLijstMixin` and the `mock_resource_list` and `mock_resource_get` functions that work with fixtures.~~

PR #1181, opened 2022-05-31
:bug: [#1182] Fix minor catalogus/zaaktype import issues

stevenbal:
Fixes: https://github.com/open-zaak/open-zaak/issues/1182

**Changes**
* Use requests-cache for zaaktype imports (previously missing)
* Fix the admin layout for custom import pages
* Fix local issue causing all caches to be flushed after performing an import

### DIFF
--- a/src/openzaak/components/catalogi/admin/admin_views.py
+++ b/src/openzaak/components/catalogi/admin/admin_views.py
@@ -24,7 +24,22 @@ from .utils import (
 )
 
 
-class CatalogusZaakTypeImportUploadView(PermissionRequiredMixin, FormView):
+class AdminContextMixin:
+    """
+    Update custom admin views with all the context info
+    """
+
+    admin_site = None
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context.update(self.admin_site.each_context(self.request))
+        return context
+
+
+class CatalogusZaakTypeImportUploadView(
+    AdminContextMixin, PermissionRequiredMixin, FormView
+):
     template_name = "admin/catalogi/import_zaaktype.html"
     form_class = ZaakTypeImportForm
     permission_required = "catalogi.add_zaaktype"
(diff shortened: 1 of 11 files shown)


src/openzaak/components/catalogi/admin/admin_views.py
```diff
@@ -54,9 +71,12 @@ def post(self, request, *args, **kwargs):
...
                     with transaction.atomic():
-                        import_zaaktype_for_catalogus(
-                            catalogus_pk, request.session["file_content"], {}, {}
-                        )
+                        with requests_cache_enabled(
+                            "import", backend=DjangoRequestsCache()
+                        ):
+                            import_zaaktype_for_catalogus(
+                                catalogus_pk, request.session["file_content"], {}, {}
+                            )
```
sergei-maertens, line 79: shouldn't we apply that context manager inside the `import_zaaktype_for_catalogus` function instead, or apply it as a decorator?

src/openzaak/components/catalogi/admin/admin_views.py
```diff
@@ -22,7 +24,22 @@
 )
 
 
-class CatalogusZaakTypeImportUploadView(PermissionRequiredMixin, FormView):
+class AdminContextMixin:
```
sergei-maertens, line 27: Please put this in `openzaak.utils.admin` so that we don't end up with 4 copies of the same thing in different components/apps again :grimacing:

sergei-maertens, line 27: Please put this in `openzaak.utils.admin` so that we don't end up with 4 copies of the same thing in different components/apps again :grimacing:

sergei-maertens (changes requested):

sergei-maertens (approved):

src/openzaak/components/catalogi/admin/admin_views.py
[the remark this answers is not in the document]
sergei-maertens: github seems to be a bit drunk

PR #1184, opened 2022-06-03
Cleaning up test cruft - first pass

sergei-maertens:
No ticket for this, but during the recent development work I noticed a lot of old patterns were used in the tests blocking our upgrade to gemma-zds-client stable. There are other aspects to fix (e.g. the existence of `openzaak.test.utils` and `openzaak.utils.tests` that are used seemingly random and the many OAS mocking solutions that have been re-implemented 3-4 times now...). That will follow in another PR, with more developer documentation for these aspects to fix the structure of the codebase.

**Changes**

* Replaced all usages of our custom `openzaak.utils.tests.mock_client` - which was a copy of `zds_client.tests.mock_client` with a minor change - with `requests_mock.Mocker()`
* Removed `openzaak.utils.tests.mock_client`
* Added a bunch of TODOs for test utilities all doing the same. The plan is to use the zgw-consumers test utility instead.
* Upgraded gemma-zds-client from 0.15.0 to 1.0.1
* Refactored some internal test utilities to use gemma-zds-client tooling instead of duplicating the implementation


src/openzaak/components/catalogi/tests/test_resultaattype.py
```diff
@@ -438,23 +428,18 @@ def test_derive_archiefactiedatum_from_selectielijstklasse(self, *mocks):
...
         }
 
-        responses = {
-            SELECTIELIJSTKLASSE_URL: {
-                "url": SELECTIELIJSTKLASSE_URL,
-                "procesType": PROCESTYPE_URL,
-                "procestermijn": Procestermijn.nihil,
-            }
-        }
+        with requests_mock.Mocker() as m:
```
stevenbal, line 431: Is there a preferred way to use `requests_mock.Mocker`, decorator vs context manager?

stevenbal (approved):

src/openzaak/components/catalogi/tests/test_resultaattype.py
sergei-maertens: not really - in a number of cases I went with it because there are a number of `unittest.mock.patch` calls which are passed as `*mocks`, and the requests mock stuff is injected _after_ those as an `arg` rather than `kwargs` so that makes it annoying in terms of method signature. using the context manager is then simpler to not have to mess around with that.

I think I personally prefer the decorator because the context manager introduces a new level of nesting and I prefer code to be "as flat as possible".

PR #1188, opened 2022-06-07
:bug: [#1042] Disallow mutating published ZaakTypen via the admin

### LINKED ISSUE #1042: The admin does not always block mutating published zaaktypen
**General information**
Open Zaak version: 1.4.x / 1.5.x

**Describe the bug**

It's possible via work-arounds to create new Eigenschappen (and possibly also statustypen, rollen...) in the catalogi admin interface for a zaaktype that is already published. This breaks the expectation that zaaktypen can not change after publication, as dictated by the standard.

The API _does_ block this however, this is limited to the admin interface.

**To Reproduce**
1. Create a zaaktype
2. Publish the zaaktype
3. Click the "Toon eigenschappen"
4. Click "Eigenschap toevoegen" in the top right
5. Fill out the details and submit - the eigenschap is now added for a published zaaktype which should not be allowed.

**Expected behavior**

A validation error should prevent the eigenschap from being saved. This should also be checked for all other sub-resources.

stevenbal:
Fixes #1042 

**Changes**

* Disallow mutating published ZaakTypen via the admin
* Fix change form incorrectly displaying as readonly

### DIFF
--- a/src/openzaak/components/catalogi/models/eigenschap.py
+++ b/src/openzaak/components/catalogi/models/eigenschap.py
@@ -206,13 +206,15 @@ class Eigenschap(ETagMixin, models.Model):
             "URL-referentie naar het ZAAKTYPE van de ZAAKen waarvoor deze EIGENSCHAP van belang is."
         ),
         on_delete=models.CASCADE,
-        validators=[validate_zaaktype_concept],
     )
 
     class Meta:
         unique_together = ("zaaktype", "eigenschapnaam")
         verbose_name = _("Eigenschap")
         verbose_name_plural = _("Eigenschappen")
+
+    def clean(self):
+        validate_zaaktype_concept(self.zaaktype)
 
     def __str__(self):
         return self.eigenschapnaam
--- a/src/openzaak/components/catalogi/models/eigenschap.py
+++ b/src/openzaak/components/catalogi/models/eigenschap.py
@@ -11,7 +11,11 @@ from django_better_admin_arrayfield.models.fields import ArrayField
 from vng_api_common.caching import ETagMixin
 
 from ..constants import FormaatChoices
-from .validators import validate_kardinaliteit, validate_letters_numbers_underscores
+from .validators import (
+    validate_kardinaliteit,
+    validate_letters_numbers_underscores,
+    validate_zaaktype_concept,
+)
 
 
 class EigenschapSpecificatie(models.Model):
@@ -202,6 +206,7 @@ class Eigenschap(ETagMixin, models.Model):
             "URL-referentie naar het ZAAKTYPE van de ZAAKen waarvoor deze EIGENSCHAP van belang is."
         ),
         on_delete=models.CASCADE,
+        validators=[validate_zaaktype_concept],
     )
 
     class Meta:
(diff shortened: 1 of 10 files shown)


src/openzaak/components/catalogi/tests/admin/test_eigenschap_admin.py
```diff
@@ -49,3 +51,42 @@ def test_validation_length_comma_separated(self):
 
         new_specificatie = EigenschapSpecificatie.objects.get()
         self.assertEqual(new_specificatie.lengte, "5,3")
+
+
+@tag("issue-1042")
```
sergei-maertens, line 56:
```suggestion
@tag("gh-1042")
```
let's see if we can stick to the pattern of tags used in other places :P

src/openzaak/components/catalogi/tests/admin/test_resultaattype_admin.py
```diff
@@ -536,3 +540,49 @@ def test_selectielijstklasse_correctly_prefiltered(self, m):
...
+        error = response.html.find("div", {"class": "field-zaaktype"}).find(
+            "ul", {"class": "errorlist"}
+        )
+
+        self.assertEqual(
+            error.text,
+            _("Creating a relation to non-concept {resource} is forbidden").format(
+                resource="zaaktype"
+            ),
+        )
```
sergei-maertens, line 587: isn't it easier to check `response.context["adminform"].errors`?

src/openzaak/components/catalogi/tests/admin/test_eigenschap_admin.py
```diff
@@ -49,3 +51,42 @@ def test_validation_length_comma_separated(self):
...
+class EigenschapAdminTests(WebTest):
+    @classmethod
+    def setUpTestData(cls):
+        cls.user = SuperUserFactory.create()
```
sergei-maertens, line 60: _always_ call the super method

src/openzaak/components/catalogi/models/mixins.py
```diff
@@ -119,3 +119,32 @@ class Meta:
...
+            return
+
+        if not getattr(self, self.resource_name).concept:
+            raise ValidationError(
+                {
+                    self.resource_name: _(
+                        "Creating a relation to non-concept {resource_name} is forbidden"
+                    ).format(resource_name=self.resource_name)
+                }
+            )
```
sergei-maertens, line 146: Can't we implement that as a validator on the FK field instead?

```py
class StatusType(...):
    zaaktype = models.ForeignKey("ZaakType", ..., validators=[validate_concept_status])
```

stevenbal: You're right, I wasn't sure about this mixin approach

sergei-maertens:
The failing tests may be because the validators kick in in DRF after all? :thinking:

stevenbal:
> The failing tests may be because the validators kick in in DRF after all? thinking

It seems that the model field validation is run before the serializer validation

src/openzaak/components/catalogi/models/eigenschap.py
```diff
@@ -209,5 +213,8 @@ class Meta:
         verbose_name = _("Eigenschap")
         verbose_name_plural = _("Eigenschappen")
 
+    def clean(self, *args, **kwargs):
```
sergei-maertens, line 216:
```suggestion
    def clean(self):
```

sergei-maertens (changes requested):
`models.Model.clean` does not accept any args or kwargs

PR #1208, opened 2022-07-04
:sparkles: Use notifications-api-common library

stevenbal:
Requires following PRs to be merged first:
- https://github.com/VNG-Realisatie/vng-api-common/pull/206
- https://github.com/maykinmedia/notifications-api-common/pull/1 

TODO:
- [x] Pin notifications-api-common to PyPI version
- [x] Pin vng-api-common to PyPI version


docs/conf.py
```diff
@@ -85,6 +85,7 @@
     r"https://.*sentry\.openzaak\.nl.*",
     # TODO temporary workaround for issue #6218 with Pleio
     r"https://commonground.nl/groups/view/d9c2f667-2f3e-4153-a79b-57dde7f56cc2/team-open-zaak",
+    r"https://passwordsgenerator.net",
```
sergei-maertens, line 88: We should replace that link - the site seems to not exist any longer

docs/installation/reference/cli.rst
```diff
@@ -30,6 +30,6 @@ Available commands
     After configuring the Notificaties API, send a test nofification to verify the
     setup.
 
-``register_kanaal``
-    Registers a notifications channel with the notifications API if it doesn't exist
+``register_kanalen``
+    Registers notifications channels with the notifications API if it doesn't exist
```
sergei-maertens, line 34:
```suggestion
    Registers notification channels with the notifications API that don't exist
```

src/openzaak/notifications/tests/utils.py
```diff
@@ -35,13 +35,13 @@
 
 class NotificationsConfigMixin:
```
sergei-maertens, line 36: This should probably be merged with the mixin from `__init__.py`, and the classmethod to configure should be invoked on `setUpTestData` automatically I think. Only transaction test case then needs special treatment.

src/openzaak/components/zaken/tests/test_notifications_kanaal.py
```diff
@@ -23,60 +25,119 @@ def setUpTestData(cls):
...
         )
 
-        client.create.assert_called_once_with(
-            "kanaal",
-            {
-                "naam": "dummy-kanaal",
-                "documentatieLink": "https://example.com/ref/kanalen/#dummy-kanaal",
-                "filters": [],
-            },
+        client.create.assert_has_calls(
```
sergei-maertens, line 68: This test now seems to be duplicated a couple of times - imo it should be moved just to the non-component location (`openzaak.notifications` seems appropriate) and removed in the other places.

sergei-maertens (changes requested):

sergei-maertens (approved):

Issue #1223, opened 2022-07-19
As user of Open Zaak, I want Open Zaak to support Documenten API 1.1


stevenbal:
... so I can use the features described in https://github.com/VNG-Realisatie/documenten-api/blob/master/CHANGELOG.rst#110-2022-06-30

**Description**
Features from Documenten API 1.1 should be added

Features:
- [x] Caching https://github.com/VNG-Realisatie/gemma-zaken/issues/1096, [PR in reference implementation](https://github.com/VNG-Realisatie/documenten-api/pull/88)
  - https://github.com/open-zaak/open-zaak/pull/1252
- [x] Large file uploads https://github.com/VNG-Realisatie/gemma-zaken/issues/1078, [PR in reference implementation](https://github.com/VNG-Realisatie/documenten-api/commit/b83efeef1db2bb3bb9e1134c2fd1e776cfe262c4)
  - https://github.com/open-zaak/open-zaak/pull/1240
- [x] Add verzoek type to possible ObjectInformatieObject.object_type and kcc to component enum, [PR](https://github.com/VNG-Realisatie/documenten-api/pull/104)
  - https://github.com/open-zaak/open-zaak/pull/1226

stevenbal:
@sergei-maertens documenten 1.1 wasn't implemented in a fork by any chance right? :p https://github.com/open-zaak/open-zaak/issues/1218#issuecomment-1192708411

sergei-maertens:
Not that I'm aware of, but you can see the open PRs yourself :P

stevenbal:
@sergei-maertens question regarding `BestandsDelen` and CMIS: I assume we should store BestandsDelen in the connected DMS as well (requiring a new resource in the CMIS model)? Since it would not be possible to `.save()` a `BestandDeel` to the Django DB, because the related `informatieobject` is unsaved as well

sergei-maertens:
> @sergei-maertens question regarding `BestandsDelen` and CMIS: I assume we should store BestandsDelen in the connected DMS as well (requiring a new resource in the CMIS model)? Since it would not be possible to `.save()` a `BestandDeel` to the Django DB, because the related `informatieobject` is unsaved as well

[...]

PR #1226, opened 2022-07-20
:sparkles: [#1223] Add verzoek option to ObjectInformatieObject

stevenbal:
Issue #1223

**Changes**

* Add `verzoek` option to ObjectInformatieObject

Requires the following PRs to be merged first (tests are failing because of this):
- https://github.com/open-zaak/alfresco-content-model/pull/3
- https://github.com/open-zaak/cmis-adapter/pull/69


src/openzaak/components/documenten/query/cmis.py
```diff
@@ -773,6 +774,7 @@ def process_filters(self, data):
                     list_value.append(make_absolute_uri(reverse(item)))
                 value = list_value
 
+            # FIXME
```
sergei-maertens, line 777: should we create a ticket for this?

src/openzaak/components/documenten/query/cmis.py
```diff
@@ -1119,7 +1126,19 @@ def format_zaak_and_zaaktype_data(
...
+    elif verzoek_url:
+        client = Service.get_client(verzoek_url)
+        objectverzoeken = client.list("objectverzoek", {"verzoek": verzoek_url})
+        zaakverzoeken = [
+            objectverzoek
+            for objectverzoek in objectverzoeken
+            if objectverzoek["objectType"] == "zaak"
+        ]
+        if zaakverzoeken:
+            # FIXME what if there is more than one objectverzoek? Is this even possible?
```
sergei-maertens, line 1138: this is where I'd expect a link to the standard :P

src/openzaak/components/documenten/query/django.py
```diff
@@ -92,14 +93,15 @@ class InformatieobjectRelatedQuerySet(
...
-        ZaakInformatieObject: ObjectTypes.zaak,
+        BesluitInformatieObject: ObjectInformatieObjectTypes.besluit,
+        ZaakInformatieObject: ObjectInformatieObjectTypes.zaak,
     }
 
     def create_from(self, relation: IORelation) -> [models.Model, None]:
         if isinstance(relation.informatieobject, ProxyMixin):
             return None
 
+        # FIXME?
```
sergei-maertens, line 104: not sure what is happening here - @SilviaAmAm, can you take a look if this makes sense?

src/openzaak/components/documenten/tests/test_objectinformatieobject.py
```diff
@@ -36,8 +35,13 @@
     get_zaak_response,
     get_zaakinformatieobject_response,
 )
-from openzaak.tests.utils import mock_brc_oas_get, mock_zrc_oas_get
+from openzaak.tests.utils import mock_brc_oas_get, mock_vrc_oas_get, mock_zrc_oas_get
+from openzaak.tests.utils.verzoeken import (
+    get_verzoek_response,
+    get_verzoekinformatieobject_response,
+)
```
sergei-maertens, line 42: perhaps this should go into `openzaak.contrib.verzoeken`?

In the future we can then move `openzaak.bag`, `openzaak.nlx`... in there as well.

sergei-maertens (changes requested):
Broad lines look fine to me :+1:

src/openzaak/components/documenten/query/cmis.py
stevenbal: I can't seem to find anything about it in https://github.com/VNG-Realisatie/gemma-zaken/blob/master/docs/_content/standaard/zaken/index.md or https://github.com/VNG-Realisatie/gemma-zaken/blob/master/docs/_content/standaard/verzoeken/index.md.

In the code there doesn't seem to be a restriction on this either

src/openzaak/components/documenten/query/cmis.py
stevenbal: Not necessary, I think I added this at the time because I wasn't sure if it would work, but because `value` is always a URL in the case of verzoeken, the `else` statement is fine for this case

src/openzaak/components/documenten/query/django.py
stevenbal: I don't think changes need to be made here, as this will never be triggered for `VerzoekInformatieObjecten` because only external verzoeken are supported

src/openzaak/components/documenten/query/cmis.py
sergei-maertens: I think we should seek clarification then - let's create an issue of the type "question" in the gemma-zaken repo.

src/openzaak/components/documenten/query/django.py
sergei-maertens: yeah, I saw the tests are passing, so that seems to confirm it. I see the PR comment is outdated now, but if not there yet, can you add that in a comment in the code?

stevenbal: added in https://github.com/open-zaak/open-zaak/pull/1226/commits/894903cb35fe0cb91ca7fc0029c34244151bd283

src/openzaak/components/documenten/query/cmis.py
stevenbal: https://github.com/VNG-Realisatie/gemma-zaken/issues/2071

stevenbal: @SilviaAmAm do you know why the zaak information is required for Alfresco when creating an ObjectInformatieObject?

src/openzaak/components/documenten/models.py
```diff
@@ -752,6 +756,8 @@ class ObjectInformatieObject(models.Model, CMISClientMixin):
         fk_field="_besluit", url_field="_besluit_url", blank=True, null=True,
     )
 
+    verzoek = models.URLField(_("extern verzoek"), blank=True, max_length=1000)
```
sergei-maertens, line 759: I think we can clean up this entire thing by only having one `_object_url` field - we can't guarantee referential integrity on that anyway. The `object_type` then determines how to interpret that URL, and we can simplify the DB constraints substantially and probably get rid of a lot of code.

src/openzaak/components/documenten/query/cmis.py
```diff
@@ -787,18 +788,22 @@ def create(self, **kwargs):
...
+
+        if len(related_objects) - num_empty_relations > 1:
             raise IntegrityError(
-                "ObjectInformatie object cannot have both Zaak and Besluit relation"
+                "ObjectInformatie object can only have one Zaak/Besluit/Verzoek relation"
             )
-        elif related_zaak_url is None and related_besluit_url is None:
+        elif len(related_objects) == num_empty_relations:
             raise IntegrityError(
                 "ObjectInformatie object needs to have either a Zaak or Besluit relation"
```
sergei-maertens, line 802: either zaak, besluit or verzoek relation

src/openzaak/components/documenten/query/cmis.py
sergei-maertens: I see in the cmis-adapter code that the document is then moved/copied to the related zaak folder so that in the file browser (in Alfresco for example), the document is listed with the zaak.

I think this isn't exactly correct for verzoeken, as the zaak is only tangentially related to the zaak. It would probably make more sense to have a directory structure for a verzoek en put the document there:

1. The verzoek is the primary object
2. There's no guarantee the verzoek is related to a zaak
3. One verzoek can lead to multiple zaken
4. The zaken can be created later _because of_ the verzoek, possibly through human action (as opposed to an automated system)

The relevant code is in `drc_cmis.client.CMISClient.create_oio`

sergei-maertens (changes requested):
If you don't mind, I'd like to follow up this PR tomorrow.

sergei-maertens (changes requested):
There's plenty of things in here that are less than ideal, but we'd need a complete overhaul for this.

Hopefully other vendors are getting their Documenten API implementation production ready soon, so the CMIS adapter can be deleted completely. Then we'll be able to clean up a lot of cruft, for now we'll have to live with it.

sergei-maertens (approved):

PR #1239, opened 2022-08-01
:wrench: Set DATA_UPLOAD_MAX_NUMBER_FIELDS to 10000

stevenbal:
to fix failing exports for large catalogi

I noticed that this setting was removed in https://github.com/open-zaak/open-zaak/pull/998, but when attempting to export a large catalogus, it raised issues again

### DIFF
--- a/src/openzaak/conf/includes/base.py
+++ b/src/openzaak/conf/includes/base.py
@@ -430,6 +430,12 @@ SILENCED_SYSTEM_CHECKS = [
     "debug_toolbar.W006",
 ]
 
+
+#
+# Increase number of parameters for GET/POST requests
+#
+DATA_UPLOAD_MAX_NUMBER_FIELDS = 10000
+
 #
 # Custom settings
 #


sergei-maertens:
Can we not do the exports in another way that we don't need this amount of post data fields? If you get a 10x as large catalogus, it'll be broken again. Why does exporting catalogi even require this amount of fields?

stevenbal:
> Why does exporting catalogi even require this amount of fields?

I'll look into it

stevenbal:
It seems that the formsets for related zaaktypen, etc. produce quite a lot of parameters for the POST request. For example, I was trying to export a catalogus with 448 zaaktypen and 60 informatieobjecttypen and apparently each *type adds two parameters to the POST request, resulting in more than the `DATA_UPLOAD_MAX_NUMBER_FIELDS` limit. This problem occurs for any post request, so the same issue will occur when trying to edit the catalogus metadata and saving.

sergei-maertens:
> It seems that the formsets for related zaaktypen, etc. produce quite a lot of parameters for the POST request. For example, I was trying to export a catalogus with 448 zaaktypen and 60 informatieobjecttypen and apparently each *type adds two parameters to the POST request, resulting in more than the `DATA_UPLOAD_MAX_NUMBER_FIELDS` limit. This problem occurs for any post request, so the same issue will occur when trying to edit the catalogus metadata and saving.
> 
> So we could look into a different way of exporting by circumventing this issue, but it will still be a problem for regular saves.

Yeah I think we have a bigger issue at hands here indeed and should work to solve the root cause. Parking this for now.

sergei-maertens:
Discussed and the only viable path is building a proper app to manage catalogi, for which there is no budget at the moment. So this will have to be it.

Issue #1209, opened 2022-07-05
Refactor references to API URLs to a configurable reference


joeribekker:
Currently, URLs are stored as-is, to other API's, for example:

`Zaak.zaaktype = https://www.example.com/subpath/api/v1/zaaktypen/6c3f87`

When the Catalogi API changes to domain or path, things start to break down. So, we need to refactor the url-fields to store a more normalized version of the URL (pseudo-code follows):

```python
class Zaak(models.Model):

    zaaktype_api = models.FK(Service)
    zaaktype_relative_url = models.CharField()

    @property
    def zaaktype(self):
        return urljoin(self.zaaktype_api.base_path, self.zaaktype_relative_url)
```

This way, we can change the `Service` and all URLs will be changed. Be aware though, that we need to figure out what happens if Open Zaak itself is moved from one URL to another. The LooseFK will probably work but I'm just not sure.

PS. This was roughly described a few years ago but never implemented: https://vng-realisatie.github.io/gemma-zaken/themas/achtergronddocumentatie/versies-en-migraties

PR #1242, opened 2022-08-04
Composite urls

annashamray:
Fixes https://github.com/open-zaak/open-zaak/issues/1209
Depends on https://github.com/maykinmedia/zgw-consumers/pull/52

**Changes**

replace url fields with composite ones for:
* besluiten
  * [x] `Besluit.besluittype`
  * [x] `Besluit.zaak`
  * [x]  `BesluitInformatieObject.informatieobject`
* documenten
  * [x] `InformatieObject.informatieobjecttype`
  * [x] `ObjectInformatieObject.zaak`
  * [x] `ObjectInformatieObject.besluit`
 * zaken
   * [x] `Zaak.zaaktype`
   * [x] `RelevanteZaakRelatie.url`
   * [x] `Status.statustype`
   * [x] `Resultaat.resultaattype`
   * [x] `Rol.roltype`
   * [x] `ZaakEigenschap.eigenschap`
   * [x] `ZaakInformatieObject.informatieobject`
   * [x] `ZaakBesluit.besluit`
* [x] CMIS support
* [x] clean after rebase

**Deployment**
Several things to care during deployment:
* Make sure all external resources used in loose-fk fields are added to `zgw_consumers.Service`. Otherwise the data migrations will fail
* Add all local component apis into `zgw_consumers.Service` (for example, if your host is https://example.openzaak.nl/ add https://example.openzaak.nl/zaken/api/v1/ and all other local components into `Service`)
* `zgw_consumers` was updated to 0.21, which has a breaking change concerning certificates. Make sure certificates work as expected
* If there are a loot of external loose-fk data (for example there are a lot of `ZaakInformatieObject` records with different external informatieobjects) the migration can take a while. The migration is a one-off thing and it was not optimized.


src/openzaak/components/besluiten/admin.py
```diff
@@ -38,15 +38,25 @@ class BesluitInformatieObjectAdmin(
...
+        "_informatieobject",
+        "_informatieobject_base_url",
+        "_informatieobject_relative_url",
+    )
     list_filter = ("besluit",)
     search_fields = (
         "besluit__uuid",
         "_informatieobject__enkelvoudiginformatieobject__uuid",
-        "_informatieobject_url",
+        "_informatieobject_relative_url",
```
sergei-maertens, line 51: can we do an annotation on the queryset that concacts the service root + relative URL, so we can still search on fully qualified URLs?

sergei-maertens (changes requested):
1. We can probably avoid a lot of the test modifications by updating the factories - keep passing fully qualified URLs, and in the factory (this can be a `@factory.post_generation` or custom subfactory construct) decompose the URL and set up the required `Service`
2. You can use `Factory.Meta.django_get_or_create` to deal with unique constraints if you're creating multiple objects from the same service

The admin tests probably will still need the explicit decomposition because of how it works, but maybe you can consider a cusotm form field that does the decomposition instead? I think that's a better presentation and user-experience anyway, as it seems unlikely you'd keep the relative URL but change the service of an object. If you're migrating from a staging -> production environment, it would make more sense to update the `Service` instance itself.

src/openzaak/components/besluiten/models.py
```diff
@@ -43,42 +45,73 @@ class Besluit(AuditTrailMixin, APIMixin, models.Model):
...
+        related_name="besluit_besluittypen",
+        help_text="Basis deel van URL-referentie naar het extern BESLUITTYPE (in een andere Catalogi API).",
+    )
+    _besluittype_relative_url = models.CharField(
+        _("besluittype relative url"),
+        max_length=200,
+        blank=True,
+        null=True,
+        help_text="Relatief deel van URL-referentie naar het extern BESLUITTYPE (in een andere Catalogi API).",
+    )
```
sergei-maertens, line 62: I think I'd go custom field (subclassing from `CharField`) here with an added validator to check that it is indeed a relative URL (so there may not be a protocol, netloc or leading slash).

src/openzaak/components/besluiten/models.py
```diff
@@ -43,42 +45,73 @@ class Besluit(AuditTrailMixin, APIMixin, models.Model):
...
-        help_text=_(
-            "URL-referentie naar extern BESLUITTYPE (in een andere Catalogi API)."
-        ),
+        on_delete=models.PROTECT,
+        related_name="besluit_besluittypen",
+        help_text="Basis deel van URL-referentie naar het extern BESLUITTYPE (in een andere Catalogi API).",
+    )
+    _besluittype_relative_url = models.CharField(
+        _("besluittype relative url"),
+        max_length=200,
```
sergei-maertens, line 58: this seems to short, given that the original URL field was able to store 1000 chars. `zgw_consumers.Service.api_root` can store 255 chars, with these added 200 chars we have maximum URL length of 455 chars.

src/openzaak/components/besluiten/models.py
```diff
@@ -43,42 +45,73 @@ class Besluit(AuditTrailMixin, APIMixin, models.Model):
...
+        null=True,
         blank=True,
-        max_length=1000,
-        help_text=_(
-            "URL-referentie naar extern BESLUITTYPE (in een andere Catalogi API)."
-        ),
+        on_delete=models.PROTECT,
+        related_name="besluit_besluittypen",
+        help_text="Basis deel van URL-referentie naar het extern BESLUITTYPE (in een andere Catalogi API).",
+    )
```
sergei-maertens, line 55: I think I'd consider making this a custom field as well that bakes in the defaults of `null=True, blank=True, on_delete=models.PROTECT, to="zgw_consumers.Service"`

src/openzaak/components/besluiten/models.py
```diff
@@ -177,7 +210,8 @@ def __init__(self, *args, **kwargs):
         # self._previous_zaak = self.zaak
 
         self._previous_zaak = self._zaak
-        self._previous_zaak_url = self._zaak_url
+        self._previous_zaak_base_url_id = self._zaak_base_url_id
+        self._previous_zaak_relative_url = self._zaak_relative_url
```
sergei-maertens, line 214: can't you use the model field machinery to just store the previous URL reference? I'd expect `self._zaak_url` here to still return the fully qualified URL. Then you would not have to make any changes in `def previous_zaak` and the `urljoin` machinery is abstracted away in the model field.

src/openzaak/components/besluiten/models.py
```diff
@@ -43,42 +45,73 @@ class Besluit(AuditTrailMixin, APIMixin, models.Model):
...
 
-    _zaak_url = models.URLField(
-        _("externe zaak"),
+    _zaak_base_url = models.ForeignKey(
+        Service,
+        null=True,
         blank=True,
-        max_length=1000,
+        on_delete=models.PROTECT,
+        related_name="besluit_zaken",
```
sergei-maertens, line 89:
```suggestion
        related_name="+",
```
or
```suggestion
        related_name="zaakbesluiten",
```

src/openzaak/components/besluiten/models.py
```diff
@@ -262,8 +313,12 @@ class Meta:
...
         constraints = [
             models.UniqueConstraint(
-                fields=["besluit", "_informatieobject_url"],
-                condition=~models.Q(_informatieobject_url=""),
+                fields=[
+                    "besluit",
+                    "_informatieobject_base_url",
+                    "_informatieobject_relative_url",
+                ],
+                condition=~models.Q(_informatieobject_relative_url__isnull=True),
```
sergei-maertens, line 321:
```suggestion
                condition=models.Q(_informatieobject_relative_url__isnull=False),
```

src/openzaak/components/besluiten/models.py
```diff
@@ -43,42 +45,73 @@ class Besluit(AuditTrailMixin, APIMixin, models.Model):
...
+        null=True,
+        help_text="Relatief deel van URL-referentie naar het extern BESLUITTYPE (in een andere Catalogi API).",
+    )
+    _besluittype_url = ServiceUrlField(
+        base_field="_besluittype_base_url",
+        relative_field="_besluittype_relative_url",
+        blank=True,
+        null=True,
+        help_text="URL-referentie naar het extern BESLUITTYPE (in een andere Catalogi API).",
     )
```
sergei-maertens, line 69: I can't tell easily from the migration files - does this field also set up a check constraint that when the FK is filled out, there must be a non-empty value for the `CharField` too?

src/openzaak/components/besluiten/tests/models/test_besluit_previous_zaak.py
```diff
@@ -22,8 +24,13 @@ def test_zaak_local(self):
 
         self.assertEqual(besluit.previous_zaak, zaak_before)
 
+    @tag("external-urls")
     @override_settings(ALLOWED_HOSTS=["testserver"])
     def test_zaak_external(self):
+        Service.objects.create(
+            api_root="https://external.documenten.nl/api/v1/", api_type=APITypes.zrc
+        )
```
sergei-maertens, line 32: is this still needed with the updated factories?

src/openzaak/components/besluiten/tests/test_auth.py
```diff
@@ -263,6 +267,9 @@ class ExternalBesluittypeScopeTests(JWTAuthMixin, APITestCase):
     component = ComponentTypes.brc
 
     def test_besluit_list(self):
+        Service.objects.create(
+            api_type=APITypes.ztc, api_root="https://externe.catalogus.nl/api/v1/"
```
sergei-maertens, line 271: you could use the constant here too - and maybe move this to `setUpTestData` instead?

src/openzaak/components/besluiten/tests/test_auth_cmis.py
```diff
@@ -30,6 +32,15 @@
...
     def test_cannot_read_without_correct_scope(self):
+        Service.objects.create(
+            api_root="http://testserver/documenten/api/v1/", api_type=APITypes.drc
+        )
+        Service.objects.create(
+            api_root="http://testserver/catalogi/api/v1/", api_type=APITypes.ztc
+        )
+        Service.objects.create(
+            api_root="http://testserver/besluiten/api/v1/", api_type=APITypes.brc
+        )
```
sergei-maertens, line 43: should we maybe make this part of `APICMISTestCase`? Just set up services for all API types and be done with it? :P

src/openzaak/components/besluiten/tests/test_besluit_with_external_zaak.py
```diff
@@ -32,18 +32,21 @@
 @override_settings(ALLOWED_HOSTS=["testserver"])
 class BesluitCreateExternalZaakTests(TypeCheckMixin, JWTAuthMixin, APITestCase):
     heeft_alle_autorisaties = True
-    base = "https://externe.catalogus.nl/api/v1/"
+    base = "https://externe.zaken.nl/api/v1/"
 
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
```
sergei-maertens, line 39: this should be `setUpTestData` rather than `setUpClass`

src/openzaak/components/besluiten/tests/test_besluit_with_external_zaak.py
```diff
@@ -169,7 +172,10 @@ def test_update_external_zaak(self):
 
         besluit.refresh_from_db()
 
-        self.assertEqual(besluit._previous_zaak_url, zaak_old)
+        self.assertEqual(besluit._previous_zaak_base_url_id, self.zaken_service.id)
+        self.assertEqual(
+            besluit._previous_zaak_relative_url, zaak_old[len(self.base) :]
+        )
```
sergei-maertens, line 178: the implementation detail doesn't really matter - I think the assert should remain the same, it was already leaky abstraction enough.

src/openzaak/components/besluiten/tests/test_besluitinformatieobjecten.py
```diff
@@ -210,14 +210,18 @@ class ExternalDocumentsAPITests(JWTAuthMixin, APITestCase):
     def setUpClass(cls):
         super().setUpClass()
```
sergei-maertens, line 211: here too, should be `setUpTestData`

src/openzaak/components/besluiten/tests/test_besluitinformatieobjecten.py
```diff
@@ -608,6 +644,12 @@ def setUpClass(cls):
             label="external documents",
             auth_type=AuthTypes.no_auth,
         )
+        Service.objects.create(
+            api_type=APITypes.ztc, api_root="http://testserver/catalogi/api/v1/"
+        )
+        Service.objects.create(
+            api_type=APITypes.ztc, api_root="http://openzaak.nl/catalogi/api/v1/"
+        )
```
sergei-maertens, line 652: seeing this repeated over and over makes me think we should perhaps have a factory with a trait instead, so we can do:

```py
ServiceFactory.create(is_ztc=True)
```
and then you can still pass the `api_root` override (but go with a default that's used most of the time)

src/openzaak/components/documenten/models.py
```diff
@@ -776,25 +816,37 @@ class Meta:
...
+                    | Q(
+                        _object_base_url__isnull=True,
+                        _zaak__isnull=True,
+                        _besluit__isnull=False,
+                    )
+                    | Q(
+                        _object_base_url__isnull=False,
+                        _zaak__isnull=True,
+                        _besluit__isnull=True,
+                    )
```
sergei-maertens, line 833: I'd also expect an additional check constraint that a relative part must be filled out when the service FK is not null (separate check, don't add it to this one)

src/openzaak/components/zaken/admin/zaken.py
```diff
@@ -346,18 +365,29 @@ def clean(self):
...
-    ordering = ("zaak", "_relevant_zaak", "_relevant_zaak_url")
-    raw_id_fields = ("zaak", "_relevant_zaak")
+    ordering = (
+        "zaak",
+        "_relevant_zaak",
+        "_relevant_zaak_base_url",
+        "_relevant_zaak_relative_url",
+    )
+    raw_id_fields = ("zaak", "_relevant_zaak", "_relevant_zaak_base_url")
     list_select_related = ("zaak", "_relevant_zaak")
```
sergei-maertens, line 391: please add the `_relevant_zaak_base_url` FK here too, otherwise performance will suffer

src/openzaak/components/zaken/admin/zaken.py
```diff
@@ -222,7 +233,8 @@ class ZaakInformatieObjectAdmin(AuditTrailAdminMixin, UUIDAdminMixin, admin.Mode
     list_display = (
         "zaak",
         "_informatieobject",
-        "_informatieobject_url",
+        "_informatieobject_base_url",
```
sergei-maertens, line 236: this needs to be added to `list_select_related` too

src/openzaak/components/zaken/admin/zaken.py
```diff
@@ -183,7 +183,13 @@ def clean(self):
 
 @admin.register(ZaakEigenschap)
 class ZaakEigenschapAdmin(AuditTrailAdminMixin, UUIDAdminMixin, admin.ModelAdmin):
-    list_display = ("zaak", "_eigenschap", "_eigenschap_url", "waarde")
+    list_display = (
+        "zaak",
+        "_eigenschap",
+        "_eigenschap_base_url",
```
sergei-maertens, line 189: please add this to `list_select_related`

src/openzaak/components/zaken/admin/zaken.py
```diff
@@ -522,7 +556,7 @@ class ZaakAdmin(
     search_fields = (
         "identificatie",
         "uuid",
-        "_zaaktype_url",
+        "_zaaktype_relative_url",
```
sergei-maertens, line 559: please add this to `list_select_related`

src/openzaak/utils/fields.py
```diff
@@ -33,19 +50,86 @@ def set_attributes_from_name(self, name):
...
+        )
 
 
 # register the override, as the upstream RelativeDeltaField registers its own admin
 # form field override as well.
 FORMFIELD_FOR_DBFIELD_DEFAULTS[DurationField] = {"form_class": RelativeDeltaFormField}
+
+
+class FkOrServiceUrlField(FkOrURLField):
+    """ supports zgw_comnsumers.ServiceUrlField as 'url_field' """
```
sergei-maertens, line 77:
```suggestion
    """
    Support :class:`zgw_consumers.ServiceUrlField` as 'url_field'
    """
```

src/openzaak/utils/fields.py
```diff
@@ -33,19 +50,86 @@ def set_attributes_from_name(self, name):
...
+        if self.model.__module__ == "__fake__":
+            return
+
+        url_base_field = self._url_field.base_field
+        # one of both MUST be filled and they cannot be filled both at the
+        # same time
+        empty_url_base_field = models.Q(**{f"{url_base_field}__isnull": True})
+        empty_fk_field = models.Q(**{f"{self.fk_field}__isnull": True})
+        fk_filled = ~empty_fk_field & empty_url_base_field
+        url_filled = empty_fk_field & ~empty_url_base_field
```
sergei-maertens, line 99: I think this needs the addition that the relative part is not empty too

src/openzaak/utils/migrations.py
```diff
@@ -0,0 +1,68 @@
...
+
+    :param fake_etag: workaround to migrate ETagMixin models
+    """
+    Service = apps.get_model("zgw_consumers", "Service")
+    cache_get_service = {}
+
+    for instance in model.objects.exclude(**{url_field: ""}):
+        url = getattr(instance, url_field)
+
+        if url in cache_get_service:
```
sergei-maertens, line 52: `@functools.lru_cache` on the `get_service` function should be a bit more straightforward and simplify this function.

src/openzaak/utils/migrations.py
```diff
@@ -0,0 +1,68 @@
...
+            service = cache_get_service[url]
+        else:
+            service = get_service(Service, url)
```
sergei-maertens, line 55: this does not handle the case/situation where the service doesn't exist (anymore) for the URL, and will then lead to migration crashes.

I'd do a detection loop first and create the missing services with a FIXME label based on a heuristic (you should be able to detected `/api/v1` in the URL and treat the prefix + that part as the root) to avoid this kind of breakage.

I'd also like to see some tests for these data migrations/utilities with those edge cases.

src/openzaak/utils/serializer_fields.py
```diff
@@ -48,3 +49,23 @@ class LengthHyperlinkedRelatedField(
...
+
+class FKOrServiceUrlValidator(FKOrURLValidator):
+    def __call__(self, url: str, serializer_field):
+        try:
+            super().__call__(url, serializer_field)
+
+        except ValueError:
+            raise serializers.ValidationError(
+                _("The service for this url is unknown"), code="unknown-service"
+            )
```
sergei-maertens, line 62:
```suggestion
            raise serializers.ValidationError(
                _("The service for this url is unknown"), code="unknown-service"
            ) from exc
```

src/openzaak/utils/serializer_fields.py
```diff
@@ -48,3 +49,23 @@ class LengthHyperlinkedRelatedField(
...
 ):
     pass
+
+
+class FKOrServiceUrlValidator(FKOrURLValidator):
+    def __call__(self, url: str, serializer_field):
+        try:
+            super().__call__(url, serializer_field)
+
+        except ValueError:
```
sergei-maertens, line 59:
```suggestion
        except ValueError as exc:
```

sergei-maertens (changes requested):
Do we have tests in the zaken admin (amongst others) that make it impossible for a Status to belong to a different Service than the zaak it belongs too?

E.g. Zaak is from zaaktype in Service 1, and Status belongs to zaak, but statustype belongs to Service 2 -> this should not be possible, as the statustype must be from the same service as the zaaktype.

src/openzaak/components/besluiten/tests/test_besluitinformatieobjecten.py
annashamray: I've followed your previous advice and added creation of necessary Services inside factories with FkorServiceUrl fields. 
Now it should work without setting services for FkorServiceUrl fields beforehand. 
I'll clean test files from `Service.objects.create` now

annashamray:
> The admin tests probably will still need the explicit decomposition because of how it works, but maybe you can consider a cusotm form field that does the decomposition instead? I think that's a better presentation and user-experience anyway, as it seems unlikely you'd keep the relative URL but change the service of an object. If you're migrating from a staging -> production environment, it would make more sense to update the `Service` instance itself.

@sergei-maertens 
I did some research:

In the `ModelAdmin` the `ModelForm` class is generated. In `ModelFormMetaclass` `django.forms.models.fields_for_model` function is used to generate form fields. 
This function loops only through concrete, private and M2M fields:
```
field_dict = {}
sortable_private_fields = [f for f in opts.private_fields if isinstance(f, ModelField)]
    for f in sorted(chain(opts.concrete_fields, sortable_private_fields, opts.many_to_many)):
        ...
        field_dict[f.name] = formfield
```

The problem is that `ServiceUrlField` is neither concrete nor private (just like `FkOrUrlField`), so it's just filtered out and ignored before calling `field.formfield()` method
I don't want to change anything this deep in ModelAdmin generation and I can't make `ServiceUrlField` private, because private fields are hidden and can't be accessed using `model._meta.fields` and `FkOrUrlField` won't see it. 

So for now I'd keep it as it is

src/openzaak/utils/migrations.py
annashamray: I will add the creation of missing services with FIXME label  and warning in logs but tbh I don't know how to test it. 
This function uses the URLField in the model which is already deleted
And I don't want to create some model just for testing, and leave it sitting in DB.

sergei-maertens:
> > The admin tests probably will still need the explicit decomposition because of how it works, but maybe you can consider a cusotm form field that does the decomposition instead? I think that's a better presentation and user-experience anyway, as it seems unlikely you'd keep the relative URL but change the service of an object. If you're migrating from a staging -> production environment, it would make more sense to update the `Service` instance itself.
> 
> @sergei-maertens I did some research:
> 
> In the `ModelAdmin` the `ModelForm` class is generated. In `ModelFormMetaclass` `django.forms.models.fields_for_model` function is used to generate form fields. This function loops only through concrete, private and M2M fields:
> 
> ```
> field_dict = {}
> sortable_private_fields = [f for f in opts.private_fields if isinstance(f, ModelField)]
>     for f in sorted(chain(opts.concrete_fields, sortable_private_fields, opts.many_to_many)):
>         ...
>         field_dict[f.name] = formfield
> ```
> 
> The problem is that `ServiceUrlField` is neither concrete nor private (just like `FkOrUrlField`), so it's just filtered out and ignored before calling `field.formfield()` method I don't want to change anything this deep in ModelAdmin generation and I can't make `ServiceUrlField` private, because private fields are hidden and can't be accessed using `model._meta.fields` and `FkOrUrlField` won't see it.
> 
> So for now I'd keep it as it is

okay for now, though I think you could use a custom form class and use that as base instead of relying completely on the `ModelForm` generation in the admin. But we'll leave it for now since I'm fed up of all the admin customizations were doing.

src/openzaak/utils/migrations.py
annashamray: Thank you, I didn't know about it. But I've changed the implementation of the `fill_service_urls` function and now I don't want to cache the empty values of `get_service` function, so I kept my implementation of caching

src/openzaak/components/besluiten/models.py
annashamray: the check constraint added to `zgw_consumers.ServiceUrlField` (PR - https://github.com/maykinmedia/zgw-consumers/pull/56)

src/openzaak/components/documenten/models.py
annashamray: the check constraint is added to zgw_consumers.ServiceUrlField (PR - https://github.com/maykinmedia/zgw-consumers/pull/56)

src/openzaak/utils/fields.py
annashamray: I've added the separate constraint for `ServiceUrlField` (PR - https://github.com/maykinmedia/zgw-consumers/pull/56)

src/openzaak/components/besluiten/models.py
annashamray: you are right, the change is excessive. I'll revoke this change

src/openzaak/components/besluiten/tests/test_besluit_with_external_zaak.py
annashamray: The assertion is changed back

src/openzaak/components/besluiten/admin.py
annashamray: annotations are added

requirements/base.txt
```diff
@@ -318,7 +323,7 @@ webencodings==0.5.1
     # via bleach
 wrapt==1.14.0
     # via deprecated
-zgw-consumers==0.19.0
+zgw-consumers==0.23.1
```
sergei-maertens, line 326: This is still causing non-unique constraint names :thinking:

src/openzaak/components/documenten/tests/test_cmis_url_mapping.py
```diff
@@ -37,6 +37,9 @@ def test_create_document_no_url_mapping(self):
         # Remove all available mappings
         UrlMapping.objects.all().delete()
 
+        Service.objects.create(
+            api_root="http://testserver/catalogi/api/v1/", api_type=APITypes.ztc
+        )
```
sergei-maertens, line 42: I don't see this test failing if this is not added, why is it needed?

The only situation I can think of is that the informatieobjecttype is treated as a non-local URL, but instead for that you should set `ALLOWED_HOSTS=["testserver"]` (if that's not the case) as that's the mechanism used by loose-fk to determine if a resource is local or remote.

src/openzaak/components/besluiten/tests/test_besluit_create_cmis.py
```diff
@@ -31,6 +33,7 @@ class BesluitCreateCMISTests(TypeCheckMixin, JWTAuthMixin, APICMISTestCase):
     @classmethod
     def setUpTestData(cls):
         super().setUpTestData()
+        Service.objects.create(api_root="http://testserver/", api_type=APITypes.orc)
```
sergei-maertens, line 36: I don't see any failing tests when this is not added :thinking:

src/openzaak/components/besluiten/tests/test_besluitinformatieobjecten_cmis.py
```diff
@@ -28,6 +30,7 @@ class BesluitInformatieObjectCMISAPITests(JWTAuthMixin, APICMISTestCase):
     def setUpTestData(cls):
         super().setUpTestData()
 
+        Service.objects.create(api_root="http://testserver/", api_type=APITypes.orc)
```
sergei-maertens, line 33: is this really necessary - same remark as other CMIS test cases :thinking:

src/openzaak/components/documenten/tests/admin/test_eio_cmis.py
```diff
@@ -23,6 +25,9 @@ class EnkelvoudigInformatieObjectCMISAdminTest(AdminTestMixin, APICMISTestCase):
     heeft_alle_autorisaties = True
 
     def test_create_eio_is_forbidden_when_cmis_enabled(self):
+        Service.objects.create(
+            api_root="http://testserver/catalogi/api/v1/", api_type=APITypes.ztc
+        )
```
sergei-maertens, line 30: is this really necessary - same remark as other CMIS test cases :thinking:

src/openzaak/components/documenten/tests/test_audittrails_cmis.py
```diff
@@ -35,6 +37,9 @@ class AuditTrailTests(JWTAuthMixin, APICMISTestCase):
     heeft_alle_autorisaties = True
 
     def _create_enkelvoudiginformatieobject(self, **HEADERS):
+        Service.objects.create(
+            api_root="http://testserver/catalogi/api/v1/", api_type=APITypes.ztc
+        )
```
sergei-maertens, line 42: is this really necessary - same remark as other CMIS test cases :thinking:

src/openzaak/components/documenten/tests/test_eio_file_cmis.py
```diff
@@ -96,6 +101,9 @@ def test_list_file(self):
         )
 
     def test_create_enkelvoudiginformatieobject_without_identificatie(self):
+        Service.objects.create(
+            api_root="http://testserver/catalogi/api/v1/", api_type=APITypes.ztc
+        )
```
sergei-maertens, line 106: is this really necessary - same remark as other CMIS test cases :thinking:

src/openzaak/components/documenten/tests/test_lock_eio_cmis.py
```diff
@@ -141,6 +143,9 @@ def test_update_fail_wrong_id(self):
         self.assertEqual(error["code"], "incorrect-lock-id")
 
     def test_create_ignores_lock(self):
+        Service.objects.create(
+            api_root="http://testserver/catalogi/api/v1/", api_type=APITypes.ztc
+        )
```
sergei-maertens, line 148: is this really necessary - same remark as other CMIS test cases :thinking:

src/openzaak/components/documenten/tests/test_upload_cmis.py
```diff
@@ -43,6 +45,9 @@ class SmallFileUpload(JWTAuthMixin, APICMISTestCase):
 
     @classmethod
     def setUpTestData(cls):
+        Service.objects.create(
+            api_root="http://testserver/catalogi/api/v1/", api_type=APITypes.ztc
+        )
```
sergei-maertens, line 50: is this really necessary - same remark as other CMIS test cases :thinking:

src/openzaak/components/documenten/tests/test_upload_cmis.py
```diff
@@ -531,6 +536,9 @@ class LargeFileAPITests(JWTAuthMixin, APICMISTestCase):
...
```
sergei-maertens, line 541: is this really necessary - same remark as other CMIS test cases :thinking:

src/openzaak/components/documenten/tests/test_notifications_send_cmis.py
```diff
@@ -85,6 +87,14 @@ class FailedNotificationTests(NotificationsConfigMixin, JWTAuthMixin, APICMISTes
     heeft_alle_autorisaties = True
     maxDiff = None
 
+    @classmethod
+    def setUpTestData(cls):
+        super().setUpTestData()
+
+        Service.objects.create(
+            api_root="http://testserver/catalogi/api/v1/", api_type=APITypes.ztc
+        )
```
sergei-maertens, line 96: is this really necessary - same remark as other CMIS test cases :thinking:

src/openzaak/tests/utils/factories.py
```diff
@@ -0,0 +1,50 @@
...
+
+            # create Service instance for composite field
+            base_url = furl(value).origin
+            services.append(Service(api_root=base_url, api_type=APITypes.orc))
+            # service, created = Service.objects.get_or_create(api_root=base_url, defaults={"api_type": APITypes.orc})
+
+        # if local host specified create service for it
+        local_host = kwargs.pop("local_host", cls.local_host)
+        if local_host:
+            services.append(Service(api_root=local_host, api_type=APITypes.orc))
```
sergei-maertens, line 46: I really don't understand why a local service needs to exist before you can use Open Zaak? Surely this cannot be needed at runtime, as that would require people (now) to configure:

* `django.contrib.sites.Site` with the domain
* `settings.OPENZAAK_DOMAIN` with the domain
* add a `Service` for the domain that Open Zaak is hosted on

That's three times we're configuring the same thing :/

src/openzaak/utils/migrations.py
```diff
@@ -0,0 +1,103 @@
...
+    for candidate in candidates.iterator():
+        if url.startswith(candidate.api_root):
+            return candidate
+
+    return None
+
+
+def calculate_api_root(url: str) -> str:
+    """
+    guess apI_root based on following logic:
```
sergei-maertens, line 38:
```suggestion
    guess api_root based on following logic:
```

src/openzaak/utils/migrations.py
```diff
@@ -0,0 +1,103 @@
...
+    * if url has /api/v{0-9}: api_root = prefix + that part
+    * if url has "/" - include prefix + that part
+    * otherwise include the whole string
+    """
+    patterns = [
+        r"(.+/api/v\d+/).+",
+        r"(\w+://.+\.\w+/).*",
```
sergei-maertens, line 45:
```suggestion
        r"(https?://.+\.\w+/).*",
```
any reason why we'd support anything that's not http(s)?

src/openzaak/utils/migrations.py
sergei-maertens: you can use this: https://github.com/open-formulieren/open-forms/blob/master/src/openforms/utils/tests/test_migrations.py#L8 :)

sergei-maertens (changes requested):

src/openzaak/tests/utils/factories.py
annashamray: Unfortunately `FkOrURLDescriptor` can't set string value without the related service. 
And afaik we use `FkOrURLDescriptor.__set__` with local url as string value in:
* `cmis_doc_to_django_model` function used for CMIS
* `virtual_model` initialization. For example, if we have ZIO, where informatieobject is external, but it has `informatieobjecttype`  which belongs to local catalogi. In this case we treat `informatieobjecttype` as external url

Adding `Service` data seemed the easiest (but of course not the prettiest) solution for me, otherwise it's the serios refactor of loose-fk descriptor, loader and/or virtual_model and I really didn't want to do it

src/openzaak/utils/migrations.py
annashamray: Tests are added, thanks for the link

PR #1252, opened 2022-08-30
:sparkles: [#1223] Implement caching for Documenten API

stevenbal:
Issue #1223

**Changes**

* Implements caching for Documenten (uses the cache in case of CMIS)


src/openzaak/components/documenten/tests/test_cmis_queries.py
```diff
@@ -54,6 +54,9 @@ def test_chained_filter(self):
         self.assertEqual(list(eios), [])
 
     def test_all(self):
```
stevenbal, line 56: Not sure why this is failing now :thinking: https://github.com/open-zaak/open-zaak/runs/8184171702?check_suite_focus=true#step:11:310

src/openzaak/components/documenten/caching.py
```diff
@@ -0,0 +1,88 @@
...
+      fields will be added to the automatically introspected serializer relations.
+    """
+
+    def decorator(viewset: type):
+        extract_dependencies(viewset, extra_depends_on or set())
+        condition = drf_condition(etag_func=partial(etag_func, etag_field=etag_field))
+        original_handler = getattr(viewset, action)
+
+        # Explicitly add the CMIS decorator, because it is overridden otherwise
+        handler = convert_cmis_adapter_exceptions(condition(original_handler))
```
stevenbal, line 81: I tried to see if I could figure out the decorators on the `original_handler` dynamically, but I couldn't get it to work

src/openzaak/components/documenten/tests/utils.py
```diff
@@ -18,6 +19,10 @@ def get_operation_url(operation, **kwargs):
     )
 
 
+def get_documenten_spec() -> dict:
+    return get_spec("src/openzaak/components/documenten/openapi.yaml")
```
sergei-maertens, line 23: this seems more like a generic test util for `openzaak.tests.utils`, taking a `component: str` argument.

Also, please use `pathlib.Path` to build filesystem paths instead of assuming unix.

sergei-maertens (approved):

PR #1254, opened 2022-09-01
Fix zaak zoek ordering parameter

### LINKED ISSUE #1198: Ordering parameter on ZaakZoek operation does not work
**General information**
Open Zaak version: 1.6.0

**Describe the bug**
the `ZaakZoek` operation in Zaken API allows for an `ordering` parameter in the request body (see https://test.openzaak.nl/zaken/api/v1/schema/#operation/zaak__zoek). This parameter does not work however.

I assume this bug also exists in the reference implementation.

**To Reproduce**
1. Create two Zaken with a different `startdatum`
2. Perform a `ZaakZoek` operation that results in these two zaken being shown and set `ordering` to `startdatum`
3. Perform the same operation again and set `ordering` to `-startdatum`
4. Observe that the ordering did not change

**Expected behavior**
I would expect the ordering to still work on `ZaakZoek`. I'm not sure whether `ordering` should be in the request body or the query parameters (like `page`)

sergei-maertens:
Fixes #1198

**Changes**

* Moved empty search body/parameters validation to view
* Added test for filter backend (search-capable)
* Added regression test for broken ordering
* Fixed looking up ordering parameter in request body

### DIFF
--- a/src/openzaak/components/zaken/api/serializers/zaken.py
+++ b/src/openzaak/components/zaken/api/serializers/zaken.py
@@ -395,14 +395,6 @@ class ZaakZoekSerializer(serializers.Serializer):
         help_text=_("Array of unieke resource identifiers (UUID4)"),
     )
 
-    def validate(self, attrs):
-        validated_attrs = super().validate(attrs)
-        if not validated_attrs:
-            raise serializers.ValidationError(
-                _("Search parameters must be specified"), code="empty_search_body"
-            )
-        return validated_attrs
-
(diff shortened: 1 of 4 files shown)


stevenbal (approved):

sergei-maertens:
Backported to release/1.7.3 in c92301215dc35828b43bec7fe6b66ce6d7b42082 and 44ee55851cd2f24897a7fae39a189bc0b96438a1

PR #1266, opened 2022-09-27
:wrench: [#621] Envvar support for NOTIFICATIONS_DISABLED

stevenbal:
Partially addresses #621

**Changes**

* Make NOTIFICATIONS_DISABLED configurable via environment variables

### DIFF
--- a/src/openzaak/conf/includes/base.py
+++ b/src/openzaak/conf/includes/base.py
@@ -592,6 +592,11 @@ ZGW_CONSUMERS_TEST_SCHEMA_DIRS = [
 ]
 
 #
+# VNG-API-COMMON
+#
+NOTIFICATIONS_DISABLED = config("NOTIFICATIONS_DISABLED", default=False)
+
+#
 # DJANGO-LOOSE-FK -- handle internal and external API resources
 #
 DEFAULT_LOOSE_FK_LOADER = "openzaak.loaders.AuthorizedRequestsLoader"
--- a/docs/installation/config/env_config.md
+++ b/docs/installation/config/env_config.md
@@ -149,6 +149,9 @@ on Docker, since `localhost` is contained within the container:
     Zaak to other services such as the [Selectie Lijst](https://selectielijst.openzaak.nl/) will be disabled, so this
     variable should be used with care to prevent unwanted side-effects.
 
+* `NOTIFICATIONS_DISABLED`: if this variable is set to `true`, `yes` or `1`, the notification mechanism will be
+    disabled. Defaults to `False`.
+
 ### Initial superuser creation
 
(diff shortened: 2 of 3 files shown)


src/openzaak/conf/includes/base.py
```diff
@@ -591,6 +591,11 @@
     os.path.join(DJANGO_PROJECT_DIR, "notifications", "tests", "files"),
 ]
 
+#
+# VNG-API-COMMON
+#
+NOTIFICATIONS_DISABLED = os.getenv("NOTIFICATIONS_DISABLED", False)
```
sergei-maertens, line 597: Please update the documentation: https://github.com/open-zaak/open-zaak/blob/main/docs/installation/config/env_config.md

sergei-maertens (changes requested):

src/openzaak/conf/includes/base.py
stevenbal: Added in https://github.com/open-zaak/open-zaak/pull/1266/commits/b07eb61588398f5786b676963e2442b9fc3ad184

I also changed the setting to use `config` and allow values such as `yes`, `True`, `1` in the .env

src/openzaak/conf/includes/base.py
```diff
@@ -591,6 +591,15 @@
     os.path.join(DJANGO_PROJECT_DIR, "notifications", "tests", "files"),
 ]
 
+#
+# VNG-API-COMMON
+#
+NOTIFICATIONS_DISABLED = config(
+    "NOTIFICATIONS_DISABLED",
+    default=False,
+    cast=lambda val: val in ["True", "true", "yes", "1"],
```
sergei-maertens, line 600: this is not needed, `config` already takes care of that because the `type` of `default` is `bool` :)

src/openzaak/conf/includes/base.py
```diff
@@ -591,6 +591,15 @@
     os.path.join(DJANGO_PROJECT_DIR, "notifications", "tests", "files"),
 ]
 
+#
+# VNG-API-COMMON
+#
+NOTIFICATIONS_DISABLED = config(
```
sergei-maertens, line 597: you just added the setting to the `base` module, but didn't remove it from the `conf/ci.py` or `conf/dev.py` files.

I'd update the `ci.py` file to include `os.environ.setdefault("NOTIFICATIONS_DISABLED", "yes")` at the top, and I think we can leave the `dev.py` module alone.

sergei-maertens (changes requested):

src/openzaak/conf/includes/base.py
stevenbal: Oh, neat :)

sergei-maertens (approved):

PR #1274, opened 2022-10-10
[#1204] Implement retry for notification sending

### LINKED ISSUE #1204: As admin, I want Open Zaak to retry sending notifications if they fail
... so I can guarantee delivery of notifications to the Notifications API

**Description**
Same implementation as the retry mechanism for Open Notificaties forwarding messages to clients (https://github.com/open-zaak/open-zaak/issues/1132)

- [x] Add celery
- [x] Refactor notification mechanism to use Celery
- [x] Apply same retry mechanism principles as in Open Notificaties:
  - https://github.com/open-zaak/open-notificaties/pull/59 
- [x] Update docs
  - https://github.com/open-zaak/open-notificaties/pull/60
- [x] Make admin configurable
    - https://github.com/open-zaak/open-notificaties/pull/70
- [ ] Fallback for if Celery is not (yet) available?
- [x] Logging only on final failure?

stevenbal:
Fixes #1204

Requires https://github.com/maykinmedia/notifications-api-common/pull/4 to be merged and released first

**Changes**
* Use celery to send notifications (making it non-blocking for API requests)
* Implement the same retry mechanism as in Open Notifications
* Only save FailedNotifications after the last try is done
* Retry mechanism settings are admin configurable

TODO:
- [x] Implement retry
- [x] Tests for retry mechanism
- [x] test statuslog after all failed?
- [x] Remove settings from env_config.md
- [x] Use PyPI version of notifications-api-common
- [x] Update docs
- [x] Update `Dockerfile` to include the celery binary + celery worker script
- [x] Use redis as celery backend
- [x] Add  flower for monitoring


docker-compose.yml
```diff
@@ -17,13 +17,19 @@ services:
     ports:
       - "11234:5432"
 
+  rabbitmq:
```
stevenbal, line 20: I figured I'd stick to RabbitMQ since Open Notificaties uses it as well, but I read that Redis can be used as a broker as well, not sure which we would prefer in this case

docs/installation/config/env_config.md
```diff
@@ -180,6 +180,22 @@ non-exhaustive list of options you may need with Open Zaak.
   within this timeout, then uWSGI will error out. This has been observed with certain
   CMIS implementations causing slow requests where 60s is not sufficient.
 
+### Celery
+
+* `CELERY_BROKER_URL`: the URL of the broker that will be used to actually send the notifications (default: `amqp://127.0.0.1:5672//`).
+
+* `CELERY_RESULT_BACKEND`: the backend where the results of tasks will be stored (default: `redis://localhost:6379/1`)
+
+* `NOTIFICATION_DELIVERY_MAX_RETRIES`: the maximum number of retries Celery will do if sending a notification failed.
```
stevenbal, line 189: Should be removed

docker-compose.yml
sergei-maertens: redis - less moving parts because we already have it in the stack and it has less peculiarities that RabbitMQ seems to have. There are also existing recipes for Highly-Available redis clusters (left as an exercise to the devops engineer)

sergei-maertens:
@annashamray notifications-api-common 0.2.0 should be on PyPI

src/openzaak/notifications/models.py
```diff
@@ -18,6 +18,8 @@ class FailedNotification(StatusLog):
     status_code = models.IntegerField(
         _("status_code"),
         help_text=_("Status code received from the Notifications API."),
+        null=True,
+        blank=True,
```
annashamray, line 22: We don't log HTTP status anymore (I don't think we did log it before tbh), and admin resend is deprecated, so I just marked `status` as optional before removing this whole model in the next OZ version

sergei-maertens (approved):

PR #1284, opened 2022-11-07
Release 1.7.5

sergei-maertens:
- [ ] Update changelog
- [ ] Bump version number
- [ ] ...


sergei-maertens:
Support colleague did their best to break the zaaktype admin and found no release blockers, so let's cross our fingers that this can be the final 1.7.x patch release :crossed_fingers:

PR #1286, opened 2022-11-10
add middleware to rewrite the request host

### LINKED ISSUE #1020: Implement OPENZAAK_DOMAIN setting/environment variable
Typically, Open Zaak builds URLs based on the request `Host` (or `X-Forwarded-Host`) header. However, this is problematic if you're only exposing Open Zaak over NLX and configuring the NLX service with the internal Kubernetes (or Docker) hostname which looks like `servicename.namespace.svc.cluster.local`. These hosts cannot be communicated to the outside name since they aren't reachable as such.

For this purpose, we should be able to specify the canonical host name through an environment variable `OPENZAAK_DOMAIN`, so this value can be used to build fully qualified URLs

sergei-maertens:
Fixes #1020

**Changes**

* Rewrite HTTP_HOST at the request level.
* Deprecate `Site.objects.get_current().domain` usage in favour of `OPENZAAK_DOMAIN`
* Use `OPENZAAK_DOMAIN` in `build_absolute_url` if available
* Added system checks to detect bad/ambiguous configurations

### DIFF
--- a/src/openzaak/components/autorisaties/tests/admin/test_autorisaties_admin.py
+++ b/src/openzaak/components/autorisaties/tests/admin/test_autorisaties_admin.py
@@ -9,7 +9,7 @@ from urllib.parse import urlparse
 
 from django.contrib.auth.models import Permission
 from django.contrib.sites.models import Site
-from django.test import TransactionTestCase, override_settings, tag
+from django.test import TestCase, override_settings, tag
 from django.urls import reverse
 from django.utils.translation import ugettext_lazy as _
 
@@ -123,27 +123,30 @@ class ApplicatieInlinesAdminTests(WebTest):
 
 
 @tag("admin-autorisaties")
-class ManageAutorisatiesAdmin(NotificationsConfigMixin, TransactionTestCase):
-    def setUp(self):
-        super().setUp()
+class ManageAutorisatiesAdmin(NotificationsConfigMixin, TestCase):
+    @classmethod
+    def setUpTestData(cls):
+        super().setUpTestData()
 
         site = Site.objects.get_current()
         site.domain = "testserver"
         site.save()
 
-        user = UserFactory.create(is_staff=True)
+        cls.user = user = UserFactory.create(is_staff=True)
         perm = Permission.objects.get_by_natural_key(
             "change_applicatie", "authorizations", "applicatie"
         )
         user.user_permissions.add(perm)
-        self.applicatie = ApplicatieFactory.create()
-        self.client.force_login(user)
+        cls.applicatie = ApplicatieFactory.create()
+
+    def setUp(self):
+        super().setUp()
+
+        self.client.force_login(self.user)
         self.url = reverse(
             "admin:authorizations_applicatie_autorisaties",
             kwargs={"object_id": self.applicatie.pk},
         )
-
-        self._configure_notifications()
 
     def test_page_returns_on_get(self):
         # set up some initial data
@@ -224,13 +227,15 @@ class ManageAutorisatiesAdmin(NotificationsConfigMixin, TransactionTestCase):
             "form-0-vertrouwelijkheidaanduiding": VertrouwelijkheidsAanduiding.beperkt_openbaar,
         }
 
-        response = self.client.post(self.url, data)
+        with self.captureOnCommitCallbacks(execute=True):
+            response = self.client.post(self.url, data)
 
         self.assertEqual(response.status_code, 302)
         self.assertFalse(Autorisatie.objects.exists())
 
         # create a ZaakType - this should trigger a new autorisatie being installed
-        ZaakTypeFactory.create()
+        with self.captureOnCommitCallbacks(execute=True):
+            ZaakTypeFactory.create()
         self.assertEqual(self.applicatie.autorisaties.count(), 1)
 
     def test_noop_all_current_and_future_zaaktypen(self):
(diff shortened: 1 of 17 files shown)


annashamray:
1. `OPENZAAK_DOMAIN` is not used and `Site.objects.get_current()` is still used for:
* sending notification to applicatie (`autorisaties.send_applicatie_changed_notification function`)
* `export` management command
* `setup_configuration`management command
* `gerelateerd_zaaktype` admin field

Do you plan to change them in the future PRs? 

2. Could you add documentation about `OPENZAAK_DOMAIN` in : 
 * env_config.md
 * docs/installation/config/openzaak_config.rst section "Setting the domain" (or at least add note that the Site configuration is to be depreciated)

alextreme:
Full confidence that Anna can review this better than I can :sweat_smile:

src/openzaak/components/autorisaties/tests/admin/test_autorisaties_admin.py
```diff
@@ -122,28 +124,31 @@ def _add_autorisatie(self, obj, **kwargs):
 
 
 @tag("admin-autorisaties")
-class ManageAutorisatiesAdmin(NotificationsConfigMixin, TransactionTestCase):
-    def setUp(self):
-        super().setUp()
+class ManageAutorisatiesAdmin(NotificationsConfigMixin, TestCase):
```
sergei-maertens, line 127: I'm sorry for this refactor, but this test case was horribly slow making TDD a pain :(

src/openzaak/conf/includes/base.py
```diff
@@ -442,6 +443,11 @@
 PROJECT_NAME = "Open Zaak"
 SITE_TITLE = "API dashboard"
 
+# if specified, this replaces the host information in the requests, and it is used
+# to build absolute URLs.
+OPENZAAK_DOMAIN = config("OPENZAAK_DOMAIN", "")
+OPENZAAK_REWRITE_HOST = config("OPENZAAK_REWRITE_HOST", False)
```
sergei-maertens, line 449: I've split this into two settings, one affects the middleware, the other is always used if it's available to build requests in non-HTTP contexts (such as exports, notifications...) to eventually replace `Site.objects.get_current()` usage.

src/openzaak/utils/__init__.py
```diff
@@ -15,12 +17,43 @@ def parse_isodatetime(val) -> datetime:
...
-    return f"{protocol}://{domain}{path}"
+    domain = get_openzaak_domain()
+    _furl = furl(
+        scheme="https" if settings.IS_HTTPS else "http", netloc=domain, path=path,
+    )
+    return _furl.url
+
+
+def build_fake_request(*, method="get", **kwargs) -> HttpRequest:
+    from rest_framework.test import APIRequestFactory
```
sergei-maertens, line 49: not doing this as a local import breaks a number of things because of drf.settings.api_settings usage :thinking:

src/openzaak/components/catalogi/admin/forms.py
```diff
@@ -667,3 +666,15 @@ class ZaakTypenRelatieAdminForm(forms.ModelForm):
...
+        extra = ["gerelateerd_zaaktype"]
+        return super()._get_validation_exclusions() + extra
+
+    def normalize_data(self, request, obj):
+        bits = urlsplit(obj.gerelateerd_zaaktype)
+        if bits.scheme and bits.netloc:
+            return
+        obj.gerelateerd_zaaktype = build_absolute_url(
+            obj.gerelateerd_zaaktype, request=request
+        )
```
sergei-maertens, line 680: :smiling_face_with_tear:

src/openzaak/components/catalogi/management/commands/export.py
```diff
@@ -59,9 +60,7 @@ def handle(self, *args, **options):
                 _("The number of resources supplied does not match the number of IDs")
             )
 
-        factory = APIRequestFactory()
-        server_name = Site.objects.get_current().domain
-        request = factory.get("/", SERVER_NAME=server_name)
+        request = Request(build_fake_request())
```
annashamray, line 63: Sorry I don't understand why you wrap `build_fake_request()` in `Request` class. 
I thought it should already be a request and `Request` wasn't used in the old code

src/openzaak/tests/test_rewrite_host.py
```diff
@@ -0,0 +1,167 @@
...
+
+        self.assertEqual(response.content, b"openzaak.example.com:8443")
+
+    @override_settings(OPENZAAK_DOMAIN="")
+    def test_setting_empty(self):
+        response = self.client.get("")
+
+        self.assertEqual(response.content, b"testserver")
+
+    @override_settings(OPENZAAK_REWRITE_HOST=False)
```
annashamray, line 48:
```suggestion
    @override_settings(OPENZAAK_DOMAIN="openzaak.example.com", OPENZAAK_REWRITE_HOST=False)
```

src/openzaak/utils/__init__.py
```diff
@@ -15,12 +17,43 @@ def parse_isodatetime(val) -> datetime:
...
+def get_openzaak_domain() -> str:
+    """
+    Obtain the domain/netloc of Open Zaak according to settings or configuration.
+    """
     from django.contrib.sites.models import Site
 
+    if settings.OPENZAAK_DOMAIN:
+        return settings.OPENZAAK_DOMAIN
+
+    warnings.warn(
```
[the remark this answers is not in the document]
sergei-maertens: Hrm no, not at this time - this is now basically a centralized location to obtain the Open Zaak domain and it warns when it's going to fallback behaviour. When we've refactored all the usage of contrib.sites as a first option in favour of the new setting, we can make this a `DeprecationWarning` instead of `PendingDeprecationWarning` and then in the next release we can remove it.

src/openzaak/components/catalogi/management/commands/export.py
sergei-maertens: (discussed in the office too)

it's because in this context DRF serializers are used, where typically the serializer context uses a DRF `Request`, which wraps around the Django request (see https://github.com/encode/django-rest-framework/blob/master/rest_framework/views.py#L385) and this serves to mimic that behaviour. `build_fake_request` returns a Django `HttpRequest` rather than a DRF `Request` as using the DRF Request outside of DRF context doesn't make much sense.

PR #1288, opened 2022-11-16
Don't measure the coverage of tests themselves

sergei-maertens:
We get some drops in coverage now because some try: except: calls in tests aren't hit (because they are now passing correctly).

Measuring the coverage of the test code by itself already skews the coverage statistics, we only care about the actual application code.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -21,3 +21,5 @@ exclude = migrations,static,media
 [coverage:run]
 branch = True
 source = src
+omit =
+    */test_*.py


sergei-maertens:
brilliant, this doesn't trigger the test suite :)

I'll have to check locally how this results in changes in coverage...

sergei-maertens:
Checked it locally, (without CMIS), the coverage goes from 88% to 87%. Interestingly, the number of lines measured goes from 38.4K to 13.6K, which is a massive drop in bloat.

Issue #1271, opened 2022-09-29
Error when 2 zaken are created at exactly the same moment


petervanmanen:
**General information**
Open Zaak version: 1.7.3-maykinlatest (with 3 openzaak nodes)

**Describe the bug**
When loadtesting we ran into an issue where if two threads create a new zaak (without specifying an identificatie) at exactly the same moment an error occurs:  HTTPError('500 Server Error: Internal Server Error for url: https://openzaak-zgw.test.denhaag.nl/zaken/api/v1/zaken')

It appears that two processes can be awarded the same identificatie and the first one to use it is succesfull, the second one gets an error the the identificatie is not unique. However openzaak itself generates the identificatie.

**To Reproduce**
1. Have script that creates a lot of zaken in multiple threads
2. Wait until the collision occurs

**Expected behavior**
Openzaak is threadsafe and multiple threads can create zaken at the same time.

joeribekker:
Judging by the screenshot, you provided the zaak "identificatie" yourself. Open Zaak uses UUID4 to create unique zaaknumbers (with near-zero chance of colissions). Are you sure this collision was caused by Open Zaak (and some custom number generator)?

petervanmanen:
@joeribekker this is the post request I use
{"bronorganisatie": "111222333","omschrijving": "BLAABLAA","toelichting": "BLAABLAA","zaaktype": "https://openzaak-zgw.blabla.nl/catalogi/api/v1/zaaktypen/b61bce86-f2ac-4aff-a155-d4d51884395d","verantwoordelijkeOrganisatie": "111222333","startdatum": "2022-09-29"}

So no identificatie provided

joeribekker:
Hmm, you're right. My memory is incorrect.

Never had this before, even when importing/migrating large amounts of cases. Maybe those were always done in serial.

sergei-maertens:
because of transaction isolation, this is indeed a risk - the only way we can prevent this is probably by taking a lock on the entire database table, and I think this will bring bigger (performance) problems with it.

I think it'd make more sense to translate this particular unique constraint violation as a HTTP 409 Conflict, so the client can retry.

Another way would be to use a different number generation algorithm (non-sequential/randomized) with lower chances of collisions, or automatically retry a number of times if we run into conflict.

petervanmanen:
@sergei-maertens today I ran a test, it ran for 20 minutes. In these 20 minutes it tried creating 280 zaken, it failed 86 times. 
[...]

PR #1299, opened 2022-12-09
Improving database/query performance of zaken create endpoint

sergei-maertens:
Relates to #1271

This PR addresses the excessive queries done by Open Zaak when a zaak is being created via a POST call. Improving performance of this endpoint reduces the total transaction time and thereby lowers the probability of race conditions. This does not 100% solve the concurrency issue, but reducing the probability gives more flexibility for pragmatic approaches to fix it without ending up in the non-happy flow _too frequently_.

And everyone benefits from better performance on this endpoint.


src/openzaak/components/zaken/api/serializers/zaken.py
```diff
@@ -340,17 +341,19 @@ class Meta:
             HoofdZaaktypeRelationValidator(),
         ]
 
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
+    def get_fields(self):
+        fields = super().get_fields()
```
sergei-maertens, line 345: refactored because `self.fields` calls `self.get_fields()` and then caches the result (through `@cached_property`). Forcing all fields to be bound in the `__init__` can break certain behaviours and even affect API spec generation.

annashamray: If changing field properties in `Serializer.__init__` is a bad practice should we also change following classes?  

* `BesluitSerializer` 
* `EigenschapSpecificatieSerializer` 
* `InformatieObjectTypeSerializer` 
* `ZaakTypeInformatieObjectTypeSerializer` 
* `BrondatumArchiefprocedureSerializer` 
* `ResultaatTypeSerializer`
* `RolTypeSerializer`
* `ZaakTypenRelatieSerializer`
* `ZaakTypeSerializer` 
* `OndertekeningSerializer`
* `IntegriteitSerializer`
* `EnkelvoudigInformatieObjectSerializer`
* `ObjectInformatieObjectSerializer`
* `RelevanteZaakSerializer`
* `RolSerializer`
* `ZaakObjectSerializer`
* `RolNatuurlijkPersoonSerializer`
* `ObjectInrichtingselementSerializer`
* `ObjectKunstwerkdeelSerializer`
* `ObjectSpoorbaandeelSerializer`
* `ObjectWaterdeelSerializer`

src/openzaak/components/zaken/api/serializers/zaken.py
```diff
@@ -410,7 +413,29 @@ def create(self, validated_data: dict):
                 }
             )
 
-        return super().create(validated_data)
+        obj = super().create(validated_data)
+        track_object_serializer(obj, self)
```
annashamray, line 417: Is it correct that the dict inside this function is the reason why you had to rewrite `__hash__` function?

src/openzaak/components/zaken/tests/test_create.py
```diff
@@ -458,3 +462,89 @@ def create_zaak():
...
+                skipped, it's create of root resource!) vng_api_common.caching.signals
+            32: select zaak relevantezaakrelatie (nested inline create, can't avoid this)
+            33: select zaak kenmerken (nested inline create, can't avoid this)
+            34: insert audit trail
+         35-36: notifications, select created zaak (?), notifs config
+            37: release savepoint (from NotificationsCreateMixin)
+            38: savepoint create transaction.on_commit ETag handler (start new transaction)
+            39: update ETag column of zaak
+            40: release savepoint (commit transaction)
+        """
```
annashamray, line 528: :heart_eyes:

src/openzaak/utils/serializer_fields.py
```diff
@@ -52,15 +52,41 @@ class LengthHyperlinkedRelatedField(
...
+        resolved_instance = resolver.resolve(host, url)
+        serializer_field.context[context_key] = resolved_instance
+
+    @staticmethod
+    def get_context_cache_key(field):
+        field_names = [field.field_name]
+        while (field := field.parent) and field.field_name:
+            field_names.append(field.field_name)
+        names = "__".join(field_names)
+        return f"_resolved_{names}"
```
annashamray, line 88: Can't we use `field.source_attrs` for the name?

src/openzaak/components/zaken/api/serializers/zaken.py
sergei-maertens: correct - django's base `__hash__` requires `self.pk` to not be `None`. In the CMIS case we can rely on the `uuid` being unique and act like a PK for that purpose, even though there's no DB record in postgres.

src/openzaak/utils/serializer_fields.py
sergei-maertens: I think I tried that but can't remember why it broke :thinking: possibly with some dynamic serializers there was no guarantee that `source_attrs` would've been set already as that only happens during the binding, but I could be wrong. I'll try it next week!

src/openzaak/components/zaken/api/serializers/zaken.py
annashamray: Could you add `TODO` comment here about it? So we won't forget which method should be replaced?

src/openzaak/components/zaken/api/serializers/zaken.py
annashamray: Thanks

src/openzaak/components/zaken/api/serializers/zaken.py
sergei-maertens: I've created #1307 for it and assigned it to the next milestone so it's on the board now :)

src/openzaak/utils/serializer_fields.py
sergei-maertens: Tried it and found the cause - you can of course have nested serializers, and those don't have the `source_attrs` attribute, resulting in:

```
AttributeError: 'ZaakSerializer' object has no attribute 'source_attrs'
```

PR #1311, opened 2023-01-17
document notification performance test

### LINKED ISSUE #1309: Publish the analysis of the reliability for sending notifications from Open Zaak
Add some statistics about the connection performance between Open Zaak en Open Notificaties to readthedocs.

**(Upd. by Anna)** There should be two types of analysis:

* how much time is required to send notifications after burst load of created zaken (5 sec burst load, 15 sec burst load). Possible bottlenecks: celery workers (can be increased) and open notificaties api itself (the amount of replicas can also be increased)
* autoretry test: what happens to notifications if ON is not available for several seconds: how many notifications go to autoretry, how many are finally failed

annashamray:
Fixes #1309

### DIFF
--- a/docs/development/performance/notifications.rst
+++ b/docs/development/performance/notifications.rst
@@ -3,18 +3,22 @@
 Notifications performance
 =========================
 
-Open Zaak sends notifications using `Celery`_ as a task queue.
-Since sending notifications is asynchronous we use auto retry in case the tasks have failed.
+Open Zaak sends notifications using `Celery`_ as an asynchronous task queue. Notification delivery
+failure to the Notifications API/service may happen, which is why we leverage "autoretry" for
+these failed tasks.
 
-Below we analyze the reliability and performance of sending notifications about
-the creation of the zaak.
+The reliability and performance of sending notifications was analyzed and the results are reported
+below. In this test case, we observed the behaviour of notifications triggered by "case create" 
+events.
 
 
 Sending notifications under burst load
 --------------------------------------
 
-We measure the notifications median time and failure rate under the increasing
-amount of concurrent users.
+We measure the notifications median time and failure rate under an increasing
+amount of concurrent users. The goal is to observe the reliability of notification
+delivery and delay between notification scheduling and it actually arriving at the
+Notifications API.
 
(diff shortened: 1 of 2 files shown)


docs/development/performance/notifications.rst
```diff
@@ -0,0 +1,66 @@
...
+   :header-rows: 1
+
+    Notification downtime (s),Request Count,Failure Count,Notifications failed,Notifications retried,Notifications processed
+    10,1486,0,0,919,2114
+    60,1478,0,0,3729,5195
+    300,1509,0,1276,6380,7908
+
+According to the result if the notification server is down for less than 1 min, then all notifications
+would be eventually send. If it's expected that the notification service is down for longer time
+the autoretry configuration should be adjusted in the admin.
```
sergei-maertens, line 66: I'm having a bit of a hard time interpreting these numbers (and labels)

* What does "notifications processed" mean? is this on the OZ side or the Notifications API side?
* The retry count relates to the request count in some way, maybe we should be able to describe what the lower and upper bound are? though we have a sliding window, a request that finished on second 59 can only be retried once with the 1s wait time before retry

"Configuration should be adjust in the admin..." -> I think we should clarify (and perhaps do a second test run to confirm) that you can do this be either increasing the delay factor or the maximum number of delays or both of those. It would be nice to get a working example of a configuration that can cover 5 minutes of down time:

E.g. without touching the factor/upper limit to delay duration:

1. At 0s request arrives and notification is scheduled
2. At 1s the first retry happens
3. At 2s the second
4. At 4s the third
5. At 8s the fourth
6. At 16s the fifth
7. At 32s the sixth
8. At 64s (capped to 48s) the sevent
9. At 96s the eight
10. At 144s the ninth
11. At 192s the tenth
12. At 240s the eleventh
13. At 288s the twelfth
14. At 332 the thirteenth

So that would imply a maximum retry count of 13 to bridge 5 minutes of downtime.

sergei-maertens (changes requested):

annashamray:
The build is failing because ECMA website (https://ecma-international.org/) is down

sergei-maertens (approved):

PR #1312, opened 2023-01-18
Fix EXTRA_CERTS_ENVVAR crash due to multiple imports

sergei-maertens:
Celery initialization appears to trigger setup.py execution multiple times, so it needs to be idempotent.

### DIFF
--- a/src/openzaak/setup.py
+++ b/src/openzaak/setup.py
@@ -34,6 +34,25 @@ def setup_env():
 
 
 def load_self_signed_certs() -> None:
+    """
+    Load self-signed/private CA bundles in an idempotent manner.
+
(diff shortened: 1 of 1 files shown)


src/openzaak/setup.py
```diff
@@ -34,6 +34,14 @@ def setup_env():
 
 
 def load_self_signed_certs() -> None:
+    needs_extra_verify_certs = os.environ.get(EXTRA_CERTS_ENVVAR)
```
annashamray, line 37: Could you please add the docstring about this change? That you are making `_load_self_signed_certs` idempotent and env var `REQUESTS_CA_BUNDLE` is set inside this function?

sergei-maertens: done!

annashamray (approved):

PR #1329, opened 2023-02-07
:memo: [#1327] document configuration of local services for the usage…

### LINKED ISSUE #1327: As administrator, I want to have a complete documentation how to setup OZ with external APIs
**Description**
For now Open Zaak local APIs should be added as `Service` model instances in order external components work in the correct way
It should be documented

Issue to refactor this part of code: https://github.com/open-zaak/open-zaak/issues/1326

annashamray:
fixes #1327


docs/installation/config/openzaak_config.rst
```diff
@@ -301,3 +301,70 @@ endpoint is accessible at ``{{base_url}}/catalogi/api/v1/zaaktypen`` (where
     A GET request to the Catalogi API using Postman
 
 .. _Postman: https://www.getpostman.com/
+
+
+Using external ZGW APIs
+=======================
```
sergei-maertens, line 307: I think this should be its own document - probably under the "Guides" TOC (https://github.com/open-zaak/open-zaak/blob/main/docs/installation/index.rst)

Additionally, this documentation page starts by pointing out the CLI-based configuration, as point-and-click configuration is not the best solution. That makes it that this part of the docs is not obvious to get by, or results in a very scattered navigation between documentation pages.

docs/installation/config/openzaak_config.rst
```diff
@@ -301,3 +301,70 @@ endpoint is accessible at ``{{base_url}}/catalogi/api/v1/zaaktypen`` (where
...
+
+Open Zaak provides production-grade :ref:`API's<api_index>` to store and retrieve data:
+
+* Zaken API (case instances)
+* Documenten API (documents)
+* Catalogi API (case types)
+* Besluiten API (decisions)
```
sergei-maertens, line 314: Autorisaties API is not listed, nor do I think it's relevant to mention what OZ itself offers.

I'd rather see something along the lines of:

```rst
While Open Zaak itself implements all the necessary APIs for Zaakgericht Werken, it is possible
(conform the VNG standard) to use these APIs from _other components_ than Open Zaak that
also implement the VNG standard(s).

For example, you might use an alternative implementation of the Documenten API. Using this
example, this guide describes the necessary configuration steps in Open Zaak.
```

sergei-maertens (changes requested):

docs/installation/config/openzaak_config.rst
annashamray: The introduction is changed. Thank you for the provided text

PR #1362, opened 2023-04-17
optimize list zaken 

### LINKED ISSUE #1361: Performance optimization: GET zaken
Master US  - #1344

The performance testing showed that GET zaken takes ~3 sec aven with superuser permission

Related APM transaction id = 65c04961e48cd9a1
APM shows that most of the time (2,3 sec) takes the query to count zaken: 

```
FROM (
   SELECT DISTINCT "zaken_zaakidentificatie"."id" AS Col1,
        "zaken_zaakidentificatie"."identificatie" AS Col2,
        "zaken_zaakidentificatie"."bronorganisatie" AS Col3,
        "zaken_zaak"."_etag" AS Col4,
        "zaken_zaak"."identificatie_ptr_id" AS Col5,
        "zaken_zaak"."uuid" AS Col6,
        "zaken_zaak"."id" AS Col7,
         ...
   FROM "zaken_zaak"
       inner JOIN "zaken_zaakidentificatie" ON ("zaken_zaak"."identificatie_ptr_id" = "zaken_zaakidentificatie"."id")
  ) subquery;

```

The join happens because of the concrete table inheritance of `zaken.Zaak` model. 

So the solution is to use custom qs in pagination, for example to use Zaken.objects.values("identificatie_ptr_id").count()

annashamray:
Fixes #1361

### DIFF
--- a/src/openzaak/components/zaken/api/pagination.py
+++ b/src/openzaak/components/zaken/api/pagination.py
@@ -1,0 +1,17 @@
+# SPDX-License-Identifier: EUPL-1.2
+# Copyright (C) 2023 Dimpact
+from django.core.paginator import Paginator as DjangoPaginator
+
+from rest_framework.pagination import PageNumberPagination
+
+
+class ZaakPaginator(DjangoPaginator):
+    def count(self):
+        """
+        ⚡ restricts values to PK to remove implicit join from SQL query
+        """
+        return self.object_list.values("pk").count()
+
+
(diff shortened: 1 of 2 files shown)


annashamray:
the pagination is already tested here - https://github.com/open-zaak/open-zaak/blob/main/src/openzaak/components/zaken/tests/test_zaken.py#L448 

I will test the impact on performance separately

sergei-maertens (approved):

PR #1364, opened 2023-04-17
:zap: [#1363] Private _voltooid property on BestandsDeel

### LINKED ISSUE #1363: Performance optimization: EnkelvoudigInformatieObject list endpoint
Master US  - #1344
Taiga issue 470

Performance testing showed that the `EnkelvoudigInformatieObject` list endpoint is slow, especially if there are lots of `BestandsDelen`.

Notes:
- It seems that the `voltooid` attribute on `BestandsDeelSerializer` adds fairly significant overhead (similar for `url`)
- The pagination `count` method seems to causes an expensive query to be executed -> mostly caused by the `.distinct`?

stevenbal:
Fixes #1363 

**Changes**
* Add prefetch_related for bestandsdelen to EIO viewset query
* Ignore unnecessary joins in pagination `.count` method
* Add private `_voltooid` attribute to `BestandsDeel` to prevent unnecessary checks using `.inhoud`


stevenbal:
Should be deployed and tested on the performance test server before merging

src/openzaak/components/documenten/migrations/0027_auto_20230417_1415.py
```diff
@@ -0,0 +1,26 @@
...
+# Generated by Django 3.2.18 on 2023-04-17 14:15
+
+from django.db import migrations
+
+
+def populate_private_voltooid_attribute(apps, schema_editor):
+    BestandsDeel = apps.get_model("documenten", "BestandsDeel")
+
+    for bestandsdeel in BestandsDeel.objects.all():
+        bestandsdeel._voltooid = bestandsdeel.inhoud.size == bestandsdeel.omvang
```
joeribekker, line 12: Will this cause a call over CMIS for every document in a migration (when CMIS is used)?

In that case, I much rather evaluatie and save this when its consulted.

stevenbal: I don't think so, `BestandsDeel` is just a regular Django model that does not use CMIS (from what I can see)

joeribekker:
Refinement: Steven will check if the CMIS-code is hit in the migration. @stevenbal @stevenbal

annashamray:
@joeribekker I've checked it locally and haven't seen a CMIS request during the migration. But to make sure I'll try to create a unit test for it

annashamray:
The test to check if data migration triggers CMIS request is added

annashamray (approved):

joeribekker (approved):

PR #1368, opened 2023-04-18
[DRAFT] :zap: [#1365] Performance improvements for permissions check

### LINKED ISSUE #1365: Performance optimization: list endpoints slow due to permission checks
Master US - https://github.com/open-zaak/open-zaak/issues/1344
Taiga issue 454

Performance testing showed that several list endpoints are slowed down if a JWT without superuser permissions:
- GET openzaak.components.zaken.api.viewsets.RolViewSet  
- GET openzaak.components.zaken.api.viewsets.ZaakInformatieObjectViewSet  
- GET openzaak.components.zaken.api.viewsets.StatusViewSet

Notes:
- determining the `count` for pagination seems to be the bottleneck here as well, in this case because the query contains an `INNER JOIN` on the `zaken_zaak` table, which seems to reduce performance significantly. I tested locally with a simple query: `SELECT COUNT(*) FROM "zaken_rol" INNER JOIN "zaken_zaak" ON ("zaken_rol"."zaak_id" = "zaken_zaak"."identificatie_ptr_id")`, this query takes around 800ms on my machine

**Upd. by Anna**
The optimization is done in several steps (and PRs):
1. use only "pk" in pagination count for all viewsets - PR https://github.com/open-zaak/open-zaak/pull/1405 - **OZ 1.9**
2. optimize authorization filtering (initially developed in PR #1368, then cherry picked into new PR #1406) - **OZ 1.9**
3. introduce fuzzy pagination - https://github.com/open-zaak/open-zaak/pull/1368 - planned for **OZ 1.11**

stevenbal:
Fixes #1365

**Changes**

* Replace default pagination with FuzzyPagination that is capped at count 500
* Improve authorizations list filter lookup by simplifying the query

Notes:
- a constant `count` that uses an offset is not possible, because then this would be interpreted by the paginator as the total count being equal to the limit (e.g. 500) and stop adding next page links, even though there may be way more than 500 results), this means that the count will go up for each next page and also that the higher the page, the worse the performance


stevenbal:
Initial results show that `Besluit` and all Zaken API resources except `Zaak` itself are faster. Catalogi API resources didn't change in performance, because our test environment only has 100 of each *type. `EnkelvoudigInformatieObject` performance didn't improve either, which is slightly strange to me. I'll look into why `Zaak` and `EIO` stayed the same, performance wise

stevenbal:
After some investigation, I don't really know how to further optimize the EIO and Zaak queries. Some things I noticed:
- both use `DISTINCT`, which could be a performance bottleneck
- both map `vertrouwelijkheidaanduiding` to a number to perform the comparison with the maximum allowed `vertrouwelijkheidaanduiding` for the authenticated application. Because this mapping happens, an index does not improve performance at all
- The biggest performance bottleneck seems to be that the ids of the objects for which the user is authorized are retrieved in a subquery, and then another query is performed to get all the objects matching these ids. The inner query is a lot faster if performed by itself

annashamray:
@stevenbal for now it's not clear what change factors most into the optimization
Could you please provide the response time for several LIST endpoints in these cases:
* no optimization
* auth query optimization
* auth query + pagination count only with pk 
* auth query + pagination count only with pk + pagination cap

joeribekker (commented):
This is a deviation from the standard, so let's do it in 1.11

annashamray:
I want to separate optimization with and without the standard change:
* Let's keep only FuzzyPagination optimization in this PR (which is a standard deviation)
* I've cherry picked commits related to auth optimization into separate PR https://github.com/open-zaak/open-zaak/pull/1406 . Now we can include them into a minor release

joeribekker:
@annashamray what should we do with this PR?

annashamray:
@joeribekker  For now it's only `FuzzyPagination` that is not implemented in previous PRs. We can rebase this PR and analyze how much it actually speeds things up. What do you think?

annashamray:
Continued in https://github.com/open-zaak/open-zaak/pull/1622

PR #1378, opened 2023-05-01
:green_heart: [#1359] Fix ZGW postman tests

### LINKED ISSUE #1359: As a developer I want Postman tests in CI to work
Now Postmane tests in CI are failing: https://github.com/open-zaak/open-zaak/actions/runs/4688321713/jobs/8309422574?pr=1357
It happens because the data (or the ordering of data in Selectielijst API (https://referentielijsten-api.vng.cloud/api/v1/resultaten) was changed

The detailed explanation: 

* Postmane tests fail because one of the initial steps "Create Zaaktype"  returns 400 instead of 201.
* The error appears because we submit strange data in POST:

```
POST localhost:8000/catalogi/api/v1/zaaktypen

{
  "identificatie": "1",
  ...
  "selectielijstProcestype": "{{procestype_url}}"
}
```

* So Open Zaak responses with 400:
```
{
  │   type: 'http://localhost:8000/ref/fouten/ValidationError/',
  │   code: 'invalid',
  │   title: 'Invalid input.',
  │   status: 400,
  │   detail: '',
  │   instance: 'urn:uuid:a50ce0dc-6f22-4b55-8c61-cd32f8182bda',
  │   invalidParams: [
  │     {
  │       name: 'selectielijstProcestype',
  │       code: 'bad-url',
  │       reason: "De URL {{procestype_url}} kon niet opgehaald worden. Foutmelding: Invalid URL '{{procestype_url}}': No scheme supplied. Perhaps yo
  │ u meant http://{{procestype_url}}?"
  │     },
  │     { name: 'selectielijstProcestype', code: 'invalid', reason: 'Voer een geldige URL in.' }
  │   ]
  │ }
```
* The only place where `procestype_url` variable is assigned before this step is "(AC) Get Selectielijstklassen and Procestypen" step
* In this step we request a specific page of the url (GET https://referentielijsten-api.vng.cloud/api/v1/resultaten?page=3 ) and set `procestype_url` global var:
```
if(results[i].procestermijn == \"vast_te_leggen_datum\" && !pm.globals.get(\"selectielijstklasse_procestermijn_vast_te_leggen_datum_url\")) 
{
...
pm.globals.set(\"procestype_url\", results[i].procesType);
}
```
So we set it up only if there is a result with `procestermijn == "vast_te_leggen_datum"` on this page. But now there is no such a result here

The possible explanation is that VNG change the implementation of Selectielijst API from python to go and the ordering is now the same now.

**So the problem arised because:**
* we hardcoded the page number of external Selectielijst API in postman tests
* the external api changed its ordering

It should be fixed in order to keep Postman tests in CI

stevenbal:
Fixes #1359 

**Changes**

* Several fixes regarding postman tests (see https://github.com/VNG-Realisatie/gemma-postman-tests/pull/31)
* Added extra required mocks
* Workaround for fixture issues with Selectielijst service

### DIFF
--- a/bin/postman_tests.sh
+++ b/bin/postman_tests.sh
@@ -2,7 +2,7 @@
 
 set -x
 
-POSTMAN_TESTS_REF=6e34f01346545bb297bb181085464f8f32646769
+POSTMAN_TESTS_REF=fb877260d421a4b909fb8b46884174ddd3633b90
 
 # These client IDs and secrets are dummy variables that are only used by
 # the Docker build in Travis, so they can be public
--- a/docker/nginx/mocks.conf
+++ b/docker/nginx/mocks.conf
@@ -16,6 +16,11 @@ server {
         add_header Content-Type text/html;
         return 404 "";
     }
+
+    location /communicatiekanaal {
+        add_header Content-Type application/json;
+        return 200 '{"url":"http://mock-endpoints.local/communicatiekanaal","naam":"foo","omschrijving":"bar"}';
+    }
 }
 
 server {
--- a/fixtures/fixture.json
+++ b/fixtures/fixture.json
@@ -82,13 +82,13 @@
         "fields":{
             "label":"referentielijsten",
             "api_type":"orc",
-            "api_root":"https://referentielijsten-api.vng.cloud/api/v1/",
+            "api_root":"https://selectielijst.openzaak.nl/api/v1/",
             "client_id": "",
             "secret":"",
             "auth_type": "no_auth",
             "header_key": "",
             "header_value": "",
-            "oas": "https://referentielijsten-api.vng.cloud/api/v1/schema/openapi.yaml",
+            "oas": "https://selectielijst.openzaak.nl/api/v1/schema/openapi.yaml",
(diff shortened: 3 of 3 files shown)


stevenbal:
I rebased https://github.com/open-zaak/open-zaak/pull/1364 on this branch to show trigger the postman tests and show that they succeed

stevenbal:
@sergei-maertens I noticed two things that https://referentielijsten-api.vng.cloud/api/v1/resultaten returns all URLs as `http://...`, which was causing issues internally for OZ (when creating a zaaktype the procestype was being fetched, but no Service could be found because it was configured for `https://...`

I circumvented this by changing the fixture to use `http://...` instead and I also set the OAS url to https://selectielijst.openzaak.nl/api/v1/schema/openapi.yaml, because there is no redirect for that

joeribekker:
Not waiting for https://github.com/VNG-Realisatie/gemma-postman-tests/pull/31 because we need to move forward. I am confident that VNG will resolve the issue on their end. I sent an email to VNG about it.

joeribekker (approved):

PR #1385, opened 2023-06-07
:heavy_plus_sign: Add factory-boy for generate_data command

stevenbal:
This library was only included in the dev/ci requirements, which makes it impossible to run currently via Docker.

I think the library should be included in the base requirements, so it can be used in dockerized/k8s environments as well (@sjoerdie is trying to run this command for DH)


stevenbal:
Closing this because the issue was resolved by migrating the data instead of running the command directly on the pod

PR #1395, opened 2023-06-28
show map in the admin for Zaak.zaakgeometrie

### LINKED ISSUE #1060: Admin map widget for Zaak.zaakgeometrie field is broken on test.openzaak.nl
When I create a Zaak with zaakgeometrie Point coordinates the map widget in the admin doesn't work properly and gives a weird location in an ocean. I checked the coordinates on Google maps and they check out.

https://test.openzaak.nl/admin/zaken/zaak/30122/change/

https://test.openzaak.nl/admin/zaken/zaak/30261/change/

This was spotted when testing the Open-Forms map component and would be a visual nice-to-have for the demo of the component.

annashamray:
Fixes #1060


joeribekker (approved):
Kudos for the documentation on why this works the way it is.

PR #1398, opened 2023-06-30
:alembic: [#1326] add handler for FKOrServiceUrlField

### LINKED ISSUE #1326: Refactor processing external resources foreign keys to local resources
### Problem description
For now we need to add extra configuration if we have chained loose-fk objects, especially if we have external resource with link to local resource. For example external document with local `informatieobjecttype`
During the processing of loose-fk objects the virtual model is generated, which treats the urls to the local resources as external ones. 
It requires to add local APIs to the `Service` model (in the example above it should be Catalogi API), which is a workaround and should not be done.

### Solution direction
Refactor loose-fk implementation so the urls to the local objects are considered as local and don't require `Service` data

annashamray:
Fixes #1326

**Changes**
* try the second approach described in https://github.com/open-zaak/open-zaak/issues/1326#issuecomment-1612915469


annashamray:
So for now it works but only if CMIS is not enabled. 
If CMIS is used you need to configure local services

annashamray:
Docker build is not successful because of the dependency with git path. It should be fixed when https://github.com/maykinmedia/django-loose-fk/pull/27 is merged and a new version of `django-loose-fk` is released

src/openzaak/components/besluiten/models.py
```diff
@@ -195,10 +196,12 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
         # save previous zaak for triggers
-        # self._previous_zaak = self.zaak
-
-        self._previous_zaak = self._zaak
-        self._previous_zaak_url = self._zaak_url
+        # for virtual models we can't use FK field because of handler assertions
+        if isinstance(self, ProxyMixin):
```
joeribekker, line 200: This will require some merge-conflict resolving with the Besluiten-list PR.

annashamray: I'll take care of it

PR #1402, opened 2023-07-05
besluiten list api optimization

### LINKED ISSUE #1400: Improve performance in Besluiten-list
Version: [1.8.2-performance-fixes](https://github.com/open-zaak/open-zaak/releases/tag/1.8.2-performance-fixes)

Both Besluiten and EnkelvoudigInformatieobject, retrieve a Zaak for each Besluit/EIO. This should be optimized with a prefetch/select related:

You can see this behaviour here (for a list of 100 besluiten, it does 100x 1 query for each related zaak):

image

No prefetch:

image

Caused by the `__init__` method:

image

annashamray:
Fixes #1400

### DIFF
--- a/src/openzaak/components/besluiten/models.py
+++ b/src/openzaak/components/besluiten/models.py
@@ -219,13 +219,13 @@ class Besluit(AuditTrailMixin, APIMixin, models.Model):
     def previous_zaak(self):
         from openzaak.components.zaken.models import Zaak
 
-        if self._previous_zaak:
+        if getattr(self, "_previous_zaak", None):
             return self._previous_zaak
 
-        if self._previous_zaak_id:
+        if getattr(self, "_previous_zaak_id", None):
             return Zaak.objects.get(pk=self._previous_zaak_id)
 
-        if self._previous_zaak_url:
+        if getattr(self, "_previous_zaak_url", None):
             remote_model = apps.get_model("zaken", "Zaak")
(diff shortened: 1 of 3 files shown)


joeribekker (changes requested):
I think this is the wrong direction entirely.

Why not simply:

```python
    def __init__(self, *args, **kwargs):

        self._previous_zaak_id = self._zaak_id  # Store the already fetched ID
        self._previous_zaak_url = self._zaak_url
```

and 

```python
    @property
    def previous_zaak(self):
        if self._previous_zaak_id:
            return Zaak.objects.get(pk=self._previous_zaak_id)  # Retrieve Zaak based on previous ID

        if self._previous_zaak_url:
            remote_model = apps.get_model("zaken", "Zaak")
            return AuthorizedRequestsLoader().load(
                url=self._previous_zaak_url, model=remote_model
            )

        return None
```

change one unit test and done?

annashamray:
@joeribekker Thank you for the simpler solution. I've checked it locally and it doesn't hit DB now, I've replaced the previous solution with the new one

joeribekker (changes requested):
Can you add a test that shows that the if-statement for max_vertrouwelijkheid is solved.

annashamray:
@joeribekker the regression test is added

joeribekker (approved):

PR #1434, opened 2023-08-02
add '_zoek' endpoint for documents

### LINKED ISSUE #1433: As a developer, I want to add 'enkelvoudiginformatieobject/_zoek' endpoint
This is a part of adding support for Documenten API 1.3
Master US - https://github.com/open-zaak/open-zaak/issues/1410
Related VNG issue - https://github.com/VNG-Realisatie/gemma-zaken/issues/1881

**NOTE:** according to OAS this endpoint supports only one parameter `uuid__in` and doesn't support query params of the list endpoint, therefore has different behavior than `zaken/_zoek` endpoint. Related issue - https://github.com/VNG-Realisatie/gemma-zaken/issues/2294

annashamray:
Fixes #1433

### DIFF
--- a/src/openzaak/components/documenten/api/serializers.py
+++ b/src/openzaak/components/documenten/api/serializers.py
@@ -829,6 +829,13 @@ class UnlockEnkelvoudigInformatieObjectSerializer(serializers.ModelSerializer):
         return self.instance
 
 
+class EIOZoekSerializer(serializers.Serializer):
+    uuid__in = serializers.ListField(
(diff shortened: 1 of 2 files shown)


joeribekker (approved):
Hmm, technically I see that the search param is "uuid_In" while we have "uuid__in" (lowercase, double underscore). I suspect an issue at VNG zo I created this issue: https://github.com/VNG-Realisatie/gemma-zaken/issues/2300

I approve with this mismatch.

PR #1443, opened 2023-08-21
Update Standard for public code assessment to 0.7.1

### LINKED ISSUE #1442: Update the Standard for Public Code Assessment to version 0.7.1
Update the [assessment](https://github.com/open-zaak/open-zaak/blob/main/docs/introduction/open-source/public_code.md) to [version 0.7.1](https://github.com/publiccodenet/standard/releases/tag/0.7.1).

Ainali:
Fixes #1442

**Changes**

Besides updating the requirement changes, this also updates the assessment to HTML-format which will make future updates much easier (see [process](https://about.publiccode.net/activities/releasing-the-standard/assessment-update.html)).


Ainali:
You are making some good points. We can restructure the PR into two commits, as you suggest, so that the first one introduces the format change and the second one the content change. That should make it a lot easier to follow the changes in the diff.

Regarding the styling, those are just some base from the template. It can certainly be removed or changed to something entirely different (as was made [here](https://www.govdirectory.org/standard-for-public-code/)).

joeribekker:
Before putting in the legwork, let me ponder on it :)

Ainali:
It was a good exercise for us to make this in a more proper way, so now you have the commits separately to view.

joeribekker:
@annashamray could you see if we can embed this page as-is in our Sphinx setup? Maybe its a matter of referring to the full filename (including `.html`) but I'm not sure.

annashamray:
@joeribekker 
1. I haven't found a way to include html page as a source in the Sphinx documentation. 
For now Sphinx supports `rst` and `md` files as sources and that's it. There is no parser for html files as far as a know. 

2. We can treat `public_code.html` as external link and have it in the ToC in such a way:
    ```
    Public Code <https://htmlpreview.github.io/?https://raw.githubusercontent.com/open-zaak/open-zaak/917de28ae5637d168ab022c9bd92781620ed7383/docs/introduction/open-source/public_code.html>
    ```
    But then it isn't included in the built documentation and the page doesn't have left panel with ToC. 
    In this case there is no sense to keep it in the `docs/` folder - we can move it into the root of the repo and then use link to github raw files.

ericherman:
> 2\. In this case there is no sense to keep it in the `docs/` folder - we can move it into the root of the repo and then use link to github raw files.

If this is the solution, we can easily rebase and add a `git mv` to wherever is best, just let us know.

joeribekker:
Thanks @Ainali for making the changes and @ericherman for the offer to move it.

I think @annashamray can move it as well? I think it's best to treat it as an asset, so it stays part of the docs folder. Can you do this Anna? Then we pack it together with the 1.10 release.

annashamray:
@joeribekker I don't think it's possible to have a relative link to something like `_assets/public_code.html` in ToC
It seems that you can have either have source files (.rst or .md) or links to external resources there. It you use internal link in ToC like `Public code <../public_code.html>` it returns error "WARNING: toctree contains reference to nonexisting document 'introduction/public_code.html'" 

So it should be an external resource which exists before Sphinx documentations is being generated

While I was researching it I found another possible solutio - we can have a source .rst file with embedded html from another file
So we can create `public_code.rst` file with the content:
```
.. _introduction_open-source_public:

Public code
===========

.. raw:: html
    :file: public_code.html
```
It will also allow embed the external html into Sphinx documentation layout 
It will look something like this:
image

The only downside is that we will have two document titles here as you can see, because rst title is mandatory for ToC generation

joeribekker:
This works for me Anna!

annashamray:
Thank you @Ainali and @ericherman for this PR and the changes! 
I can't work in your remote repository, so I've created a branch `Ainali-issue/1442` in this repo, where I've pulled your commits and I'll keep working there
Here is the new PR - https://github.com/open-zaak/open-zaak/pull/1496 

@joeribekker please check the changes in the new PR.
I also suggest to keep discussion there

PR #1487, opened 2023-10-17
change unique check for ZaakType

### LINKED ISSUE #1486: As a developer, I want to change unique constraint for ZaakType
**Description**
It's a part of Catalogi 1.2 support
Master US - https://github.com/open-zaak/open-zaak/issues/1411

Replace unique constraint of  `ZaakType.omschrijving & ZaakType.catalogus` with `ZaakType.identificatie & ZaakType.catalogus`

annashamray:
Fixes #1486
Depends on https://github.com/VNG-Realisatie/gemma-postman-tests/pull/41

### DIFF
--- a/bin/postman_tests.sh
+++ b/bin/postman_tests.sh
@@ -2,7 +2,7 @@
 
 set -x
 
-POSTMAN_TESTS_REF=e8d1b827ebc7927255f8f462bcd8a2987ef8a99e
+POSTMAN_TESTS_REF=d13c71796be9360a22f86c98a39fa0352f08dafe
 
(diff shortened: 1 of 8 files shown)


joeribekker (changes requested):
I'm missing the data migration?

annashamray:
@joeribekker 
To be honest I don't know what data migration should be here. If in our DB there are two zaken with the same identificatie and overlapping dates what should we do with them?

joeribekker (approved):
Right, the unique check is on serializer level due to versioning...

Issue #1497, opened 2023-11-03
Zaaktype imports: improve errormessage when overlap exists, improve IOT selectors


Coperh:
Ref. DH 530

**Description**

- [x] If there is an overlap-error, mention for which informatieobjecttype this occurred
- [x] If there is an error when importing, reuse the submitted data to avoid having to re-select IOT's
- [x] Sort the available options per selector alphabetically

Issue #1493, opened 2023-11-03
Auto-match IOT's when importing zaaktypes based on name


Coperh:
Ref. DH 500

**Description**
For each IOT in the import I need to select the existing IOT, even though the names match exactly.
This can be done automatically

PR #1499, opened 2023-11-08
Feature/1497 zaaktype imports various improvments

Coperh:
Fixes: #1497

Various zaaktype import improves from taiga issue 530

**Changes**

* Sort the available options per selector alphabetically
*  Reuse the submitted data to avoid having to re-select IOT's

### DIFF
--- a/src/openzaak/components/catalogi/admin/admin_views.py
+++ b/src/openzaak/components/catalogi/admin/admin_views.py
@@ -185,5 +185,4 @@ class CatalogusZaakTypeImportSelectView(
             return HttpResponseRedirect(reverse("admin:catalogi_catalogus_changelist"))
         except (CommandError, IntegrityError) as exc:
             messages.add_message(request, messages.ERROR, exc)
-        context = self.get_context_data(**kwargs)
         return TemplateResponse(request, self.template_name, context)
--- a/src/openzaak/components/catalogi/admin/admin_views.py
+++ b/src/openzaak/components/catalogi/admin/admin_views.py
@@ -198,9 +198,6 @@ class CatalogusZaakTypeImportSelectView(
 
                 besluittypen_uuid_mapping = {}
                 if besluittype_forms:
-                    besluittype_forms = BesluitTypeFormSet(
-                        request.POST, prefix="besluittype"
-                    )
                     if besluittype_forms.is_valid():
                         besluittypen_uuid_mapping = construct_besluittypen(
                             request.session["besluittypen"],
--- a/src/openzaak/components/catalogi/admin/admin_views.py
+++ b/src/openzaak/components/catalogi/admin/admin_views.py
@@ -94,8 +94,10 @@ class CatalogusZaakTypeImportSelectView(
                 .order_by("omschrijving", "-datum_begin_geldigheid")
                 .distinct("omschrijving")
             }
-
+            # sort alphabetically, save for use in post
             iotypen = sorted(iotypen, key=lambda x: x["omschrijving"])
+            self.request.session["iotypen"] = iotypen
+
             iotype_forms = InformatieObjectTypeFormSet(
                 initial=[
                     {"existing": iot_dict.get(instance["omschrijving"])}
@@ -119,7 +121,10 @@ class CatalogusZaakTypeImportSelectView(
                 .order_by("omschrijving", "-datum_begin_geldigheid")
                 .distinct("omschrijving")
             }
+            # sort alphabetically, save for use in post
             besluittypen = sorted(besluittypen, key=lambda x: x[0]["omschrijving"])
+            self.request.session["besluittypen"] = besluittypen
+
             besluittype_forms = BesluitTypeFormSet(
                 initial=[
                     {"existing": besluittypen_dict.get(instance["omschrijving"])}
--- a/src/openzaak/components/catalogi/admin/admin_views.py
+++ b/src/openzaak/components/catalogi/admin/admin_views.py
@@ -87,7 +87,7 @@ class CatalogusZaakTypeImportSelectView(
         catalogus = Catalogus.objects.get(pk=catalogus_pk)
 
         iotypen = self.request.session.get("iotypen")
-        if iotypen:
+        if iotypen and "iotype_forms" not in context:
             iot_dict = {
                 obj.omschrijving: obj.pk
                 for obj in InformatieObjectType.objects.filter(catalogus=catalogus)
@@ -112,7 +112,7 @@ class CatalogusZaakTypeImportSelectView(
             context["iotype_forms"] = iotype_forms
 
         besluittypen = self.request.session.get("besluittypen")
-        if besluittypen:
+        if besluittypen and "besluittype_forms" not in context:
             besluittypen_dict = {
                 obj.omschrijving: obj.pk
                 for obj in BesluitType.objects.filter(catalogus=catalogus)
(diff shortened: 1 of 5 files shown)


Coperh:
This will probably have tonnes of merge conflicts with #1493

src/openzaak/components/catalogi/tests/admin/test_import_export_zaaktype.py
```diff
@@ -1074,6 +1244,139 @@ def test_import_zaaktype_invalid_eigenschap(self, *mocks):
...
+            _(
+                "Dit {} komt al voor binnen de catalogus en opgegeven geldigheidsperiode."
+            )
+            .format(informatieobjecttype._meta.verbose_name)
+            .title()
+        )
+        expected_html = f'<ul class="errorlist"><li>begin_geldigheid: overlap — {error_text}</li></ul>'
+
+        self.assertIn(
+            expected_html, response.text,
```
Coperh, line 1302: There does not seem to be an easy way to test if the error appears properly. Might be easier if I use Django TestCase instead of Webtest

src/openzaak/components/catalogi/admin/admin_views.py
```diff
@@ -167,4 +177,25 @@ def post(self, request, *args, **kwargs):
             return HttpResponseRedirect(reverse("admin:catalogi_catalogus_changelist"))
         except (CommandError, IntegrityError) as exc:
             messages.add_message(request, messages.ERROR, exc)
+
+        catalogus_pk = kwargs.get("catalogus_pk")
+        catalogus = Catalogus.objects.get(pk=catalogus_pk)
+
+        if iotype_forms:
+            for i, form in zip(request.session["iotypen"], iotype_forms.forms):
```
Coperh, line 185: This feels kinda hacky but otherwise it I would need to iterative over the labels above. Not sure if this actually improves performance.

src/openzaak/components/catalogi/admin/admin_views.py
```diff
@@ -167,4 +177,25 @@ def post(self, request, *args, **kwargs):
...
+                form._bound_fields_cache["existing"].label = (
+                    str(catalogus) + " - " + i["omschrijving"]
+                )
+            kwargs["iotype_forms"] = iotype_forms
```
Coperh, line 189: This feels kinda hacky but otherwise it I would need to iterative over the labels above. Not sure if this actually improves performance.

src/openzaak/components/catalogi/admin/admin_views.py
```diff
@@ -129,21 +129,30 @@ def get(self, request, *args, **kwargs):
...
+
+        if "iotype-TOTAL_FORMS" in request.POST:
+            iotype_forms = InformatieObjectTypeFormSet(request.POST, prefix="iotype",)
+        else:
+            iotype_forms = None
+
+        if "besluittype-TOTAL_FORMS" in request.POST:
+            besluittype_forms = BesluitTypeFormSet(request.POST, prefix="besluittype",)
+        else:
+            besluittype_forms = None
```
Coperh, line 141: Can also set labels here

src/openzaak/components/catalogi/tests/admin/test_import_export_zaaktype.py
annashamray: If you want another way to search for error list I can think of two other options: 

1. You can use context from response both here and in usual Django TestCase:
``` iotype_forms = response.context["iotype_forms"] ```
And then assert iotype_forms.errors

2. In Webtest you can use `response.html` which is a Beautiful Soup object to scrape web page
It helps to look through html, so you can make sure that you have an error list rendered together with the correct iotype with somethink like this:
    ```
    iotype_tag = response.html.find(id="id_iotype-0-existing")
    errorlist_tag = iotype_tag.parent(class_="errorlist")
    ```
    Then you can assert that this tag exists and has the expected content

**Review by annashamray — CHANGES_REQUESTED** [review]
Looks nice, so a couple of minor remarks

File: src/openzaak/components/catalogi/admin/admin_views.py (hunk 3 above)
Comment (Coperh) [reply src/openzaak/components/catalogi/admin/admin_views.py:189]: All the fields default to `Existing` if you do not manually set the labels after getting the form from post. 

I places this at the end of the code since you only need the labels if the code fails. It might make more sense just to initialize the labels regardless.

File: src/openzaak/components/catalogi/tests/admin/test_import_export_zaaktype.py (hunk 1 above)
Comment (Coperh) [reply src/openzaak/components/catalogi/tests/admin/test_import_export_zaaktype.py]: I swear I tried iotype_forms.errors but it didn't work for me. Anyway, I changed it.

File: src/openzaak/components/catalogi/admin/utils.py [hunk 5]
```diff
@@ -94,20 +94,29 @@ def construct_iotypen(iotypen, iotype_form_data):
                 instance = InformatieObjectType(**data)
                 instance.omschrijving_generiek = omschrijving_generiek
             else:
+                error_message = ", ".join(
+                    [
+                        f"{k}: {v[0].code} — {v[0].title()}"
```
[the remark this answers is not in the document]
Coperh: This case I don't think should happened but I made a recursive formatter thingy that should handle this if it happends

src/openzaak/components/catalogi/admin/admin_views.py
annashamray: Why can't it be solved by putting formsets into the context?

annashamray: Ok I managed to reproduce it.
But you create formsets in GET and in POST in different ways - in POST call you don't have `form_kwargs` parameter, which has labels. If you add `form_kwargs` when initializing formsets in POST the labels would be correct
At least it works on my machine

src/openzaak/components/catalogi/admin/admin_views.py
```diff
@@ -112,14 +114,18 @@ def get_context_data(self, **kwargs):
...
                 obj.omschrijving: obj.pk
                 for obj in BesluitType.objects.filter(catalogus=catalogus)
                 .order_by("omschrijving", "-datum_begin_geldigheid")
                 .distinct("omschrijving")
             }
+            # sort alphabetically, save for use in post
             besluittypen = sorted(besluittypen, key=lambda x: x[0]["omschrijving"])
+            self.request.session["besluittypen"] = besluittypen
+
             besluittype_forms = BesluitTypeFormSet(
```
Coperh, line 129: Saves the new order so it is the same when the order fails. 

Is it possible to test the label of a form field with Django webtest or regular test? I cannot find info online.

src/openzaak/components/catalogi/admin/admin_views.py
```diff
@@ -148,28 +154,62 @@ def get(self, request, *args, **kwargs):
         return super().get(request, *args, **kwargs)
 
     def post(self, request, *args, **kwargs):
+
+        catalogus_pk = kwargs.get("catalogus_pk")
+        catalogus = Catalogus.objects.get(pk=catalogus_pk)
+
+        if "iotype-TOTAL_FORMS" in request.POST:
+            iotype_forms = InformatieObjectTypeFormSet(
```
annashamray, line 162: This comment is optional but I think it would be more readable to refactor this view and to initialize formsets in one place with parameters depending on if it's POST of GET. Like django does in it's `FormView` and its `get_form_kwargs` method. It will make code DRYer and will reduce a good amount of if-else from `post` method.

src/openzaak/components/catalogi/admin/utils.py
```diff
@@ -22,6 +22,19 @@
 setattr(REQUEST, "version", "1")
 
 
+def format_serializer_errors(errors, is_fk=False):
+
```
annashamray, line 26: Please add a docstring describing the function and parameters

src/openzaak/components/catalogi/tests/admin/test_utils.py
```diff
@@ -0,0 +1,75 @@
+# SPDX-License-Identifier: EUPL-1.2
+# Copyright (C) 2023 Dimpact
+from django.test import TestCase
+
+from rest_framework.exceptions import ErrorDetail
+
+from openzaak.components.catalogi.admin.utils import format_serializer_errors
+
+
+class SerializerErrorFormatterTests(TestCase):
```
annashamray, line 10: Nice tests!

src/openzaak/components/catalogi/admin/admin_views.py
annashamray: For regular Django forms you can check the label pretty straightforward:
```
my_form = response.context["form"]
my_field = my_form.fields["my_field"]
my_field.label
```
In this case `besluittype_forms` is a Formset, i.e. a list of forms, so if you know the index of your form, you can  check the labels of this particular form
For example you have 1 form in your formset, so:

```
besluittype_forms = response.context["besluittype_forms"]
besluittype_forms.forms[0].fields["exisiting"].label
```
you can also use `Formset.get_form_kwargs` method:
```
besluittype_forms = response.context["besluittype_forms"]
besluittype_forms.get_form_kwargs(0).labels
```

if you want to use webtest, then I think you need to search html for `label` tag with `response.html.find_all` or something similar

annashamray (changes requested):
For the future:

In Open Zaak project we prefer `rebase` over `merge` to keep commit history cleaner. 
So in the future when you have merge conflicts with `main` branch, please rebase your branch on `main`

src/openzaak/components/catalogi/admin/admin_views.py
Coperh: Moved everything into the `get_context_data` funciton

src/openzaak/components/catalogi/admin/admin_views.py
```diff
@@ -148,28 +167,31 @@ def get(self, request, *args, **kwargs):
         return super().get(request, *args, **kwargs)
 
     def post(self, request, *args, **kwargs):
+
+        context = self.get_context_data(**kwargs)
+
         try:
-            with transaction.atomic():
+            with (transaction.atomic()):
```
annashamray, line 174:
```suggestion
            with transaction.atomic():
```

annashamray (approved):
Nice work! It looks good

PR #1507, opened 2023-11-14
:construction_worker: [#1502] add trivy image scaner to CI

### LINKED ISSUE #1502: As municipality I want an up-to-date and secure image
..so that I can deploy without worries.

Ref: Amsterdam - Aanpassing Docker

**Tasks**

_Open Zaak_

* [x] Add Trivy to Github Actions. An initial  setup was made here: https://github.com/open-formulieren/open-forms/pull/3129/files.
* [x] Make Trivy report to Github security/dependabot so we can check all issues. Trivy however should never cause failures.
* [x] Remove Vim, and add Nano to the Docker image
* [x] Add apt-get update and upgrade statements to Dockerfile (see also the link in first task)

_Open Notificaties_

* [x] Add Trivy to Github Actions
* [x] Make Trivy report to Github security/dependabot so we can check all issues. Trivy however should never cause failures.
* [x] Remove Vim, and add Nano to the Docker image
* [x] Add apt-get update and upgrade statements to Dockerfile

annashamray:
Fixes #1502

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ RUN apt-get update && apt-get upgrade -y && apt-get install -y --no-install-reco
 WORKDIR /app
 
 COPY ./requirements /app/requirements
-RUN pip install pip -U
+RUN pip install pip setuptools -U
 RUN pip install -r requirements/production.txt
 
 


Coperh (commented):
I am not really familiar with github actions or Docker but it seems fine to me.

PR #1508, opened 2023-11-16
Feature/1479 zaak eigenschap validation

### LINKED ISSUE #1479: `waarde` of ZAAKEIGENSCHAP is not validated by its EIGENSCHAP SPECIFICATIE
**General information**
Open Zaak version: 1.8.2

**Describe the bug**
The value of a ZAAKEIGENSCHAP is not validated by the EIGENSCHAP SPECIFICATIE of the EIGENSCHAP.

**To Reproduce**
1. Create ZAAKTYPE
2. Create EIGENSCHAP for ZAAKTYPE (1) with a EIGENSCHAP SPECIFICATIE that has `formaat`: `Datum` (i.e. length must be 8).
3. Create ZAAK of ZAAKTYPE (1)
4. Create ZAAKEIGENSCHAP in ZAAK (3) with `waarde`: `asdasdasdasasd`. 
5. See there is no validation. Not through the admin nor the REST API.

**Expected behavior**
EIGENSCHAP SPECIFICATIE to be used as validation for the `waarde` of ZAAKEIGENSCHAP.

annashamray:
Fixes #1479

### DIFF
--- a/src/openzaak/components/zaken/api/validators.py
+++ b/src/openzaak/components/zaken/api/validators.py
@@ -586,7 +586,9 @@ class ZaakEigenschapValueValidator:
         if not eigenschap or not waarde:
             return
 
-        if not match_eigenschap_specificatie(
-            eigenschap.specificatie_van_eigenschap, waarde
-        ):
+        spec = eigenschap.specificatie_van_eigenschap
+        if not spec:
+            return
+
+        if not match_eigenschap_specificatie(spec, waarde):
             raise serializers.ValidationError(self.message, code=self.code)
--- a/src/openzaak/components/zaken/api/validators.py
+++ b/src/openzaak/components/zaken/api/validators.py
@@ -574,7 +574,9 @@ def match_eigenschap_specificatie(spec, value: str) -> bool:
 
 class ZaakEigenschapValueValidator:
     code = "waarde-incorrect-format"
-    message = _("The 'waarde' value doesn't match the related eigenschap.specificatie")
+    message = _(
+        "The 'waarde' value doesn't match the related eigenschap.specificatie '{spec}'"
+    )
     requires_context = True
 
     def __call__(self, attrs, serializer):
@@ -594,4 +596,6 @@ class ZaakEigenschapValueValidator:
             return
 
         if not match_eigenschap_specificatie(spec, waarde):
-            raise serializers.ValidationError(self.message, code=self.code)
+            raise serializers.ValidationError(
+                self.message.format(spec=spec), code=self.code
+            )
--- a/src/openzaak/components/zaken/api/validators.py
+++ b/src/openzaak/components/zaken/api/validators.py
@@ -2,7 +2,8 @@
 # Copyright (C) 2019 - 2020 Dimpact
 import json
 import logging
-from datetime import date
+import re
+from datetime import date, datetime
 from typing import Iterable, Optional
 
 from django.conf import settings
@@ -20,6 +21,7 @@ from vng_api_common.validators import (
     URLValidator,
 )
 
+from openzaak.components.catalogi.constants import FormaatChoices
 from openzaak.components.documenten.constants import Statussen
 from openzaak.components.documenten.models import (
     EnkelvoudigInformatieObject,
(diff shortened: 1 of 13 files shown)


src/openzaak/components/zaken/api/viewsets.py
```diff
@@ -722,6 +722,9 @@ class ZaakEigenschapViewSet(
 
     Maak een ZAAKEIGENSCHAP aan.
 
+    **Er wordt gevalideerd op**
+    - `waarde` moet worden gevalideerd tegen `eigenschap.specificatie`.
```
joeribekker, line 726: I guess we need to add "EXPERIMENTEEL" here.

joeribekker (dismissed):
Can we add a management command that checks and reports all eigenschappen that are not compliant? Out put can be like:

```
Zaak {uuid} has Eigenschap {uuid} with waarde="{val}" that does not match specificatie="{}"
```

annashamray (commented):
Management comman `check_zaak_eigenschappen` is added

joeribekker:
Can we add one more thing? :)

An environment variable that enables the validation? By default it should off. The person who requested this already ran into trouble with this branch.

src/openzaak/components/zaken/api/viewsets.py
```diff
@@ -767,7 +767,13 @@ def perform_destroy(self, instance):
     ),
     create=extend_schema(
         summary="Maak een ZAAKEIGENSCHAP aan.",
-        description="Maak een ZAAKEIGENSCHAP aan.",
+        description=(
+            "Maak een ZAAKEIGENSCHAP aan.\n"
+            "\n"
+            "**Er wordt gevalideerd op**\n"
+            "\n"
+            "- **EXPERIMENTEEL** `waarde` moet worden gevalideerd tegen `eigenschap.specificatie`."
```
SonnyBA, line 775: Small nitpick but this description might make it seem like the validation would always run (which is the case if the option is enabled through the added env variable).
```suggestion
            "- **EXPERIMENTEEL** `waarde` moet worden gevalideerd tegen `eigenschap.specificatie` indien `ZAAK_EIGENSCHAP_WAARDE_VALIDATION` op `True` staat."
```

SonnyBA (approved):
Added a suggestion about the description used in the API documentation but overall looks good to me

src/openzaak/components/zaken/api/validators.py
```diff
@@ -525,3 +527,71 @@ def __call__(self, attrs, serializer):
...
+            return False
+        else:
+            return True
+
+    return True
+
+
+class ZaakEigenschapValueValidator:
+    code = "waarde-incorrect-format"
+    message = _("The 'waarde' value doesn't match the related eigenschap.specificatie")
```
Coperh, line 577: Maybe this should be more specific and mention what the expected format is i.e. `does not match eigenschap.specificatie datum [8, #1]` or `str(spec)`

src/openzaak/components/zaken/api/validators.py
```diff
@@ -525,3 +527,71 @@ def __call__(self, attrs, serializer):
...
+    requires_context = True
+
+    def __call__(self, attrs, serializer):
+        if not settings.ZAAK_EIGENSCHAP_WAARDE_VALIDATION:
```
Coperh, line 581: Not really a suggestion, but I wonder if you can conditionally add it to the serializer validators so it does not check this every time.

I am not sure if this is considered DRF bad practice.

src/openzaak/components/zaken/api/validators.py
```diff
@@ -525,3 +527,71 @@ def __call__(self, attrs, serializer):
...
+
+        eigenschap = get_from_serializer_data_or_instance(
+            "eigenschap", attrs, serializer
+        )
+        waarde = get_from_serializer_data_or_instance("waarde", attrs, serializer)
+
+        if not eigenschap or not waarde:
+            return
+
+        spec = eigenschap.specificatie_van_eigenschap
```
Coperh, line 592: Another non suggestions, but the `specificatie_van_eigenschap` definition is weird. Can be blank and null but also has CASCADE

Coperh (approved):
Seems functionally correct and works as expected.
Just a few comments that are not really necessary.

src/openzaak/components/zaken/api/validators.py
annashamray: I haven't seen such thing in the docs. I think it's expected that there is a list of validators for each field and the conditions are put inside the validator. 
In this particular case I don't think that it will win a lot of performance, since we don't do any extra DB or HTTP requests before the condition. 

And even if it happens the way to go is to cache something on the serializer level and then use context to pass it to the validator

src/openzaak/components/zaken/api/validators.py
annashamray: I think it's `blank=False`. I don't know why it is both `null=True` and `on_delete=CASCADE`, perhaps something is a leftover from the old code
It's out of scope of this issue, so I'll leave it as it is for now.

src/openzaak/components/zaken/tests/test_zaakeigenschap.py
```diff
@@ -451,3 +461,257 @@ def test_zaakeigenschap_create_jwt_expired(self):
...
+            "zaak": self.zaak_url,
+            "eigenschap": f"http://testserver{reverse(eigenschap)}",
+            "waarde": "20201201155555",
+        }
+
+        response = self.client.post(self.url, data)
+
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED, response.data)
+
+    def test_datetime_incorrect_format(self):
```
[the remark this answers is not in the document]
annashamray: tests are changed

src/openzaak/components/zaken/api/validators.py
annashamray: Good idea. I've changed the message

PR #1516, opened 2023-11-28
:bookmark: release 1.10.1

annashamray:


annashamray:
It looks like it's fixed here
image

joeribekker (approved):

PR #1517, opened 2023-11-28
:memo:[#1490] return text choice descriptions

### LINKED ISSUE #1490: As developer, I want to restore descriptions for model.TextChoices
**Description**
After django choices were removed in https://github.com/open-zaak/open-zaak/issues/1418, the descriptions for some enums were sacrificed
Let's bring them back

**Tasks**:
* [x] restore descriptions in `commonground-api-common`
* [x] restore descriptions in `openzaak`

Coperh:
Fixes #1490 

**Changes**

Readd descriptions to text choices


Coperh:
Depends on: https://github.com/maykinmedia/commonground-api-common/pull/17 to be merged and released.

annashamray (changes requested):
Could you regenerate OAS for all components? Some of them (like `Rol` from `zaken` component) use constants from `commmonground-api-common`, if the description is restored there, the OAS should be updated

annashamray (approved):
:+1:

PR #1520, opened 2023-11-30
create zaak with incorrect zaaktype

### LINKED ISSUE #1418: As developer, I want to get rid of Django Choices
because the library is not supported anymore

**Description**
The migration manual - https://django-choices.readthedocs.io/en/latest/migrating.html 

**Tasks**

* [ ] update dependencies to versions which don't use django choices lib
* [x] convert all constants from `DjangoChoices` to `models.TextChoices`
* [x] remove `django-choices` from requirements

### LINKED ISSUE #1481: Create zaak returns 403 instead of 404 if zaaktype doesn't exist
**General information**
Open Zaak version: 1.8.2

**Describe the bug**
If `zaaktype` doesn't exist when creating a `zaak` a 403 is returned instead of a 404. This is confusing behavior when debugging the cause of the 4xx response.

**To Reproduce**
1. Make sure `zaaktype` doesn't exist
2. Create `zaak` with non-existing `zaaktype`.

**Expected behavior**
A 404 detailing what is not found. I.e., <zaaktype-url> is not found or similar.

annashamray:
Fixes #1481

### DIFF
--- a/src/openzaak/components/documenten/tests/test_validation.py
+++ b/src/openzaak/components/documenten/tests/test_validation.py
@@ -163,7 +163,7 @@ class EnkelvoudigInformatieObjectTests(JWTAuthMixin, APITestCase):
 
     @temp_private_root()
     def test_inhoud_incorrect_padding(self):
-        iotype = InformatieObjectTypeFactory.create()
+        iotype = InformatieObjectTypeFactory.create(concept=False)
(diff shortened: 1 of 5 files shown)


Coperh (approved):
Looks good. There does not seem to be any obvious issue to me.

No longer get the 403 error for non-existent zaaktypes

PR #1524, opened 2023-12-05
Feature/1522 documenten expand

### LINKED ISSUE #1522: As developer, I want to add `expand` to Documenten endpoints
Master US - https://github.com/open-zaak/open-zaak/issues/1464

**Tasks**
Add `expand` to:
* [x] Enkelvoudiginformatieobjecten 
* [x] Gebruiksrechten 
* [x] Verzendingen

annashamray:
Fixes #1522

Depends on https://github.com/open-zaak/open-zaak/pull/1526

### DIFF
--- a/src/openzaak/components/documenten/api/serializers.py
+++ b/src/openzaak/components/documenten/api/serializers.py
@@ -17,7 +17,6 @@ from django.db import transaction
 from django.utils.http import urlencode
 from django.utils.translation import ugettext_lazy as _
 
-from django_loose_fk.drf import FKOrURLField
 from drc_cmis.utils.convert import make_absolute_uri
 from drf_extra_fields.fields import Base64FileField
 from humanize import naturalsize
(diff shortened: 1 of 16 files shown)


src/openzaak/components/documenten/api/serializers.py
```diff
@@ -210,6 +196,9 @@ def get_object(self, view_name, view_args, view_kwargs):
         except (TypeError, AttributeError):
             self.fail("does_not_exist")
 
+    def get_attribute(self, instance):
+        return instance.get_informatieobject()
+
```
joeribekker, line 201: Can you explain why you did this?

annashamray: `EnkelvoudigInformatieObjectHyperlinkedRelatedField` is used for all documenten subresources like `Gebruiksrechten` for `informatieobject` field.
The custom field was needed because:
* in case CMIS is enabled CMIS is requested
* in case CMIS is disabled the subresource has FK to `EnkelvoudigInformatieObjectCanonical`, which doesn't have `uuid`, so `canonical.latest_version` property should be used. 

In the old implementation `get_url` method  was customized to solve this issue.  `get_url` is used inside `to_representation` method, which is used on `field.get_attribute` value
So how it was before:
* `Gebruiksrechten` serializer calls `field.get_attribute` and receives the fk object which is `EnkelvoudigInformatieObjectCanonical`
* `field.to_representation` is called with the value from previous step as parameter
* `field.get_url` is called inside `field.to_representation` to convert model objects to the url.

I thought that `get_url` was not the best place for this logic, because we need to change nt the url  implementation but retrieving model object. Moreover all subresources models have `get_informatieobject` which is used for permissions and already has the implementation to get the correct `informatieobject` , so I moved the logic from `get_url` to `get_attribute` method

PR #1534, opened 2023-12-18
:sparkles:[#1504] show warning if zaaktype import does nothing

### LINKED ISSUE #1504: Zaaktype imports look like they are working, but aren't
Ref. DH 509

**Description**
When importing this zaaktype, it appears like this is working as expected and the zaaktype is available.

However no zaaktype is created…

Coperh:
Fixes #1504

### DIFF
--- a/src/openzaak/components/catalogi/admin/admin_views.py
+++ b/src/openzaak/components/catalogi/admin/admin_views.py
@@ -86,10 +86,6 @@ class CatalogusZaakTypeImportUploadView(
                 except CommandError as exc:
                     messages.add_message(request, messages.ERROR, exc)
 
-                except ValueError as exc:
-
-                    messages.add_message(request, messages.WARNING, exc)
-
         return TemplateResponse(request, self.template_name, context)
 
 
--- a/src/openzaak/components/catalogi/admin/admin_views.py
+++ b/src/openzaak/components/catalogi/admin/admin_views.py
@@ -9,6 +9,8 @@ from django.db.utils import IntegrityError
 from django.http.response import HttpResponseRedirect
 from django.template.response import TemplateResponse
 from django.urls import reverse
+from django.utils.html import escape
+from django.utils.safestring import mark_safe
 from django.utils.translation import ugettext_lazy as _
 from django.views.generic import FormView, TemplateView
 
@@ -69,17 +71,41 @@ class CatalogusZaakTypeImportUploadView(
             else:
                 try:
                     with transaction.atomic():
-                        import_zaaktype_for_catalogus(
+                        (
+                            files_found,
+                            files_not_found,
+                            files_received,
+                        ) = import_zaaktype_for_catalogus(
                             request.session["identificatie_prefix"],
                             catalogus_pk,
                             request.session["file_content"],
                             {},
                             {},
                         )
-
-                    messages.add_message(
-                        request, messages.SUCCESS, _("ZaakType successfully imported")
-                    )
+                    if len(files_found) < 1:
+                        messages.add_message(
+                            request,
+                            messages.WARNING,
+                            mark_safe(
+                                _(
+                                    "No files found. Expected: {} but received:<br>{}"
+                                ).format(
+                                    ", ".join(files_not_found),
+                                    ", ".join(
+                                        [
+                                            f"<b>{escape(file)}</b>"
+                                            for file in files_received
+                                        ]
+                                    ),
+                                ),
+                            ),
+                        )
+                    else:
+                        messages.add_message(
+                            request,
+                            messages.SUCCESS,
+                            _("ZaakType successfully imported"),
+                        )
                     return HttpResponseRedirect(
                         reverse("admin:catalogi_catalogus_changelist")
                     )
--- a/src/openzaak/components/catalogi/admin/admin_views.py
+++ b/src/openzaak/components/catalogi/admin/admin_views.py
@@ -9,8 +9,7 @@ from django.db.utils import IntegrityError
 from django.http.response import HttpResponseRedirect
 from django.template.response import TemplateResponse
 from django.urls import reverse
-from django.utils.html import escape
-from django.utils.safestring import mark_safe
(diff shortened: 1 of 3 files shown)


src/openzaak/components/catalogi/admin/admin_views.py
```diff
@@ -64,13 +66,37 @@ def post(self, request, *args, **kwargs):
...
+                                    ", ".join(files_not_found),
+                                    ", ".join(
+                                        [
+                                            f"<b>{escape(file)}</b>"
+                                            for file in files_received
+                                        ]
+                                    ),
+                                ),
+                            ),
+                        )
```
Coperh, line 93: Shows a warning if the importer finds no file.

How should it be formatted?

annashamray, line 93: Instead of `str.format` and `mark_safe` you can use `format_html`
I'm also not sure that `str.format` and `mark_safe` work with lazy translations
```suggestion
                        msg = _("No files found. Expected: {files_not_found} but received: {files_received}")
                        msg_dict = {"files_found": ", ".join(files_not_found), "files_received": ". ".join(files_received)}
                        messages.add_message(request, messages.WARNING, format_html(msg, **msg_dict))

```

annashamray: Please take a look on the comment below

src/openzaak/components/catalogi/admin/utils.py
```diff
@@ -220,6 +226,10 @@ def import_zaaktype_for_catalogus(
                                 "A validation error occurred while deserializing a {}\n{}"
                             ).format(resource, deserialized.errors)
                         )
+            else:
+                files_not_found.append(f"{resource}.json")
+
+    return files_found, files_not_found, files_received
```
annashamray, line 232: Please add a typehint and perhaps a docstring to describe what is being returned in the function

src/openzaak/components/catalogi/admin/utils.py
```diff
@@ -174,7 +174,12 @@ def import_zaaktype_for_catalogus(
 
     uuid_mapping = {}
 
+    files_not_found = []
+    files_found = []
+
     with zipfile.ZipFile(import_file, "r") as zip_file:
+
+        files_received = zip_file.namelist()
```
annashamray, line 182: Now you have both folder names and filenames here. I think it's better to include only files.

annashamray (changes requested):
I have minor remarks, but overall is good

src/openzaak/components/catalogi/admin/admin_views.py
Coperh: This is done.

src/openzaak/components/catalogi/admin/utils.py
Coperh: I  just removed the list returns and throw and catch an exception instead.

src/openzaak/components/catalogi/tests/admin/test_import_export_zaaktype.py
```diff
@@ -1379,6 +1380,65 @@ def test_import_iotype_without_omschrijving_generiek(self, *mocks):
...
+        url = reverse("admin:catalogi_catalogus_import_zaaktype", args=(catalogus.pk,))
+        response = self.app.get(url)
+
+        form = response.form
+        f = io.BytesIO()
+        f.name = "test.zip"
+        f.seek(0)
+
+        with zipfile.ZipFile(f, "w") as zip_file:
+            # zip_file.mkdir("some_dir")
```
Coperh, line 1401: Added in python 3.11.

Not sure how to make directories in its current state. I don't think ZipFile was designed for this TBH.

src/openzaak/components/catalogi/admin/utils.py
```diff
@@ -219,6 +226,19 @@ def import_zaaktype_for_catalogus(
...
+    if len(files_found) < 1:
+        msg = _(
+            "No files found. Expected: {files_not_found} but received:<br> {files_received}"
+        )
+        msg_dict = {
+            "files_not_found": ", ".join(files_not_found),
+            "files_received": ", ".join(files_received),
+        }
+
+        raise ValueError(format_html(msg, **msg_dict))
```
annashamray, line 241: I think `CommandError` is a better fit here, because you don't need to write a separate handler for it. Now you don't catch ValueError in `CatalogusZaakTypeImportSelectView`. Is it not possible to be raised there?

Coperh: I replaced `ValueError` with `CommandError`.

It can be raised in `CatalogusZaakTypeImportSelectView` but it would be weird for `InformatieObjectType.json` or `BesluitType.json` are in the correct directory but the other files are not.

Coperh: Also based on the second comment, it is possible that some of the files are in the base zip directory and some files are in a subdirectory, which will not throw the error. Probably will still fail, but how likely is this to happen?

Coperh: I have tested with the `CatalogusZaakTypeImportSelectView` and it works as expected

src/openzaak/components/catalogi/tests/admin/test_import_export_zaaktype.py
```diff
@@ -1439,6 +1438,96 @@ def test_import_zaaktype_with_bad_filenames(self, *mocks):
...
+        fin.name = "test.zip"
+        fin.seek(0)
+
+        fout = io.BytesIO()
+        fout.name = "test.zip"
+        fout.seek(0)
+
+        # create zip with bad structure
+        with zipfile.ZipFile(fin, "r") as zin:
+            with zipfile.ZipFile(fout, "w") as zout:
```
Coperh, line 1492: 🧂

annashamray (approved):

Coperh:
Some annoying merges needed to be done.

PR #1535, opened 2023-12-20
:sparkles:[#1533] Auto-publish zaaktype related objects

### LINKED ISSUE #1533: [FEATURE-REQUEST] auto-publish zaaktype including besluittypes/informatieobjecttypes

Organisatie / Organization: DH
Taiga 537

Currently when publishing a zaaktype, all the related entities already need to be published. This requires additional unnecessary work because you have to publish these entities one by one.

The idea is to have an option in the importer to auto-publish all relations below the zaaktype when you publish a zaaktype.

This to avoid the need to investigate which informatieobjecttypes still need publishing before it's possible to publish the zaaktype…

PA: "De keuzemogelijkheid om een geïmporteerd zaaktype direct te publiceren, inclusief onderliggende nog niet gepubliceerde relaties"

Coperh:
Fixes #1533 

**Changes**

Auto-publish zaaktype related objects

### DIFF
--- a/src/openzaak/components/catalogi/admin/admin_views.py
+++ b/src/openzaak/components/catalogi/admin/admin_views.py
@@ -9,11 +9,11 @@ from django.http.response import HttpResponseRedirect
 from django.template.response import TemplateResponse
 from django.urls import reverse
 from django.utils.translation import ugettext_lazy as _
-from django.views.generic import FormView, TemplateView
+from django.views.generic import DetailView, FormView, TemplateView
 
 from openzaak.utils.admin import AdminContextMixin
 
-from ..models import BesluitType, Catalogus, InformatieObjectType
+from ..models import BesluitType, Catalogus, InformatieObjectType, ZaakType
 from .forms import BesluitTypeFormSet, InformatieObjectTypeFormSet, ZaakTypeImportForm
 from .utils import (
     construct_besluittypen,
@@ -209,3 +209,72 @@ class CatalogusZaakTypeImportSelectView(
             messages.add_message(request, messages.ERROR, exc)
 
         return TemplateResponse(request, self.template_name, context)
+
+
+class ZaaktypePublishView(AdminContextMixin, PermissionRequiredMixin, DetailView):
+    template_name = "admin/catalogi/publish_zaaktype.html"
+    permission_required = "catalogi.change_zaaktype"
+
+    queryset = ZaakType.objects.all()
+
+    def post(self, request, *args, **kwargs):
+
+        self.object = self.get_object()
+
+        if self.object.concept:
+            errors = []
+
+            if "_auto-publish" in request.POST:
+
+                published_besluittypen = []
+                published_informatieobjecttypen = []
+                # publish related types
+                for besluittype in self.object.besluittypen.filter(concept=True):
+                    besluittype.publish()
+                    published_besluittypen.append(besluittype.omschrijving)
+
+                for iot in self.object.informatieobjecttypen.filter(concept=True):
+                    iot.publish()
+                    published_informatieobjecttypen.append(iot.omschrijving)
+
+                if len(published_besluittypen) > 0:
+                    messages.add_message(
+                        request,
+                        messages.INFO,
+                        _("Auto-published related besluittypen: {besluittypen}").format(
+                            besluittypen=", ".join(published_besluittypen)
+                        ),
+                        "autopublish",
+                    )
+                if len(published_informatieobjecttypen) > 0:
+                    messages.add_message(
+                        request,
+                        messages.INFO,
+                        _(
+                            "Auto-published related informatieobjecttypen: {iots}"
+                        ).format(iots=", ".join(published_informatieobjecttypen)),
+                        "autopublish",
+                    )
+
+            if (
+                self.object.besluittypen.filter(concept=True).exists()
+                or self.object.informatieobjecttypen.filter(concept=True).exists()
+            ):
+                errors.append(_("All related resources should be published"))
+
+            if len(errors) > 0:
+                messages.add_message(
+                    request, messages.ERROR, ". ".join([str(e) for e in errors])
+                )
+                context = self.get_context_data()
+                return self.render_to_response(context)
+
+            self.object.publish()
+        else:
+            messages.add_message(
+                request, messages.WARNING, _("Zaaktype object is already published")
+            )
+
+        return HttpResponseRedirect(
+            reverse("admin:catalogi_zaaktype_change", args=(self.object.pk,))
+        )
--- a/src/openzaak/components/catalogi/admin/admin_views.py
+++ b/src/openzaak/components/catalogi/admin/admin_views.py
@@ -13,6 +13,7 @@ from django.views.generic import DetailView, FormView, TemplateView
 
 from openzaak.utils.admin import AdminContextMixin
 
+from ..api.viewsets import ZaakTypeViewSet
 from ..models import BesluitType, Catalogus, InformatieObjectType, ZaakType
 from .forms import BesluitTypeFormSet, InformatieObjectTypeFormSet, ZaakTypeImportForm
 from .utils import (
@@ -278,6 +279,8 @@ class ZaaktypePublishView(AdminContextMixin, PermissionRequiredMixin, DetailView
                 messages.SUCCESS,
                 _("The resource has been published successfully!"),
             )
+            self.send_notification(request)
+
         else:
             messages.add_message(
                 request, messages.WARNING, _("Zaaktype object is already published")
@@ -286,3 +289,22 @@ class ZaaktypePublishView(AdminContextMixin, PermissionRequiredMixin, DetailView
         return HttpResponseRedirect(
             reverse("admin:catalogi_zaaktype_change", args=(self.object.pk,))
         )
+
+    def send_notification(self, context_request):
+
+        viewset = ZaakTypeViewSet()
+        viewset.action = "update"
+
+        reference_object = self.object
+
+        # set versioning to context_request
+        (
+            context_request.version,
+            context_request.versioning_scheme,
+        ) = viewset.determine_version(context_request)
+
+        data = viewset.serializer_class(
+            reference_object, context={"request": context_request}
+        ).data
+
+        viewset.notify(status_code=200, data=data, instance=reference_object)
--- a/src/openzaak/components/catalogi/admin/admin_views.py
+++ b/src/openzaak/components/catalogi/admin/admin_views.py
@@ -215,7 +215,10 @@ class ZaaktypePublishView(AdminContextMixin, PermissionRequiredMixin, DetailView
     template_name = "admin/catalogi/publish_zaaktype.html"
     permission_required = "catalogi.change_zaaktype"
 
-    queryset = ZaakType.objects.all()
+    queryset = ZaakType.objects.all().prefetch_related(
+        "besluittypen", "informatieobjecttypen"
+    )
+    pk_url_kwarg = "zaaktype_pk"
 
     def post(self, request, *args, **kwargs):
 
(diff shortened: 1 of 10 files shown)


src/openzaak/components/catalogi/admin/zaaktypen.py
```diff
@@ -307,6 +307,16 @@ def get_related_objects(self, obj):
 
         return resource_list, id_list
 
+    def response_post_save_change(self, request, obj):
+        if "_publish" in request.POST:
+            # publish related types
+            for besluittype in obj.besluittypen.filter(concept=True):
+                besluittype.publish()
```
alextreme, line 314: Add a message to state which objects have been auto-published, this to avoid surprises

annashamray (commented):
This change makes publishing behavior in the admin different from the API
I think @joeribekker should approve it before it can be proceeded 

I also think that this behavior should be described somewhere, so the user will expect it

annashamray:
Discussed with @joeribekker . As specified in US let's add a checkbox which explicitly allows to auto-publish zaaktype relations. This will solve the issue
@Coperh feel free to continue working on this PR

src/openzaak/components/catalogi/admin/zaaktypen.py
```diff
@@ -307,6 +307,23 @@ def get_related_objects(self, obj):
...
+
+            for iot in obj.informatieobjecttypen.filter(concept=True):
+                iot.publish()
+            self.message_user(
+                request,
+                _("Published related zaaktype objects"),
+                messages.INFO,
+                "autopublish",
+            )
```
Coperh, line 323: This does nothing and I do not know how to make it do something.

Coperh:
Added a checkbox to the admin page and now only auto-publishes if it is selected.

src/openzaak/components/catalogi/admin/zaaktypen.py
Coperh: It is cleared here: https://github.com/open-zaak/open-zaak/blob/e08dbbcf405c88322806764f60bf4643d52aea8f/src/openzaak/components/catalogi/admin/mixins.py#L62-L65

I'm not sure if I change it to allow the message without affecting other things.

Maybe add it after the super function is run.

alextreme: Seems like a hack, check with @stevenbal about this: https://github.com/open-zaak/open-zaak/commit/414ddafacff0f507de3d2f01baa96da7b9a49c36

alextreme: Better would be to also mention the names of the related entities which have been auto-published

stevenbal: The clearing of the messages is a bit of a hack indeed, it was to prevent the default admin save messages from showing up. I couldn't find another way to suppress these default 

Adding the message after `super().response_post_save_change` has run would make it work, though if I remember correctly the actual publishing of the related resources must happen before the `ZaakType` is published.

Coperh: It was originally an aesthetic thing, but if I add the name of all the related published items, it might make more sense to display the published types bellow.

The question is: is it better to show the confirmation first then the auto-published items or the other way around.
For the former, I can just move the `self.message_user`.

Coperh: I just displayed the auto-published messages bellow the confirmation.

annashamray (changes requested):
image

I don't think the checkbox looks good in the middle of button group
Should it be moved above? Or should it be converted into the separate button? Or maybe create an intermediate page for publishing objects with the form and this checkbox? 

@alextreme @joeribekker what do you think?

alextreme:
> image
> 
> I don't think the checkbox looks good in the middle of button group Should it be moved above? Or should it be converted into the separate button? Or maybe create an intermediate page for publishing objects with the form and this checkbox?
> 
> @alextreme @joeribekker what do you think?

Adding an intermediate formview makes more sense imho

Coperh:
New publish zaaktype view added. Bypasses most of the `PublishAdminMixin` in zaaktype admin. `BesluitTypeAdmin` and `InformatieObjectTypeAdmin` remain the same.

Also moved is_published to the model mixin rather than the admin mixin.

src/openzaak/components/catalogi/admin/mixins.py
```diff
@@ -77,14 +77,14 @@ def response_post_save_change(self, request, obj):
         else:
             return super().response_post_save_change(request, obj)
 
-    def is_published(self, obj):
+    def _is_published(self, obj):
```
annashamray, line 80: this function looks unnecessary. If you want it just to add description and boolean flag to the field, you can use `@admin.display` decorator for model method in the `ConceptMixin` model mixin

src/openzaak/templates/admin/catalogi/publish_zaaktype.html
```diff
@@ -0,0 +1,58 @@
+{% extends "admin/index.html" %}
+{% comment %} SPDX-License-Identifier: EUPL-1.2 {% endcomment %}
+{% comment %} Copyright (C) 2019 - 2024 Dimpact {% endcomment %}
+{% load i18n %}
+
+{% block title %} {% trans "Import zaaktype" %} {{ block.super }} {% endblock %}
```
annashamray, line 6:
```suggestion
{% block title %} {% trans "Publish zaaktype" %} {{ block.super }} {% endblock %}
```

src/openzaak/templates/admin/catalogi/publish_zaaktype.html
```diff
@@ -0,0 +1,58 @@
...
+
+
+
+    <div id="content-main">
+        <form action="" method="POST" enctype="multipart/form-data">
+            <div>
+                {% csrf_token %}
+
+                <label for="_auto-publish">{% translate "Auto-publish related objects" %}</label>
+                <input type="checkbox" name="_auto-publish" id="_auto-publish" {% if zaaktype.concept == False %}disabled="disabled"{% endif %}>
```
annashamray, line 51: For better layout 

```suggestion
<fieldset class="module aligned ">
                {% csrf_token %}

                <label for="_auto-publish">{% translate "Auto-publish related objects" %}</label>
                <input type="checkbox" name="_auto-publish" id="_auto-publish" {% if zaaktype.concept == False %}disabled="disabled"{% endif %}>
</fieldset>
```

src/openzaak/components/catalogi/admin/admin_views.py
```diff
@@ -209,3 +209,72 @@ def post(self, request, *args, **kwargs):
...
+                errors.append(_("All related resources should be published"))
+
+            if len(errors) > 0:
+                messages.add_message(
+                    request, messages.ERROR, ". ".join([str(e) for e in errors])
+                )
+                context = self.get_context_data()
+                return self.render_to_response(context)
+
+            self.object.publish()
```
annashamray, line 272: please add message that the zaaktype is successfully published

src/openzaak/templates/admin/catalogi/publish_zaaktype.html
```diff
@@ -0,0 +1,58 @@
...
+{% block content %}
+    <h1>{% trans 'Publiceren zaaktype' %}</h1>
+
+    <h2>{{ zaaktype }}</h2>
+
+{% if zaaktype.informatieobjecttypen.all %}
+    <table class="module aligned ">
+        <tr><th>{% trans 'Informatieobjecttypen' %}</th> <th>{% trans 'Gepubliceerd' %}</th></tr>
+        {% for iot in zaaktype.informatieobjecttypen.all %}
+        <tr><td>{{iot}}</td> <td>{{iot.is_published | yesno}}</td></tr>
```
annashamray, line 28: Should it be displayed in the same way like in list view? Red cross or green checkmark? 
This comment is not critical. If it's too much work you can skip it

src/openzaak/components/catalogi/admin/zaaktypen.py
```diff
@@ -307,6 +323,15 @@ def get_related_objects(self, obj):
 
         return resource_list, id_list
 
+    def response_post_save_change(self, request, obj):
+
+        if "_publish" in request.POST:
```
annashamray, line 328: I think it should be a link, not a submit button now. And it should be excluded from submit process. Because when you click on this button you are redirected on the publish page with this message: 
image

src/openzaak/components/catalogi/admin/zaaktypen.py
```diff
@@ -269,6 +271,20 @@ class ZaakTypeAdmin(
     # For export mixin
     resource_name = "zaaktype"
 
+    def get_urls(self):
+        urls = super().get_urls()
+
+        my_urls = [
+            path(
+                "<int:pk>/publish/",
```
annashamray, line 279:
```suggestion
                "<path:zaaktype_pk>/publish/",
```

src/openzaak/components/catalogi/admin/admin_views.py
```diff
@@ -209,3 +209,72 @@ def post(self, request, *args, **kwargs):
             messages.add_message(request, messages.ERROR, exc)
 
         return TemplateResponse(request, self.template_name, context)
+
+
+class ZaaktypePublishView(AdminContextMixin, PermissionRequiredMixin, DetailView):
+    template_name = "admin/catalogi/publish_zaaktype.html"
+    permission_required = "catalogi.change_zaaktype"
+
+    queryset = ZaakType.objects.all()
```
annashamray, line 218: Please add `prefetch_related` here. There are loops through elated objects in the template, which could produce a huge pile of DB requests

annashamray (changes requested):
I like refactoring and a separate view with the list of related objects. 
It looks much more convenient!

src/openzaak/components/catalogi/admin/mixins.py
Coperh: It seems you cannot set the boolean flag in the model, it must be set in the admin.

src/openzaak/components/catalogi/admin/zaaktypen.py
```diff
@@ -307,6 +323,15 @@ def get_related_objects(self, obj):
 
         return resource_list, id_list
 
+    def response_post_save_change(self, request, obj):
```
annashamray, line 326: I think this method is not needed anymore

src/openzaak/components/catalogi/admin/mixins.py
annashamray: I think you can set it in the model for the method, but not for the property

Coperh: Can confirm it does work.

src/openzaak/components/catalogi/admin/zaaktypen.py
Coperh: Removed

src/openzaak/components/catalogi/tests/admin/test_admin_publish.py
```diff
@@ -99,10 +99,11 @@ def test_publish_zaaktype(self, m, mock_notif):
...
-        form = response.forms["zaaktype_form"]
+        url = reverse("admin:catalogi_zaaktype_publish", args=(zaaktype.pk,))
+        publish_page = self.app.get(url)
 
         with self.captureOnCommitCallbacks(execute=True):
-            response = form.submit("_publish").follow()
+            response = publish_page.form.submit("_publish").follow()
 
         zaaktype.refresh_from_db()
         self.assertFalse(zaaktype.concept)
```
Coperh, line 109: The notification bellow now fails to be called and I cant find why. @annashamray do you know?

Also why cant I comment on unchanged lines?

alextreme: Because Github isn't Bitbucket, not being able to comment on unchanged lines in a PR is one of my (few) annoyances with Github

Coperh: This actually seems to be from the SideEffectsMixin

Coperh:
Suggestions implemented. Added notifications to the publish view.

annashamray (approved):
Looks good!

PR #1538, opened 2023-12-27
:sparkles:[#1531] Add ability to import zaaktype under different iden…

### LINKED ISSUE #1531: [FEATURE-REQUEST] ability to import a zaaktype under a different identification

Organisatie / Organization: DH
Taiga 535

Currently it isn't possible to import a zaaktype via the admin of Open Zaak if the identification already exists. 

Add the ability to import a zaaktype with a different zaaktype.identificatie, allowing the admin user to import an existing (duplicate) zaaktype with a different identification.

Coperh:
Fixes #1531 

Adds option to enter zaaktype identification on import page. Uses identification from file if not entered.

### DIFF
--- a/src/openzaak/components/catalogi/admin/forms.py
+++ b/src/openzaak/components/catalogi/admin/forms.py
@@ -487,10 +487,10 @@ class CatalogusImportForm(forms.Form):
 
 class ZaakTypeImportForm(forms.Form):
 
-    identificatie = forms.CharField(
-        label=_("identificatie"),
+    identificatie_prefix = forms.CharField(
+        label=_("identificatie prefix"),
         required=False,
-        help_text=_("Zaaktype name. Leave blank to use imported."),
+        help_text=_("Zaaktype identification prefix. Leave blank to use imported."),
         validators=[alphanumeric_excluding_diacritic,],
     )
 
--- a/src/openzaak/components/catalogi/admin/forms.py
+++ b/src/openzaak/components/catalogi/admin/forms.py
@@ -18,6 +18,7 @@ from vng_api_common.constants import (
     BrondatumArchiefprocedureAfleidingswijze as Afleidingswijze,
 )
 from vng_api_common.tests import reverse as _reverse
+from vng_api_common.validators import alphanumeric_excluding_diacritic
 from zds_client import ClientError
 from zgw_consumers.models import Service
 
@@ -485,6 +486,14 @@ class CatalogusImportForm(forms.Form):
 
 
 class ZaakTypeImportForm(forms.Form):
+
+    identificatie = forms.CharField(
+        label=_("identificatie"),
+        required=False,
+        help_text=_("Zaaktype name. Leave blank to use imported."),
+        validators=[alphanumeric_excluding_diacritic,],
+    )
+
     file = forms.FileField(
         label=_("bestand"),
         required=True,
--- a/src/openzaak/components/catalogi/admin/utils.py
+++ b/src/openzaak/components/catalogi/admin/utils.py
@@ -164,7 +164,7 @@ def construct_besluittypen(
 
 @requests_cache_enabled("import", backend=DjangoRequestsCache())
 def import_zaaktype_for_catalogus(
-    identificatie,
+    identificatie_prefix,
     catalogus_pk,
     import_file_content,
     iotypen_uuid_mapping,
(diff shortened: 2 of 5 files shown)


src/openzaak/components/catalogi/admin/forms.py
```diff
@@ -485,6 +486,14 @@ class CatalogusImportForm(forms.Form):
 
 
 class ZaakTypeImportForm(forms.Form):
+
+    identificatie = forms.CharField(
+        label=_("identificatie"),
+        required=False,
+        help_text=_("Zaaktype name. Leave blank to use imported."),
```
Coperh, line 493: Translation needed here.

annashamray, line 493:
```suggestion
        help_text=_("Zaaktype identification. Leave blank to use imported."),
```

It's not a name but rather a unique id number

annashamray (changes requested):
This implementation works only if only one zaaktype being imported. If there are two or more zaaktypen 400 error appears because all of the zaaktypen are assigned the same identificatie

If you want to add `identificatie` field in the same page when you upload the file, then it should be rather some prefix or postfix to append for existng zaaktype identificaties

If you want to add manually the identificatie for every zaaktype, then it would be a big change in the process. Perhaps you can add a formset into the `<path:catalogus_pk>/zaaktype/import/select` page

I think the easiest approach would be to add a checkbox - `generate new identificatie` like we do it with uuids during the catalogus import (page `admin/catalogi/catalogus/import/`). When we will remove the existing identificaties and generate them ourselves. The downside is that you won't know how to find the imported zaaktype, because you don't know their identificaties.

src/openzaak/components/catalogi/admin/forms.py
Coperh: Was supposed to change that but now it is done.

Coperh:
The prefix seems the easiest one to implement, so I did that. Might cause potential issues with the identification being too long.

src/openzaak/components/catalogi/admin/utils.py
```diff
@@ -202,6 +206,10 @@ def import_zaaktype_for_catalogus(
 
                 for entry in data:
                     if resource == "ZaakType":
+                        if identificatie_prefix:
+                            entry[
+                                "identificatie"
+                            ] = f"{identificatie_prefix}_{entry['identificatie']}"
```
annashamray, line 212: What happens if the length of the field exceeds max length (it's only 40)? Could you please create a unittest for it?

annashamray (changes requested):
See my comment about an extra unit test

src/openzaak/components/catalogi/admin/utils.py
Coperh: Apparently its actually 50. I've just truncated the string and added a unit test for this.

annashamray: Let's discuss it. For example you have 2 zaken to import with identificaties that take 30 char, for example "SOME-GEMEENTE--2023-0000000001" and "SOME-GEMEENTE--2023-0000000002". You set prefix which has 25 char length. 
Now the last part of identificaties would be truncates and now all zaken would have the same identificaties. So the import would fail with `overlap` error, which is not very clear for user how to fix

I think it's better to check if the length of the new prefixed identificatie is longer than max. And raise 400 with the message that the combined identificatie is too long.

alextreme:
Discussed with @Coperh , what is still necessary is to avoid truncating the identificatie if it's >50 characters, instead raise an error. After that Anna can review

src/openzaak/components/catalogi/admin/utils.py
```diff
@@ -201,6 +206,21 @@ def import_zaaktype_for_catalogus(
...
+
+                            if len(new_identification) > 50:
+                                raise ValidationError(
+                                    _(
+                                        "Identification {} is too long with prefix. Max 50 characters."
+                                    ).format(new_identification)
+                                )
+
+                            entry["identificatie"] = new_identification
+
```
Coperh, line 223: Now raises error.

Translation needed maybe

Coperh:
@annashamray Changes made, ready for re-review

annashamray (approved):

alextreme (approved):

PR #1562, opened 2024-01-31
csrf trusted origins envvar

### LINKED ISSUE #1561: As a devops/architect I want to expose Open Zaak on a different domain for the API than for the admin
Currently we have OPENZAAK_DOMAIN, which works for exposing the Open Zaak APIs and the admin via a different domainname than which is used internally.

However, accessing the admin from the internal domainname then no longer works. This can be corrected by exposing CSRF_TRUSTED_ORIGINS as an envvar

alextreme:
Fixes #1561 

Exposes CSRF_TRUSTED_ORIGINS via an envvar


alextreme:
Suggested change is from @sjoerdie

PR #1565, opened 2024-02-01
Release/1.11.0

annashamray:


joeribekker (approved):

joeribekker (commented):
I feel like the changelog does not reflect all the pain that went into this update and interpreting all the changes VNG made from 3 different resources but let's not bore people with this.

PR #1576, opened 2024-02-26
Feature/setup configuration test

annashamray:
Fixes https://github.com/maykinmedia/django-setup-configuration/issues/1 (integration test part)
Depends on https://github.com/open-zaak/open-zaak/pull/1572
Depends on https://github.com/open-zaak/open-notificaties/pull/131


annashamray:
It looks like it's not needed

PR #1588, opened 2024-03-22
 🐛 [#1585] Only rewrite host for API requests

### LINKED ISSUE #1585: `OPENZAAK_REWRITE_HOST` & `OPENZAAK_DOMAIN` breaks OIDC login
**General information**
Open Zaak version: 1.11.0

**Describe the bug**
When setting `OPENZAAK_REWRITE_HOST` and `OPENZAAK_DOMAIN` the OIDC redirect url is set to `OPENZAAK_DOMAIN` value, causing OIDC login to fail.

**To Reproduce**
1. Set `OPENZAAK_REWRITE_HOST=true`, `OPENZAAK_DOMAIN=api.gateway.example.com`
2. Configure your OIDC provider
3. Try to login with OIDC.

**Expected behavior**
From the change logs: https://github.com/open-zaak/open-zaak/blob/c52837c37912698b7f674d979c3c0617bc62e8e0/CHANGELOG.rst#L274

We expect the domain rewrite only to happen for API calls, this is what the docs imply. The redirect uri for OIDC should remain the domain openzaak is hosted on.

stevenbal:
Fixes #1585

**Changes**

* Only apply rewrite host middleware to API requests
* Specifically apply host rewrite for authorization admin notifications

### DIFF
--- a/src/openzaak/components/autorisaties/tests/admin/test_autorisaties_admin.py
+++ b/src/openzaak/components/autorisaties/tests/admin/test_autorisaties_admin.py
@@ -476,6 +476,57 @@ class ManageAutorisatiesAdmin(NotificationsConfigMixin, TestCase):
             InformatieObjectTypeFactory.create()
         self.assertEqual(self.applicatie.autorisaties.count(), 2)
 
+    @tag("notifications")
+    @override_settings(
+        NOTIFICATIONS_DISABLED=False,
+        OPENZAAK_DOMAIN="openzaak.example.com",
+        OPENZAAK_REWRITE_HOST=True,
+        ALLOWED_HOSTS=["testserver", "openzaak.example.com"],
+    )
+    def test_add_autorisatie_send_notificatie_with_rewrite_host(self):
+        data = {
+            # management form
+            "form-TOTAL_FORMS": 1,
+            "form-INITIAL_FORMS": 0,
+            "form-MIN_NUM_FORMS": 0,
+            "form-MAX_NUM_FORMS": 1000,
+            "form-0-component": ComponentTypes.brc,
+            "form-0-scopes": ["besluiten.lezen"],
+            "form-0-related_type_selection": RelatedTypeSelectionMethods.all_current_and_future,
+        }
+
+        with self.captureOnCommitCallbacks(execute=True):
+            response = self.client.post(self.url, data)
+
+        self.assertEqual(response.status_code, 302)
+        self.assertFalse(Autorisatie.objects.exists())
+
+        with patch(
+            "notifications_api_common.viewsets.send_notification.delay"
+        ) as mock_notif:
+            # create a besluittype - this should trigger a new autorisatie being installed
+            # We need to capture on_commit callbacks twice, because the `AutorisatieSpec.sync`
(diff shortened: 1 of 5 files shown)


src/openzaak/components/autorisaties/tests/admin/test_autorisaties_admin.py
```diff
@@ -548,52 +548,6 @@ def test_changes_send_notifications(self, mock_notif):
...
         )
 
-    @override_settings(
-        NOTIFICATIONS_DISABLED=False,
-        OPENZAAK_DOMAIN="openzaak.example.com",
-        OPENZAAK_REWRITE_HOST=True,
-        ALLOWED_HOSTS=["testserver", "openzaak.example.com"],
-    )
-    @patch("notifications_api_common.viewsets.send_notification.delay")
-    def test_changes_send_notifications_with_openzaak_domain_setting(self, mock_notif):
```
annashamray, line 558: Please restore this test. I think it catches a bug here

annashamray (changes requested):
To be honest I don't like that OZ admin and API will behave differently, I think it could create a lot of errors for Authorizations, like that failing test that was removed. 

In the Autorisaties admin page we convert local model instances of Zaaktype, Besluittype and Informatieobjecttype into absolute urls and save them like urls in the DB. So the different domains could result in errors.

I can think of some ways to solve it:
* allow rewriting of domain for OZ components in the admin
* use OPENZAAK_DOMAIN to build urls in Autorisaties admin form, but it seems more like a band-aid than a fix, because it doesn't prevent future errors
* follow Sergei's advice and use `X-Forwarded-Host` header and see if it helps

To be honest I don't know what to choose, we can discuss it with the team
@joeribekker What do you think?

src/openzaak/components/autorisaties/tests/admin/test_autorisaties_admin.py
stevenbal: I fixed the behavior for the authorizations admin notifications and restored the test

joeribekker:
@annashamray lets discuss with Steven, you and me.

src/openzaak/components/autorisaties/utils.py
```diff
@@ -97,10 +97,14 @@ def send_applicatie_changed_notification(
...
     from openzaak.utils import build_fake_request
+    from openzaak.utils.middleware import override_request_host
 
     viewset = ApplicatieViewSet()
     viewset.action = "update"
     if new_version is None:
         request = build_fake_request(method="put")
+        # we have to explicitly override the request host (because this is not an API request)
+        # to ensure the notification has the right URLs
+        override_request_host(request)
```
annashamray, line 108: Is it necessary? I think `build_fake_request` should already use `OPENZAAK_DOMAIN`
Maybe it could be also refactored with your `override_request_host` function

annashamray:
@joeribekker sure! I've made comments in the US. But let's discuss it !

src/openzaak/components/autorisaties/utils.py
stevenbal: You're right, it's not needed so I removed it

annashamray (approved):
Good work!

PR #1601, opened 2024-04-02
:sparkles:[#1584] fix autorisaties 500 error when related object deleted

### LINKED ISSUE #1584: When specifying a Zaken API authorization with permissions linked to an internal zaaktype, deleting the zaaktype causes a 500 error
**General information**
Open Zaak version: 1.11
Rotterdam nr 80
Sentry 352159

**Describe the bug**
When specifying a Zaken API authorization with permissions linked to an internal zaaktype (eg. you check the zaaktypes individually using the 'Selecteer handmatig' option), deleting the zaaktype that is linked to the autorization causes a 500 error.

ZaakType.DoesNotExist
ZaakType matching query does not exist.
openzaak/components/autorisaties/utils.py in _get_related_object at line 30

A workaround is to delete the corresponding Autorisatie (but note that there isn't a FK so the zaaktype delete isn't cascaded to the authorizations)

I'd suggest to ignore autorizations to entities which no longer exist, if only to allow managing the autorizations. Normally zaaktypes aren't deleted, however this is something that can be done via the admin and was encountered during testing

Coperh:
Fixes #1584

### DIFF
--- a/src/openzaak/components/autorisaties/tests/admin/test_autorisaties_admin.py
+++ b/src/openzaak/components/autorisaties/tests/admin/test_autorisaties_admin.py
@@ -1009,3 +1009,22 @@ class ManageAutorisatiesAdmin(NotificationsConfigMixin, TestCase):
         self.assertEqual(response.status_code, 200)
         initial_data = response.context["formdata"]
         self.assertEqual(len(initial_data), 2)  # 1 empty form, 1 unrelated auth spec
+
+    @tag("gh-1584")
+    def test_related_object_does_not_exist(self):
+        """
+        Assert that the autorisaties view does not crash
+        if an autorisatie fails to delete after its related object is deleted.
+        """
+
+        AutorisatieFactory.create(
+            applicatie=self.applicatie,
+            component=ComponentTypes.drc,
+            scopes=["documenten.lezen"],
+            max_vertrouwelijkheidaanduiding=VertrouwelijkheidsAanduiding.openbaar,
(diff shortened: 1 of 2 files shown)


Coperh:
@annashamray The thing causing this seems to be deleting based on the URL instead of the UUID. 
The autoraties does not use the same URL as the related objects, defaulting to localhost for me locally.

It seems to deleted in three different ways:

- https://github.com/open-zaak/open-zaak/blob/main/src/openzaak/components/catalogi/signals.py
- https://github.com/open-zaak/open-zaak/blob/main/src/openzaak/components/autorisaties/management/commands/sync_autorisaties.py (this one will delete correctly)
- https://github.com/open-zaak/open-zaak/blob/9bcdb75764b2d50b880384f42ac78b202e8429fd/src/openzaak/components/autorisaties/models.py#L74

I wonder if it is worth changing them to work the same way.

annashamray:
@Coperh  The management command was created when we had a synchronization bug, but it's was for a one-time fix. 
The main problem is why the signal doesn't work properly for Rotterdam. 

>The autorisaties does not use the same URL as the related objects, defaulting to localhost for me locally.

Then Zaaktype is being deleted its url is build based on `OPENZAAK_DOMAIN` or `Sites` config.
If the signal doesn't do anything, it could mean that  `OPENZAAK_DOMAIN` or `Sites` aren't configured properly. I think it should be investigated further. 


I don't mind this workaround, but it doesn't resolve this issue. I've asked Alex to reproduce the bug and provide logs in the US

annashamray (approved):

PR #1603, opened 2024-04-02
:sparkles:[#1577] Fix RelativeDeltaField allowing empty required values

### LINKED ISSUE #1577: Can save empty Doorlooptijd behandeling when field is required in Zaaktype Admin
Ref: DH552

**Description**
Can a save an empty doorlooptijd value in the zaaktype admin while it is required in when importing.

Empty Doorlooptijd field defaults to a value of `P0D` value which is technically not empty.

Coperh:
Fixes #1577

### DIFF
--- a/src/openzaak/forms/fields.py
+++ b/src/openzaak/forms/fields.py
@@ -16,3 +16,8 @@ class RelativeDeltaField(RelativeDeltaFormField):
         assert "empty_value" not in kwargs, "empty_value may not be provided"
         kwargs["empty_value"] = None
         super().__init__(*args, **kwargs)
+
+    def clean(self, value):
+        self.validate(value)  # must be first to check for empty value
+        value = super().clean(value)
+        return value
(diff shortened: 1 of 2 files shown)


src/openzaak/forms/fields.py
```diff
@@ -16,3 +16,8 @@ def __init__(self, *args, **kwargs):
         assert "empty_value" not in kwargs, "empty_value may not be provided"
         kwargs["empty_value"] = None
         super().__init__(*args, **kwargs)
+
+    def clean(self, value):
+        self.validate(value)
+        value = super().clean(value)
```
annashamray, line 22:
```suggestion
        value = super().clean(value)
        self.validate(value)
```
I think self.validate expects the `value` which was already prepared, so it should be called after
In this particular case we don't have validators to the field, so it doesn't matter, but for the future let's switch the lines

annashamray (approved):

src/openzaak/forms/fields.py
Coperh: `self.validate(value)` must be first to check that the field is not empty. `RelativeDeltaFormField` does not use the validate function by default.

PR #1605, opened 2024-04-04
resolve local and external urls with the same host

### LINKED ISSUE #1592: OPENZAAK_REWRITE_HOST & OPENZAAK_DOMAIN breaks external services, path isn't taken into account when determining the service
**General information**
Open Zaak version: 1.11

**Describe the bug**
Taiga AMS 13

When using OPENZAAK_DOMAIN with an API gateway (for instance api.amsterdam.nl with a subpath /openzaak), Open Zaak acts as if it has this domain.

This however breaks the determining if a URL is external or internal, leading to external Document APIs routed through the same hostname (api.amsterdam.nl/alfresco) being seen as local URLs instead of external URLs. As such, creating a zaakinformatieobject with a external alfresco URL is handled as a local URL and causes an error that the object can't be found.

https://github.com/maykinmedia/django-loose-fk/blob/84facdbae45be468aa7f0d4b2def6e6593e4d72e/django_loose_fk/drf.py#L48

`
>>> import logging
>>> from dataclasses import dataclass
>>> from typing import Tuple, Union
>>> from urllib.parse import ParseResult, urlparse
>>> s = "https://api.amsterdam.nl/ozgv-t/documenten/service/drc/v1/enkelvoudiginformatieobjecten/foo"
>>> urlparse(s)
ParseResult(scheme='https', netloc='api.amsterdam.nl', path='/alfresco/documenten/service/drc/v1/enkelvoudiginformatieobjecten/foo', params='', query='', fragment='')
>>> parsed = urlparse(s)
>>> parsed.netloc
'api.amsterdam.nl'
`

This is then the same hostname as OPENZAAK_DOMAIN, causing is_local to be True

annashamray:
Fixes #1592 
Depends on https://github.com/maykinmedia/django-loose-fk/pull/29

### DIFF
--- a/docs/installation/config/env_config.md
+++ b/docs/installation/config/env_config.md
@@ -172,6 +172,13 @@ on Docker, since `localhost` is contained within the container:
 * `NOTIFICATIONS_DISABLED`: if this variable is set to `true`, `yes` or `1`, the notification mechanism will be
     disabled. Defaults to `False`.
 
+* `LOOSE_FK_LOCAL_BASE_URLS`: explicitly list the allowed prefixes of local urls.
+  Defaults to an empty list. This setting can be used to separate local and external urls, when
+  Open Zaak and other services are deployed within the same domain or API Gateway.
+  If this setting is not defined, all urls with the same host as in the request are considered local. 
+  Example:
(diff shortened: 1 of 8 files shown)


alextreme (approved):
I'd prefer to have https://github.com/maykinmedia/django-loose-fk/pull/29 as a release on pypi first, but whatever works for you guys

stevenbal (approved):

annashamray:
@alextreme , sure, this PR was created to make sure, that the changes in `django-loose-fk` actually fix the regression test and don't crash anything else

PR #1615, opened 2024-04-09
:sparkles: [#1569] added experimental query filters

### LINKED ISSUE #1569: Add filter and sorting params to Documenten API enkelvoudiginformatieobjecten endpoint
See my comment here: https://github.com/generiekzaakafhandelcomponent/roadmap-items/issues/24

bart-maykin:
Fixes #1569

**Changes**

Added Experimental query params:
- auteur
- beschrijving
- aanmaakdatum
- titel
- vertrouwelijkheidaanduiding
- lock
- zaak
- ordering
  - auteur
  - bestandsomvang
  - bestandstype
  - aanmaakdatum
  - status
  - title
  - vertrouwelijkheidaanduiding


src/openzaak/components/documenten/tests/test_filters.py
```diff
@@ -1,18 +1,82 @@
...
+            )
+            self.assertEqual(response.status_code, status.HTTP_200_OK)
+            data = response.json()
+            self.assertEqual(data["count"], 1)
+            self.assertEqual(data["results"][0]["locked"], False)
+
+    @override_settings(
+        NOTIFICATIONS_DISABLED=True, ALLOWED_HOSTS=["testserver", "testserver.com"],
+    )
+    def test_zaak_filter(self):
```
stevenbal, line 57: Since the zaak filter is a `NestedFkOrUrlFieldFilter`, could you also add a test to check if filtering works if the zaak is external?

src/openzaak/utils/filterset.py
```diff
@@ -15,3 +21,67 @@ def filter_for_field(cls, field, field_name, lookup_expr=None):
         if not filter.extra.get("help_text"):
             filter.extra["help_text"] = getattr(field, "help_text", None)
         return filter
+
+
+# Updated FkOrUrlFieldFilter for nested model field filtering
+class NestedFkOrUrlFieldFilter(FkOrUrlFieldFilter):
```
stevenbal, line 27: This seems like a lot of copied code from `FkOrUrlFieldFilter`, is this necessary? If needed, we could also modify the `FkOrUrlFieldFilter` from the library to support nested fields

stevenbal (changes requested):

src/openzaak/components/documenten/api/filters.py
```diff
@@ -31,12 +33,99 @@
...
+        field_name="creatiedatum",
+        lookup_expr="gte",
+    )
+    aanmaakdatum__lte = filters.DateFilter(
+        help_text=_(
+            "The creation date of this information object (lesser or equal to the given date): EXPERIMENTAL."
+        ),
+        field_name="creatiedatum",
+        lookup_expr="lte",
+    )
```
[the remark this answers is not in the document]
joeribekker: Correct, the query params should match up with the field name

src/openzaak/components/documenten/api/filters.py
```diff
@@ -31,12 +33,99 @@
...
+        lookup_expr="icontains",
+    )
+    locked = filters.BooleanFilter(
+        help_text=_(
+            "Indication of the Information Object being locked or not: EXPERIMENTAL."
+        ),
+        method="locked_filter",
+    )
+    titel = filters.CharFilter(
+        help_text=_("Titel van het Information Object: EXPERIMENTAL."),
```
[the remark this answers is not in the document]
annashamray: The main language is dutch despite we have some English help texts here and there/ Let's stick to Dutch if possible

src/openzaak/components/documenten/tests/test_filters.py
```diff
@@ -1,18 +1,103 @@
...
+            response = self.client.get(
+                reverse(EnkelvoudigInformatieObject), {"locked": False}
+            )
+            self.assertEqual(response.status_code, status.HTTP_200_OK)
+            data = response.json()
+            self.assertEqual(data["count"], 1)
+            self.assertEqual(data["results"][0]["locked"], False)
+
+    @override_settings(
+        NOTIFICATIONS_DISABLED=True,
```
annashamray, line 59: This is a default options for tests so it should be unnecessary here

src/openzaak/components/documenten/api/filters.py
```diff
@@ -31,12 +33,99 @@
...
+        #     object_type=ObjectInformatieObjectTypes.zaak
+        # ),
+        queryset=ObjectInformatieObject.objects.all(),
+        field_name="canonical__objectinformatieobject__zaak",
+    )
+
+    ordering = OrderingFilter(
+        help_text=_("sort on: EXPERIMENTAL."),
+        fields=(
+            "auteur",
```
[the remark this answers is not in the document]
annashamray: Agree with @stevenbal . Could you add missing tests for created query params ?

src/openzaak/components/documenten/api/filters.py
```diff
@@ -30,13 +33,119 @@
 logger = logging.getLogger(__name__)
 
 
+class ZaakFilter(FkOrUrlFieldFilter):
+    def filter(self, qs, value):
+        if not value:
+            return qs
+
+        if not check_path(value, "zaken"):
```
annashamray, line 41: I don't quite understand why you need this/ We don't have any special conditions for other `FkOrUrlFieldFilter` filters

src/openzaak/utils/help_text.py
```diff
@@ -0,0 +1,10 @@
+# SPDX-License-Identifier: EUPL-1.2
+# Copyright (C) 2019 - 2020 Dimpact
+from django.utils.translation import ugettext_lazy as _
+
+
+def mark_experimental(text):
+    warning_msg = _(
+        "Warning: this feature is experimental and not part of the API standard"
+    )
+    return "{} **{}**".format(text, warning_msg)
```
annashamray, line 10:
```suggestion
    warning_msg = _(
        "Warning: this feature is experimental and not part of the API standard"
    )
    return "**EXPERIMENTEEL** {}".format(text)
```

I think one word is enough. I saw Steven's suggestion, but it looks a little excessive. @stevenbal What do you think? 

Anyway this is an optional suggestion, no need to implement it if you disagree.

src/openzaak/utils/filterset.py
```diff
@@ -15,3 +16,12 @@ def filter_for_field(cls, field, field_name, lookup_expr=None):
...
         return filter
+
+
+class OrderingFilter(_OrderingFilter):
+    def filter(self, qs, value):
+        if value in constants.EMPTY_VALUES:
+            return qs
+
+        ordering = [self.get_ordering_value(param) for param in value]
+        return qs.order_by(*ordering).distinct()
```
annashamray, line 27: Do you really need `distinct` here? It can degrade our performance

src/openzaak/tests/test_help_text.py
```diff
@@ -0,0 +1,15 @@
+# SPDX-License-Identifier: EUPL-1.2
+# Copyright (C) 2020 Dimpact
+from django.test import TestCase
+
+from openzaak.utils.help_text import mark_experimental
+
+
+class HelptextTests(TestCase):
+    def mark_experimental_test(self):
```
annashamray, line 9: Nice to have such a case!

For the future: if you don't need an access to DB in your testcase you can use `uniitest.TestCase` class, which is a little faster than the django `TestCase`

src/openzaak/components/documenten/api/filters.py
```diff
@@ -30,13 +33,119 @@
...
+            "De indicatie of dit informatie object gelocked is of niet."
+        ),
+        method="locked_filter",
+    )
+    titel = filters.CharFilter(
+        help_text=mark_experimental("De titel van het informatie object."),
+        lookup_expr="icontains",
+    )
+    trefwoorden = CharArrayFilter(
+        help_text=mark_experimental(
```
annashamray, line 86: This is not an experimental feature, it exists in the standard

src/openzaak/components/documenten/api/filters.py
```diff
@@ -30,13 +33,119 @@
...
+    creatiedatum__lte = filters.DateFilter(
+        help_text=mark_experimental(
+            "De aanmakings datum van dit informatie object (kleiner of gelijk aan de gegeven datum)."
+        ),
+        field_name="creatiedatum",
+        lookup_expr="lte",
+    )
+    auteur = filters.CharFilter(
+        help_text=mark_experimental(
+            "De persoon of organisatie die dit informatie object heeft aangemaakt."
```
annashamray, line 67: Could you please add that this filter supports partial case insensitive match. We don't have it for other query params, so let's specify it explicitly.
The same comment applies for all `icontains` filters

src/openzaak/components/documenten/api/filters.py
```diff
@@ -30,13 +33,119 @@
...
+    vertrouwelijkheidaanduiding = filters.ChoiceFilter(
+        help_text=mark_experimental(
+            "De vertrouwelijkheidaanduiding van het informatie object"
+        ),
+        choices=VertrouwelijkheidsAanduiding.choices,
+    )
+    object_type = filters.ChoiceFilter(
+        help_text=mark_experimental("Het type van het gerelateerde OBJECT."),
+        field_name="canonical__objectinformatieobject__object_type",
+        lookup_expr="exact",
```
annashamray, line 100: I think it's a default lookup

src/openzaak/components/documenten/api/filters.py
```diff
@@ -30,13 +33,119 @@
...
+    )
+    titel = filters.CharFilter(
+        help_text=mark_experimental("De titel van het informatie object."),
+        lookup_expr="icontains",
+    )
+    trefwoorden = CharArrayFilter(
+        help_text=mark_experimental(
+            "Filter op de trefwoorden die eventueel gelinkt zijn aan het informatie object."
+        ),
+        lookup_expr="overlap",
```
annashamray, line 89: Why has the lookup been changed? This is a deviation of the standard. In the OAS there is no mentioning that it can be `OR` and in the reference implementation they also don't use `overlap`.
I think we should leave it as it was

src/openzaak/components/documenten/tests/test_filters.py
```diff
@@ -1,18 +1,379 @@
...
+            response = self.client.get(
+                reverse(EnkelvoudigInformatieObject), {"titel": "Lorem"},
+            )
+            self.assertEqual(response.status_code, status.HTTP_200_OK)
+            data = response.json()
+            self.assertEqual(data["count"], 2)
+            self.assertEqual(data["results"][0]["titel"], "Lorem")
+            self.assertEqual(data["results"][1]["titel"], "Lorem Ipsum")
+
+    def test_trefwoorden(self):
```
annashamray, line 150: I appreciate test coverage, but we already have tests for trefwoorden in [test_enkelvoudiginformatieobject.py](https://github.com/open-zaak/open-zaak/pull/1615/files?diff=unified&w=0#diff-f0d7f7a1ab4199b773388ce72a67d1e869d68e2f450658b21114b09d7dc8ad66)

If you want to refactor test you can move all filter tests from `test_enkelvoudiginformatieobject.py` to `test_filters (or vice versa)

src/openzaak/components/documenten/api/filters.py
```diff
@@ -31,12 +33,99 @@
...
+            "Indication of the Information Object being locked or not: EXPERIMENTAL."
+        ),
+        method="locked_filter",
+    )
+    titel = filters.CharFilter(
+        help_text=_("Titel van het Information Object: EXPERIMENTAL."),
+        field_name="titel",
+        lookup_expr="icontains",
+    )
+    trefwoorden = CharArrayFilter(field_name="trefwoorden", lookup_expr="overlap")
```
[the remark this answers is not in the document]
annashamray: Because it was not experimental. It existed before, Bart just changed lookup. I'll address the lookup change in my review

annashamray (changes requested):
It looks good. Just some minor remarks

src/openzaak/components/documenten/api/filters.py
joeribekker: I wasn't aware the `trefwoorden` filter was already part of the OAS. Let's keep the default behaviour in that case. We probably need some way to differentiate between AND and ORs...later :)

src/openzaak/components/documenten/api/filters.py
joeribekker: Please rename `object_type` to `objectinformatieobjecten__objectType`

joeribekker:
Please add the query param filter `informatieobjecttype`

src/openzaak/utils/help_text.py
stevenbal: That's fine for me, I just wanted a function for it to make it consistent everywhere :+1:

src/openzaak/components/documenten/api/filters.py
```diff
@@ -30,13 +33,119 @@
...
+        help_text=mark_experimental(
+            "URL-referentie naar de gerelateerde ZAAK (in deze of een andere API)."
+        ),
+        queryset=EnkelvoudigInformatieObject.objects.all(),
+        field_name="canonical__objectinformatieobject__zaak",
+    )
+
+    ordering = OrderingFilter(
+        help_text=mark_experimental("Sorteer op."),
+        fields=(
```
stevenbal, line 114: I looked over this in the first review, but in the PR description I see `bestandstype` as an ordering filter as well, which is missing. It's not a field on any of the models though, I think that might be a filter on `formaat`? @joeribekker

stevenbal (changes requested):
Just the question about the missing `bestandstype` ordering filter, other than that it looks good :+1:.

A general thing (not sure what @annashamray thinks about this), but I would prefer it if PR feedback is added in a separate commit (or commits), rather than amending the original commits that were already reviewed. That way it's easier to see the diff of what changed after feedback was processed

annashamray:
Agree with Steven about separate commits. Github then politely mark the adjusted files as changes since the last review

src/openzaak/utils/filterset.py
bart-maykin: Yes we need distinct here because the original queryset already is ordered with distinct. Using order_by without it will result in errors.

stevenbal (approved):

joeribekker (approved):

PR #1624, opened 2024-04-19
✨ [#1611] Add setup configuration step for authorization

### LINKED ISSUE #1611: Allow CICD-configuration for authorisations
### Thema / Theme

Other

**Probleem**

1. Technische configuratie-items zijn niet via CI/CD pipeline te besturen:
• Dit zorgt ervoor dat alle configuratie handmatig moet worden uitgevoerd
• Elke gemeente moet separaat en handmatig worden geconfigureerd met veel deplicatie
• Het gevolg is dat Dimpact niet schaalbaar kan zijn als beheerorganisatie

2. Technisch en functioneel beheer zijn onvoldoende gescheiden
• Dit is een probleem aangezien gevolgen van aanpassingen voor technisch beheer groot zijn
(risico voor beschikbaarheid applicaties)
• Niet elke beheerder moet technisch beheer rechten krijgen en voor sommige gemeentes zal
alleen Dimpact het technisch beheer doen

**Acceptance criteria**

1. Technische configuratie kan programmatisch worden ingesteld vanuit de CI/CD pipeline
2. Dit is van toepassing op alle componenten die in gebruik zijn binnen PodiumD
3. Technisch beheer kan nog steeds via een GUI worden aangepast
4. Middels rollen en rechten is technisch beheer gescheiden van functioneel beheer

**Oplossingsrichting**

1. Er wordt van elk component een inventarisatie gemaakt welke environment variabelen wel/niet
ondersteund worden.
2. Er wordt overkoepelend bepaald welke environment variabelen nog moeten worden
toegevoegd.
3. Een aantal environment variabelen zal effect hebben op in de database opgeslagen waarden. De
database waarden worden overschreven met de waarden uit de environment variabelen.
4. Er komt in alle componenten een “setup_configuration” commando, waarmee environment
variabelen worden opgeslagen in de database. Dit commando kan middels een Kubernetes job
worden aangeroepen, met een init-container, of handmatig.
5. Een aantal configuratie instellingen zijn dermate uitgebreid dat deze niet met eenvoudige key-
value (standaard environment variabelen) methodiek kunnen worden opgegeven. Daarom zijn
sommige waarden middels een YAML-bestand in te laden, waarnaar verwezen kan worden met
een environment variabele. Het YAML-bestand zelf kan een gemounte ConfigMap zijn.
6. Environment variabelen die database waarden overschrijven die ook in de beheer UI zijn aan te
passen, blijven aanpasbaar in de beheer UI. Middels rechten kan je deze alleen-lezen maken. We
gebruiken hiervoor het standaard rechten-model in het technische framework, en bouwen geen
laag daar boven op.
7. Het einddoel is om Open Zaak en Open Notificaties volledig werkend en geconfigureerd te
hebben op een Kubernetes cluster, in het bijzonder op PodiumD, zonder dat in de
beheerinterface hoeft te worden ingelogd.

No more admin UI after deploy/update

_No response_

### LINKED ISSUE #1655: Add documentation for authorization setup configuration step
Feature is implemented in https://github.com/open-zaak/open-zaak/pull/1624
Related issue: https://github.com/open-zaak/open-zaak/issues/1611

describe:
- [x] dumping config
- [x] describe that it's important to use the same uuids when importing/exporting catalogi
- [x] loading config
- [x] mapping for domains
- [x] describe behavior of regular vs --overwrite (also what happens in case of collisions)
  * no overwrite: if all applicatie UUIDs and JWTsecret identifiers are present, no changes are made (todo check what happens if one of them is not present, will it update everything?)
  * overwrite: update existing entries

stevenbal:
Fixes #1611
Fixes #1655 
Requires https://github.com/maykinmedia/commonground-api-common/pull/27 to be merged and released

**Changes**
* Add command to dump authorization configuration to a file
* Add configuration step to take this file as input (together with a domain mapping file) and import the authorization setup
* Add documentation


src/openzaak/config/bootstrap/authorizations.py
```diff
@@ -0,0 +1,90 @@
...
+                            content = content.replace(
+                                domain, entry[settings.ENVIRONMENT]
+                            )
+
+            # Load via stdin to avoid having to write to a temporary file
+            with override_stdin(StringIO(content)):
+                call_command("loaddata", "-", format="yaml")
+
+    def test_configuration(self):
+        ...
```
stevenbal, line 90: I'm not sure how we would test if this works, it seems way too excessive to try and access every zaaktype in the autorisaties we just loaded :sweat_smile:

annashamray:
You can add the doc in the separate issue, let's just not forget about it

stevenbal:
created a separate issue for the docs: https://github.com/open-zaak/open-zaak/issues/1655

src/openzaak/management/commands/dump_auth_fixtures.py
```diff
@@ -0,0 +1,31 @@
+# SPDX-License-Identifier: EUPL-1.2
```
annashamray, line 1: You know it looks like you just reuse the existing command without modifying it. I think it could better be a  bash file, just like `bin/generate_admin_index_fixture.sh`, which we already have.

src/openzaak/config/tests/bootstrap/test_load_authorizations.py
```diff
@@ -0,0 +1,326 @@
...
+
+
+class AuthorizationConfigurationTests(TestCase):
+    maxDiff = None
+
+    def setUp(self):
+        super().setUp()
+
+        self.temp_dir = tempfile.TemporaryDirectory()
+        self.dump_path = str(Path(self.temp_dir.name) / "auth_dump.yaml")
```
annashamray, line 41: I think it could be nice to have a real `auth_dump.yaml` file added to the test folder, so we could see its content as an example. The same applies for `mapping.yaml` file

src/openzaak/config/bootstrap/authorizations.py
```diff
@@ -0,0 +1,90 @@
...
+            }
+            secrets_configured = identifiers.issubset(existing_identifiers)
+
+        return applicaties_configured and secrets_configured
+
+    def configure(self):
+        with open(settings.AUTHORIZATIONS_CONFIG_FIXTURE_PATH) as original_file:
+            content = original_file.read()
+
+            if settings.AUTHORIZATIONS_CONFIG_DOMAIN_MAPPING_PATH:
```
annashamray, line 71: Have you discussed the shape of mapping file with Joeri or Sjoerd? It looks like a bit of overkill, it could be just an env var `AUTHORIZATIONS_CONFIG_TARGET_DOMAIN` or something like this
On the other hand we make these commands for devops people. so I don't know what it more convenient for them

src/openzaak/management/commands/dump_auth_fixtures.py
```diff
@@ -0,0 +1,31 @@
...
+
+        with open(path, "w") as f:
+            call_command(
+                "dumpdata",
+                "authorizations",
+                "autorisaties",
+                "vng_api_common.jwtsecret",
+                natural_foreign=True,
+                natural_primary=True,
+                format="yaml",
```
annashamray, line 29: TIL that you can generate fixtures in yaml out of the box

annashamray (changes requested):
It looks very neat and nice! 

1. For now it works like `loaddata` django command with domain overwriting, right? Do we plan to make it more like a `AutorisatieForm` in the admin, so the user could choose component and `related_type_selection` for this component? Do we want to provide it? Do clients want it? Maybe it can be discussed with @joeribekker as a next iteration of this config? It seems nice to have all authorizations configured without providing a single zaaktype url. But it could be the case that no one needs it and I overthink it.
2. Please add this step to the web_init in docker-compose file. It can be then shown to devops as an example how to configure this step

stevenbal:
@annashamray 

> For now it works like loaddata django command with domain overwriting, right? 

that's right, it mostly relies on loaddata

> Do we plan to make it more like a AutorisatieForm in the admin, so the user could choose component and related_type_selection for this component?

I think this should already be possible, because the `AutorisatieSpec`s are exported/imported as well?

> Please add this step to the web_init in docker-compose file. It can be then shown to devops as an example how to configure this step

will do :+1:

src/openzaak/config/bootstrap/authorizations.py
stevenbal: I have discussed it with Joeri, it could indeed be a single target domain, but the thing is that we do support the usage of external catalogi as well, so if we have an import file that contains multiple different domains, the import wouldn't hold up.

I do still wonder if this is actually used anywhere in practice though, maybe we should give two options to simplify the config for instances that don't use external catalogi:
- either supply a `AUTHORIZATIONS_CONFIG_TARGET_DOMAIN` and just map all URLs to that domain (this assumes that the import file does not contain any external catalogi
- or supply a mapping file in case there are external catalogi?

src/openzaak/management/commands/dump_auth_fixtures.py
stevenbal: Hmm yeah I'll move it to a bash script, though I do think at some point we'll want to be able to export from the admin interface, but we can just do `call_command` with `dumpdata` in that case

src/openzaak/config/tests/bootstrap/test_load_authorizations.py
stevenbal: I removed the tempfiles and added two static .yaml files

annashamray:
@joeribekker @stevenbal Should we park this issue until https://github.com/open-zaak/open-zaak/issues/1661 is implemented? I see that nobody is assigned to it, what is our plan?

stevenbal:
@annashamray yeah I think it would be better to wait for that. I think we're waiting for approval by Den Haag for the estimation

src/openzaak/config/tests/bootstrap/test_load_authorizations.py
```diff
@@ -0,0 +1,324 @@
...
+    def test_configure_catalogus_does_not_exist(self):
+        """
+        Running `.configure` while some of the expected data (Catalogus) does not exist
+        should raise errors
+        """
+        self.catalogus.delete()
+
+        # Attempt to run the import
+        with self.assertRaises(ConfigurationRunFailed):
+            AuthorizationConfigurationStep().configure()
```
stevenbal, line 163: I'm not sure what else we can do in case the Catalogus does not exist, not sure if we can avoid errors from being raised because we use `loaddata`

docs/installation/config/openzaak_config_cli.rst
```diff
@@ -80,16 +80,36 @@ The Selectielijst API is not expected to require any authentication.
...
+environments, for more detailed information about how this works, see :ref:`Authorization configuration <installation_configuration_auth_cicd>`.
+
+* ``AUTHORIZATIONS_CONFIG_ENABLE``: enable authorization configuration, defaults to ``False``.
+* ``AUTHORIZATIONS_CONFIG_FIXTURE_PATH``: the full path of the fixture ``.yaml`` file containing the
+  configuration (for more information about how to generate this file, see :ref:`Generating the fixture <authorization_config_generate_fixture>`).
+* ``AUTHORIZATIONS_CONFIG_DOMAIN_MAPPING_PATH``: the full path to the file with the domain mapping.
+* ``AUTHORIZATIONS_CONFIG_DELETE_EXISTING``: a boolean indicating whether or not the existing configuration should be removed before loading the fixture.
+  set this to ``True``, ``true`` or ``1`` if you want to use the fixture as a single source of truth. Make sure to only do this if you are **not**
+  managing any Applicaties/credentials via the admin or API. Defaults to ``False``.
+* ``ENVIRONMENT``: the environment name of this Open Zaak instance. This is used to determine what the target domain is in the domain mapping.
```
stevenbal, line 109: Can we rely on this existing envvar? Or should I add a new one for this?

src/openzaak/config/tests/bootstrap/test_load_authorizations.py
stevenbal: On second thought, it currently will only be possible to run this step after catalogi have been configured (be it manually or via some fixture or imports), because otherwise this step will fail

bin/setup_configuration.sh
```diff
@@ -9,4 +9,5 @@ set -e
 ${SCRIPTPATH}/wait_for_db.sh
 
 src/manage.py migrate
+${SCRIPTPATH}/load_fixtures.sh
```
stevenbal, line 12: I added this because otherwise it won't be possible to run the setup configuration part for authorizations, because the catalogi do not exist yet

src/openzaak/config/bootstrap/authorizations.py
```diff
@@ -0,0 +1,118 @@
...
+                if row["model"] == "vng_api_common.jwtsecret"
+            }
+            secrets_configured = identifiers.issubset(existing_identifiers)
+
+        return applicaties_configured and secrets_configured
+
+    @transaction.atomic
+    def configure(self):
+        if settings.AUTHORIZATIONS_CONFIG_DELETE_EXISTING:
+            delete_existing_configuration()
```
stevenbal, line 86: @annashamray I added this to allow using the fixture as a single source of truth, but I just realized that other ConfigurationSteps also add applicaties and jwtsecrets. How should I treat those? I guess I could exclude the stuff created in previous steps from the .delete, but I wanted to check with you before I make it more complicated

stevenbal:
I rebased this PR to work with CatalogusAutorisaties, so it is ready for review, though I do have several questions relating to the implementation in the PR

docs/installation/config/openzaak_config_cli.rst
annashamray: Let's reuse it, I think it should be fine

src/openzaak/config/bootstrap/authorizations.py
annashamray: It can be a leaky abstraction. but I'd check if other steps which add autorisation data are enabled. 
If they are enabled and AUTHORIZATIONS_CONFIG_DELETE_EXISTING = True, I'd raise an Exception

It's first time then we have configuration steps that can't be used together. Maybe in the future we can add some steprelation validator that can check which will have some mapping of steps that can't be run together?

src/openzaak/config/bootstrap/authorizations.py
```diff
@@ -0,0 +1,118 @@
...
+                            )
+
+            # Load via stdin to avoid having to write to a temporary file
+            logger.info(
+                "Loading authorization configuration from %s",
+                settings.AUTHORIZATIONS_CONFIG_FIXTURE_PATH,
+            )
+            with override_stdin(StringIO(content)):
+                try:
+                    call_command("loaddata", "-", format="yaml")
```
annashamray, line 112: My favorite question - should we have an additional validation for the fixture? For example, checking if zaaktypen are existing urls or checking that catalog autorisatie and individual autorisatie don't overlap? 

Maybe we can do it in the `test_configuration` part?

src/openzaak/config/tests/bootstrap/test_load_authorizations.py
annashamray: I think raising `ConfigurationRunFailed` if the auth fixture is not correct is an expected behaviour :+1:

src/openzaak/config/tests/bootstrap/test_load_authorizations.py
```diff
@@ -0,0 +1,324 @@
...
+        self.assertEqual(autorisatie_on.scopes, [str(SCOPE_ZAKEN_ALLES_LEZEN)])
+        self.assertEqual(autorisatie_on.zaaktype, ZAAKTYPE2)
+        self.assertEqual(
+            autorisatie_on.max_vertrouwelijkheidaanduiding,
+            VertrouwelijkheidsAanduiding.openbaar,
+        )
+
+    def test_configure_overwrite_with_existing_other_configuration(self):
+        """
+        Running `.configure` should preserve data that was not present in the fixture
```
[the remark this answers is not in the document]
annashamray: I think you already solved it by introducing `AUTHORIZATIONS_CONFIG_DELETE_EXISTING` env var

annashamray (commented):
Nice code and documentation!

src/openzaak/config/tests/bootstrap/test_load_authorizations.py
stevenbal: I think so too, and with the change to `bin/setup_configuration.sh` to load fixtures first if `AUTHORIZATIONS_CONFIG_ENABLE` is True, this is no longer an issue

src/openzaak/config/bootstrap/authorizations.py
stevenbal: Since the mapping file construction is already there and documented, I'll leave it in

src/openzaak/config/bootstrap/authorizations.py
stevenbal: I think validating it might be a good idea, though I'm not sure if `test_configuration` is the right place, since that runs after `configure`, so you'd be left with a selftestfailure, but the configuration was loaded.

I'm not sure if it's easy to keep all the validation between the admin interface, API and this configuration step in sync, I'll see what I can do with it

stevenbal: I've been trying to do this, but I don't think it's possible to share the same validation between the admin form and the configuration step without duplicating it unfortunately, particularly validating the overlap between autorisaties/catalogusautorisaties is annoying, because the fixture has a different shape for the data than the AutorisatieFormSet (where multiple autorisaties are grouped together). To further complicate things, in the form validation `zaaktypen/iotypen/besluittypen` are resolved using django loose fk, while in the configurationstep this has to be done manually as well

@annashamray do you have any ideas? :grimacing:

src/openzaak/config/bootstrap/authorizations.py
stevenbal: I added code to raise an exception if any of OPENZAAK_NOTIF_CONFIG_ENABLE, NOTIF_OPENZAAK_CONFIG_ENABLE, DEMO_CONFIG_ENABLE are True

src/openzaak/config/bootstrap/authorizations.py
annashamray: Let's park this for now and see how this configuration step is used to determine the next step

annashamray (approved):
Good job!

PR #1614, opened 2024-04-09
:arrow_up: [#1597] update to django 4.2

annashamray:
Fixes #1597 (OZ part)

PR #1634, opened 2024-04-23
Feature/update syntax

### LINKED ISSUE #1597: Update to Django 4.2 in Open Zaak
This depends on OAf library.

Tasks:
* [x] Update drc-cmis https://github.com/open-zaak/cmis-adapter/pull/82
* [x] Update django-loose-fk https://github.com/maykinmedia/django-loose-fk/pull/33
* [x] Update related dependencies in Open Zaak - https://github.com/open-zaak/open-zaak/pull/1614
* [x] Update code using Django 4.2 and python 3.10 syntax - #1634

annashamray:
Fixes #1597 (syntax part)
Depends on #1614


src/openzaak/components/documenten/tests/test_upload_cmis.py
```diff
@@ -286,13 +288,17 @@ def test_update_eio_metadata(self):
         self.assertEqual(eio_new.versie, 2)
         self.assertEqual(eio_new.titel, "another summary")
         self.assertEqual(eio_new.bestandsomvang, 4)
-        self.assertEqual(eio_new.canonical.bestandsdelen.count(), 0)
+        self.assertEqual(
+            BestandsDeel.objects.filter(informatieobject_uuid=eio_new.uuid).count(), 0
```
Coperh, line 292: I am wondering why this has changed. Will look when Django 4.2 is merged.

src/openzaak/components/zaken/tests/models/test_zio_block_change.py
```diff
@@ -33,5 +33,5 @@ def test_bulk_update(self):
         )
 
     def test_bulk_create(self):
-        zio = ZaakInformatieObjectFactory.build()
+        zio = ZaakInformatieObjectFactory.build(with_etag=False)
```
Coperh, line 36: Similar Question here.

src/openzaak/components/catalogi/admin/forms.py
```diff
@@ -631,8 +631,9 @@ class Meta:
         model = ZaakTypenRelatie
 
     def _get_validation_exclusions(self):
-        extra = ["gerelateerd_zaaktype"]
-        return super()._get_validation_exclusions() + extra
+        exclude = super()._get_validation_exclusions()
+        exclude.add("gerelateerd_zaaktype")
```
Coperh, line 635: Also here. Why needed now, not before.

src/openzaak/components/catalogi/tests/admin/test_published_readonly.py
```diff
@@ -80,13 +80,13 @@ def test_readonly_zaaktype(self, m):
...
-        self.assertEqual(procestype.text, "1 - Instellen en inrichten organisatie")
+        self.assertTrue("1 - Instellen en inrichten organisatie" in procestype.text)
         behandeling = response.html.find(
             class_="field-doorlooptijd_behandeling"
         ).div.div
-        self.assertEqual(
-            behandeling.text,
-            ngettext_lazy("{days} day", "{days} days", 10).format(days=10),
+        self.assertTrue(
+            ngettext_lazy("{days} day", "{days} days", 10).format(days=10)
```
Coperh, line 88: Why now in instead of equals?

Coperh (commented):
Looks good. Some minor questions. I'll tests these again with Django 4.2 but 170 out of 176 files I have no issue with.

src/openzaak/components/documenten/tests/test_upload_cmis.py
annashamray: Hi! This change was made in https://github.com/open-zaak/open-zaak/issues/1597 
This Pr is only to run `pyupgrade` and `django-upgrade` and remove soon to be deprecated syntax

Explanation: In Django 4.2 the descriptor for reverse-fk fields was change. In the function `create_reverse_many_to_one_manager` in the `get_queryset` they added a check:
```
  if self.instance.pk is None:
      raise ValueError(
          f"{self.instance.__class__.__name__!r} instance needs to have a "
          f"primary key value before this relationship can be used."
      )
```
This check breaks all attempts to get reverse-fk fields for instances without `pk`. In this case `eio_new.canonical` doesn't have a pk, so it crushes now.

src/openzaak/components/catalogi/admin/forms.py
Coperh: I did not see the `+ extra` in the previous version, so no issues here.

src/openzaak/components/zaken/tests/models/test_zio_block_change.py
annashamray: Short explanation: it's because of the change in `create_reverse_many_to_one_manager` just like in the comment above. 

Detailed explanation: 
* factory step builder loops through all post_declarations (`with_etag` is also a post_declaration). 
* It tries to get `with_etag` value, fails and tries to raise `AttributeError`
* `AttributeError` message needs `EnkelvoudigInformatieObjectCanonical.__str__`, which uses `latest_version` property
* `latest_version` poperty uses reverse relations and calls `create_reverse_many_to_one_manager` which failes because the instance doesn't have a pk.

In Django 3.2 getting `with_etag` raises `AttributeError` successfully, which is caught later in the builder and the default value is used.

src/openzaak/components/catalogi/tests/admin/test_published_readonly.py
annashamray: In django 4.2 they added additional `div` with `class="flex-container"` inside the field template, so now we need either to add one extra `.div` when finding the html element or change the assetion. I used the latter approach. Tbh I don't remember why the tests for formatting were added, but it looks like we test a django template here. So I think we can remove them in the future

Coperh: Got it

Coperh (approved):

SonnyBA (approved):

PR #1641, opened 2024-04-29
:sparkles:[#1636] Remove AutorisatieSpec when all_current_and_future …

### LINKED ISSUE #1636: Authorizations aren't being stored as expected

1.12

Taiga DH 554

Als ik Autorisaties op een zaaktype probeer toe te voegen wordt dit niet doorgezet in een aantal gevallen.

Reproductie:   

1. Switch op de bestaande autorisatie van de zaken api de selectie van ‘Alle huidige en toekomstige…’ naar ‘Selecteer handmatig’
2. Vul bij zaken autorisaties een zaaktype uit de (onderste) catalogus en druk op opslaan.
3. Kijk daarna weer in het autorisatie overzicht. Het is niet doorgevoerd

----

Confirmed on https://test.openzaak.nl/admin/authorizations/applicatie/10654/autorisaties/

For the existing application toggle ‘Selecteer handmatig’ and select one or more zaaktypes. After saving the autorisaties are created, however I suspect that the existing autorisatie that gives permissions to all current and future zaaktypes isn't cleared as expected.

Workaround is to re-create the application with the correct authorizations, which is a bit of a pain

image

_No response_

_No response_

Coperh:
Fixes #1636


src/openzaak/components/autorisaties/models.py
```diff
@@ -63,7 +63,7 @@ class Meta:
         # informatieobjecttypen/besluittypen, the scopes/VA are constant for
         # these. The exact *type is implied by the component choice, so we need
         # a unique constraint on applicatie + component
-        unique_together = ("applicatie", "component")
+        unique_together = ("applicatie", "component", "scopes")
```
Coperh, line 66: Is this recommend? It feels wrong.

Coperh:
Did not end up making too much changes. I think it should be refactored eventually because that admin view is a big mess.

Coperh:
Oh BTW, the issue is that `AutorisatieSpec` is not removed when all_current_and_future is not selected. 

Also it shared by a component, so if there is an `AutorisatieSpec` for Zaken for example, all autorisaties for zaken will default to all_current_and_future selected when you open the admin.

stevenbal (changes requested):

src/openzaak/components/autorisaties/models.py
Coperh: As far as I know, the form should maintain the order of the scopes. I have added a test for this.

annashamray: @Coperh it is possible to use array field in unique index in postgres, but yeah it needs either presorting or adding custom Postgres `sort` function and creating index with the usage of this function.

stevenbal (approved):

PR #1642, opened 2024-04-29
:sparkles:[#1584] get autorisaties URL from OPENZAAK_DOMAIN or site

### LINKED ISSUE #1584: When specifying a Zaken API authorization with permissions linked to an internal zaaktype, deleting the zaaktype causes a 500 error

Coperh:
Fixes #1584

### DIFF
--- a/src/openzaak/components/autorisaties/forms.py
+++ b/src/openzaak/components/autorisaties/forms.py
@@ -24,6 +24,7 @@ from openzaak.components.catalogi.models import (
     InformatieObjectType,
     ZaakType,
 )
+from openzaak.utils import build_absolute_url
 from openzaak.utils.auth import get_auth
 from openzaak.utils.middleware import override_request_host
 from openzaak.utils.validators import ResourceValidator
@@ -397,8 +398,11 @@ class AutorisatieForm(forms.Form):
             autorisaties = []
             for _type in types:
                 data = autorisatie_kwargs.copy()
-                url = request.build_absolute_uri(_type.get_absolute_api_url())
-                data[_field_info["_autorisatie_type_field"]] = url
+
+                # signal uses build_absolute_url to find autorisaties to delete
+                data[_field_info["_autorisatie_type_field"]] = build_absolute_url(
+                    _type.get_absolute_api_url()
+                )
                 autorisaties.append(
                     Autorisatie(
                         max_vertrouwelijkheidaanduiding=vertrouwelijkheidaanduiding,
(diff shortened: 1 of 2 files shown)


src/openzaak/components/autorisaties/forms.py
```diff
@@ -392,8 +393,11 @@ def save(self, applicatie: Applicatie, request, commit=True):
             autorisaties = []
             for _type in types:
                 data = autorisatie_kwargs.copy()
-                url = request.build_absolute_uri(_type.get_absolute_api_url())
-                data[_field_info["_autorisatie_type_field"]] = url
+
+                # signal uses build_absolute_url to find autorisaties to delete
```
Coperh, line 397: Also does not have access to the request

Coperh:
This only matters if the request is sent with a domain other than the expected one.

Assuming the site or openzaak domain is set properly this should work. If not set properly, the deleting signal will not work regardless.

I am not sure if this affects the rewrite host stuff from #1585

annashamray (commented):
Isn't it already fixed by https://github.com/open-zaak/open-zaak/pull/1588/? 
@stevenbal What do you think?

annashamray:
@Coperh Could you reproduce the error in the regression test?

stevenbal:
Is there a way to reproduce this bug locally as well?

Coperh:
Idk if my test works, but it became bugged for me locally.

@stevenbal  basically save autorisaties with on `localhost` or `127.0.0.1` and have your `OPENZAAK_DOMAIN` or site domain be different.

That's what I assume the issue was for the original complain.

stevenbal:
> basically save autorisaties with on localhost or 127.0.0.1 and have your OPENZAAK_DOMAIN or site domain be different. Delete a zaaktype with a related autorisatie. It should be deleted automatically.

Ah I see, I tested it locally on main and this branch, and this fixes it indeed :+1:

stevenbal (approved):

Issue #1637, opened 2024-04-26
Add maykin_2fa


stevenbal:
I thought 2FA was already part of Open Zaak and Open Notificaties, but it turns out that was never the case. I think it would be good idea to add that, just like it was added in Objects/Objectypes API

https://maykin-2fa.readthedocs.io/en/stable/

joeribekker:
Since, this is up... I would really like to add the same construction we did in Open Formulieren: based on the auth backend, use 2fa or not. And ofcourse, add 2fa to OZ and ON :)

For the one pickinf this up, please updated this ticket to reflect what needs to be done in all the projects :)

joeridiederen:
Hi @joeribekker; weet je wanneer deze optie in Open Notificaties gaat landen? Ik zie dat dit ticket zowel OZ als ON beschrijft, maar op de ON-backlog vind ik 'm niet terug (https://github.com/open-zaak/open-notificaties/milestones). Alvast dank!

joeribekker:
@joeridiederen klopt, we merkten dat we steeds tickets kopieerden van component A naar B dus we proberen iets meer te centraliseren maar daardoor zie je het niet direct op de backlog van het component.

joeribekker:
Het is overigens wel gedaan: https://github.com/open-zaak/open-notificaties/pull/151

PR #1644, opened 2024-05-03
:sparkles: [#1637] added 2fa

bart-maykin:
Fixes [#1637](https://github.com/open-zaak/open-zaak/issues/1637)

Added 2fa


src/openzaak/templates/maykin_2fa/base.html
```diff
@@ -0,0 +1,11 @@
...
+{% comment %} SPDX-License-Identifier: EUPL-1.2 {% endcomment %}
+{% comment %} Copyright (C) 2019 - 2020 Dimpact {% endcomment %}
+
+{# Django 3.2 #}
+{% block breadcrumbs %}{% endblock %}
+
+{# Do not show any version information #}
+{% block footer %}
+    <div id="footer"></div>
+{% endblock %}
```
stevenbal, line 11: Same comment as for ON, this file can be removed

src/openzaak/templates/maykin_2fa/login.html
```diff
@@ -0,0 +1,25 @@
...
+{% block extra_recovery_options %}
+<li>
+    {% trans 'Contact support to start the account recovery process' %}
+</li>
+{% endblock extra_recovery_options %}
+
+{# Do not show any version information #}
+{% block footer %}
+    <div id="footer"></div>
+{% endblock %}
```
stevenbal, line 25: Same comment as for ON, this block can be removed
```suggestion
```

src/openzaak/templates/maykin_2fa/base.html
stevenbal: Sorry, I was wrong about removing the `base.html` file, it turns out Open Zaak does use this `footer` block https://github.com/open-zaak/open-zaak/blob/e701f8155735949033fc53f0c78ac4ce9fcc97ca/src/openzaak/templates/admin/base_site.html#L43 (while Open Notificaties does not). Could you re-add it here?

stevenbal (changes requested):
Just the last comment to re-add `base.html`, then I'll approve

stevenbal (approved):

annashamray:
What about backward compatibility? Now we will force all our clients to set up 2fa for the admin without any notice. 
@bart-maykin please make it possible to disable it for now. 
We will announce this feature but not make it required this time

PR #1654, opened 2024-05-14
Feature/1596 document bulk import

### LINKED ISSUE #1596: [FEATURE-REQUEST] Bulk-import of Documents in Open Zaak

Organisatie / Organization: Utrecht
Naam / Name: Alex vd Linde

### Omschrijving / Summary

Taiga Utrecht 175

The request is to add a bulk-import ability for Documents in Open Zaak, which uses a JSON or CSV containing filenames, types and references to the file (from local storage). 

For further details see the issue and PDF in Taiga, issue has been discussed by Sergei with PO

### Urgentie / Priority
High prio

SonnyBA:
Fixes #1596 

**Changes**
Adds the feature to bulk import documents through a metadata import file and thereby bypassing the need to do a lot of API calls through the document create endpoint.


src/openzaak/utils/management/commands/create_export.py
SonnyBA, line 1: This command is helpful (in combination with the `generate_data` command) to create a test import metadata file but should be removed before actually merging this PR.

src/openzaak/utils/models.py
```diff
@@ -12,3 +22,100 @@ def clone_object(instance):
...
+    @classproperty
+    def report_choices(cls):
+        return {cls.finished, cls.error}
+
+
+class ImportTypeChoices(models.TextChoices):
+    documents = "documenten", _("Enkelvoudige informatie objecten")
+
+
+class Import(models.Model):
```
SonnyBA, line 50: I wasn't really sure in which app we would like to add this model hence its in `utils` at the moment

annashamray: Let's put it into separate app. I think it deserves it

annashamray: On the second thought maybe it could be put into `components.documenten.import` app, since we don't make it generic

src/openzaak/utils/permissions.py
```diff
@@ -188,3 +189,9 @@ def has_object_permission(self, request: Request, view, obj) -> bool:
         main_object_data = self.format_data(main_object, request)
         fields = self.get_fields(main_object_data)
         return request.jwt_auth.has_auth(scopes_required, component, **fields)
+
+
+class ImportAuthRequired(AuthRequired):
+    def get_component(self, view) -> str:
+        importer_type = view.import_type
+        return ImportTypeChoices.get_component_from_choice(importer_type)
```
annashamray, line 197: It looks like overkill now, before Import process becomes generic. 

Let's also discuss permissions. What happens if you try to import a document which has an informatieobjecttype which you don't have rights for? 
Should we stop the whole process or should we just not import this particular document? I think it should be refined with Joeri

src/openzaak/utils/serializers.py
```diff
@@ -48,3 +56,22 @@ def get_from_serializer_data_or_instance(
         return None
 
     return serializer_field.get_attribute(instance)
+
+
+class ImportSerializer(ModelSerializer):
+    status = SerializerMethodField()
+
+    @extend_schema_field(OpenApiTypes.STR)
+    def get_status(self, instance):
```
annashamray, line 65: It looks excessive. Let's just use `status` field

src/openzaak/utils/views.py
```diff
@@ -113,3 +128,162 @@ def initialize_request(self, request, *args, **kwargs):
...
+class ImportStatusView(mixins.RetrieveModelMixin, viewsets.GenericViewSet):
+    import_type: ImportTypeChoices
+    serializer_class = ImportSerializer
+    permission_classes = (ImportAuthRequired,)
+
+    lookup_field = "uuid"
+
+    def get_queryset(self):
+        return Import.objects.filter(
+            import_type=self.import_type, status__in=ImportStatusChoices.visible_choices
```
annashamray, line 268: I don't quite get the logic behind filtering on status here

src/openzaak/utils/serializers.py
```diff
@@ -48,3 +56,22 @@ def get_from_serializer_data_or_instance(
         return None
 
     return serializer_field.get_attribute(instance)
+
+
+class ImportSerializer(ModelSerializer):
+    status = SerializerMethodField()
```
annashamray, line 62: Why is it not `HyperlinkedModelSerializer` ?

src/openzaak/utils/views.py
```diff
@@ -113,3 +128,162 @@ def initialize_request(self, request, *args, **kwargs):
...
+            # This should be updated when the import metadata file is parsed
+            total=0,
+        )
+
+        return Response(
+            data={
+                "upload_url": import_instance.get_upload_url(request=request),
+                "status_url": import_instance.get_status_url(request=request),
+                "report_url": import_instance.get_report_url(request=request),
+            },
```
annashamray, line 172: This can be added to a separate serializer, which can be used then by drf-spectacular, so there would be no need to specify it manually

src/openzaak/utils/views.py
```diff
@@ -113,3 +128,162 @@ def initialize_request(self, request, *args, **kwargs):
...
+        ) % ",".join(missing_headers)
+
+        raise ValidationError(
+            {"__all__": [error_message]}, code="missing-import-headers"
+        )
+
+
+class ImportUploadView(mixins.CreateModelMixin, viewsets.GenericViewSet):
+    lookup_field: str = "uuid"
+    parser_classes = (CSVParser,)
```
annashamray, line 218: Can it be JSON and not CSV? This way we could validate the shape of the file

src/openzaak/utils/views.py
```diff
@@ -113,3 +128,162 @@ def initialize_request(self, request, *args, **kwargs):
...
+            error_message = (
+                _(
+                    "Starting an import process is not possible due to the current "
+                    "status of the import: %s"
+                )
+                % import_status.label
+            )
+            raise ValidationError({"__all__": [error_message]}, code="invalid-status")
+
+        request_data = request.data or ""
```
annashamray, line 244: Is empty request body allowed?

src/openzaak/utils/tasks.py
```diff
@@ -0,0 +1,765 @@
...
+
+        if include_header:
+            return total
+
+        return total - 1 if total > 0 else 0
+
+
+@dataclass
+class DocumentRow:
+    _identificatie: str
```
annashamray, line 63: Supporting both csv format and document serializer seems so bothersome and not worth it. Can we reuse `EnkelvoudigInformatieObjectSerializer` for it? We do it in Catalogi `import` command, and it adds up as an argument to use JSON format 

It will save us so much headache and will reduce operational risk significantly

src/openzaak/utils/tasks.py
```diff
@@ -0,0 +1,765 @@
...
+
+    batch: list[DocumentRow] = []
+
+    zaak_uuids = {str(uuid): id for uuid, id in Zaak.objects.values_list("uuid", "id")}
+
+    for row_index, row in _get_csv_generator(file_path):
+        if row_index == 1:  # skip the header row
+            continue
+
+        if len(batch) % batch_size == 0:
```
annashamray, line 694: Could we use celery chunks (https://docs.celeryq.dev/en/stable/userguide/canvas.html#chunks) instead of creating batches manually?

src/openzaak/utils/tasks.py
```diff
@@ -0,0 +1,765 @@
...
+        EnkelvoudigInformatieObjectCanonical.objects.bulk_create(canonicals)
+    except IntegrityError as e:
+        for row in batch:
+            row.processed = True
+            row.comment = f"Unable to load row due to batch error: {str(e)}"
+
+        raise e
+
+    try:
+        eios = EnkelvoudigInformatieObject.objects.bulk_create(
```
annashamray, line 446: There is should be a permission check before creating a document

I also have a question - should we check that the content of the document is reachable?

src/openzaak/components/documenten/api/viewsets.py
```diff
@@ -431,6 +440,107 @@ def _zoek(self, request, *args, **kwargs):
...
+                    "uploadUrl": "https://example.com",
+                    "statusUrl": "https://example.com",
+                    "reportUrl": "https://example.com",
+                },
+            },
+            **COMMON_ERROR_RESPONSES,
+        },
+    )
+)
+class EnkelvoudigInformatieObjectImportView(ImportCreateview):
```
annashamray, line 479: all import endpoints should not be available if CMIS is enabled. Please take a look at Verzending viewset, which is also not available for CMIS

annashamray (commented):
My main question and concern is a CSV format. I think we should support bulk-create with only JSON format
* it will still allow bulk-create of the documents
* we don't need to support 2 formats of the documents, which will reduce operation risks and development time
* client always knows the shame of JSON, because it is shown in the OAS
* it's consistent with import of Catalogi component 

Yes it has a drawback - it can't allow you to import 1mln documents in 1 file, but if it's the case it should rather be a DB dump, I don't think we need to support it on application level.

annashamray:
Positive note: 
Good job with import process! I think the setup of steps and endpoints is very well thought and structured.

src/openzaak/utils/tasks.py
```diff
@@ -0,0 +1,733 @@
...
+        document_row = _import_document_row(row, row_index)
+
+        batch.append(document_row)
+
+        if not len(batch) % batch_size == 0:
+            continue
+
+        try:
+            logger.debug(f"Creating EIO's for batch {batch_number}")
+            created_batch: list[DocumentRow] = _batch_create_eios(batch)
```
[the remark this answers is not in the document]
SonnyBA: This was discussed yesterday during a call. We don't want to have EIO instances dangling around without them having a ZEIO whenever the `zaakId` was specified.

src/openzaak/utils/tasks.py
```diff
@@ -0,0 +1,791 @@
...
+        instance.identificatie = generate_unique_identification(
+            instance, "creatiedatum"
+        )
+
+    try:
+        instance.clean()
+    except ValidationError as e:
+        error_message = (
+            "A validation error occurred while validating a "
+            f"EnkelvoudigInformtatieObject on line {row_index}: \n"
```
stevenbal, line 423:
```suggestion
            f"EnkelvoudigInformatieObject on line {row_index}: \n"
```

src/openzaak/utils/tasks.py
```diff
@@ -0,0 +1,791 @@
...
+        document_row.processed = True
+
+        return document_row
+
+    file_path = document_row.bestandspad
+    path = Path(settings.IMPORT_DOCUMENTEN_BASE_DIR) / Path(file_path)
+
+    if not path.exists() or not path.is_file():
+        error_message = (
+            f"The given filepath {path} does not exist or is a file for "
```
stevenbal, line 438:
```suggestion
            f"The given filepath {path} does not exist or is not a file for "
```

src/openzaak/utils/tasks.py
```diff
@@ -0,0 +1,791 @@
...
+        )
+
+        logger.warning(error_message)
+        document_row.comment = error_message
+        document_row.processed = True
+
+        return document_row
+
+    default_dir = _get_default_path(instance.inhoud)
+    import_path = f"{default_dir}/{path.name}"
```
stevenbal, line 449: `_get_default_path` returns a `Path`, so I think this should be possible:
```suggestion
    import_path = default_dir / path.name
```

src/openzaak/utils/tasks.py
```diff
@@ -0,0 +1,791 @@
...
+            "beschrijving": self._beschrijving,
+            "indicatieGebruiksrecht": self._indicatie_gebruiksrecht,
+            "verschijningsvorm": self._verschijningsvorm,
+            "ondertekening.soort": self._ondertekening_soort,
+            "ondertekening.datum": self._ondertekening_datum,
+            "integriteit.algoritme": self._integriteit_algoritme,
+            "integriteit.waarde": self._integriteit_waarde,
+            "integriteit.datum": self._integriteit_datum,
+            "informatieobjecttype": self._informatieobjecttype,
+            "zaakId": self._zaak_id,
```
stevenbal, line 300: I see this is part of the ticket, but what does `zaakId` mean here? Is it `zaak.identificatie` or `zaak.uuid`?

src/openzaak/utils/tasks.py
stevenbal: > I also have a question - should we check that the content of the document is reachable?

I think this is done as part of `_import_document_row`? After checking if the path exists and is a file, it's copied (and if the new destination would be unreachable, I assume the copy operation would fail)

src/openzaak/utils/tasks.py
```diff
@@ -0,0 +1,791 @@
...
+    canonicals = []
+
+    for row in batch:
+        if row.instance is None:
+            continue
+
+        canonical = EnkelvoudigInformatieObjectCanonical()
+        row.instance.canonical = canonical
+
+        canonicals.append(canonical)
```
stevenbal, line 482: Can't this be done as part of `_import_document_row`, to avoid doing the same loop multiple times?

src/openzaak/utils/tasks.py
```diff
@@ -0,0 +1,791 @@
...
+    except IntegrityError as e:
+        for row in batch:
+            row.processed = True
+            row.comment = f"Unable to load row due to batch error: {str(e)}"
+
+        raise e
+
+    try:
+        eios = EnkelvoudigInformatieObject.objects.bulk_create(
+            [row.instance for row in batch if row.instance is not None]
```
stevenbal, line 495: Does this mean that not all rows have to succeed for the import to have effect? Is this intentional?

src/openzaak/utils/tasks.py
```diff
@@ -0,0 +1,791 @@
...
+    logger.info(f"Writing batch number {batch_number} to report file")
+    _write_to_file(import_instance, batch)
+
+
+LOCK_EXPIRE = 60 * (60 * 24)  # 24 hours
+
+
+@contextmanager
+def task_lock(lock_id, oid):
+    timeout_at = monotonic() + LOCK_EXPIRE - 3
```
stevenbal, line 682: Is there a specific reason we need to use `monotonic` here? Also why is the `- 3` needed? It looks a bit random :stuck_out_tongue:

src/openzaak/utils/tasks.py
```diff
@@ -0,0 +1,791 @@
...
+
+        if not row.zaak_id:
+            row.processed = True
+            row.succeeded = True if instance and instance.pk is not None else False
+            continue
+
+        # Note that ZaakInformatieObject's will not be created using
+        # `bulk_create` (see queryset MRO).
+        zaak_eio = ZaakInformatieObject(
+            zaak_id=zaak_uuids.get(row.zaak_id),
```
stevenbal, line 524: I'm not sure if this works? `zaak_id` is not a field on `ZaakInformatieObject`

src/openzaak/utils/tasks.py
```diff
@@ -0,0 +1,791 @@
...
+            )
+
+        document_row = _import_document_row(row, row_index, eio_uuids)
+
+        batch.append(document_row)
+
+        # Note that the UUID's are only used for validation
+        for row in batch:
+            if row.instance and row.instance.uuid:
+                eio_uuids.append(str(row.instance.uuid))
```
stevenbal, line 744: This loop is happening inside the loop that is appending stuff to `batch`, is that intentional? This will add a lot of duplicates to `eio_uuids` if I understand it correctly

src/openzaak/utils/views.py
```diff
@@ -113,3 +128,168 @@ def initialize_request(self, request, *args, **kwargs):
...
+        active_imports = existing_imports.filter(
+            import_type=self.import_type, status__in=ImportStatusChoices.started_choices
+        )
+
+        if active_imports:
+            raise ValidationError(
+                {
+                    "__all__": [
+                        _(
+                            "Er is een import process gaande. Probeer het later"
```
stevenbal, line 154:
```suggestion
                            "Er is een importproces gaande. Probeer het later"
```

src/openzaak/utils/views.py
```diff
@@ -113,3 +128,168 @@ def initialize_request(self, request, *args, **kwargs):
...
+        if import_instance.status != ImportStatusChoices.pending:
+            import_status = ImportStatusChoices(import_instance.status)
+            error_message = (
+                _(
+                    "Starting an import process is not possible due to the current "
+                    "status of the import: %s"
+                )
+                % import_status.label
+            )
+            raise ValidationError({"__all__": [error_message]}, code="invalid-status")
```
stevenbal, line 242: Do we need to specify `__all__` here?
```suggestion
            raise ValidationError(error_message, code="invalid-status")
```

stevenbal (changes requested):
My comments are mostly small remarks (probably also on stuff that is still WIP), but the overall implementation looks good to me :+1:

src/openzaak/utils/models.py
SonnyBA: I created the `import_data` app for the import code. `components.documenten.import` would probably clash with the current  `openzaak.components.documenten` app

src/openzaak/utils/views.py
SonnyBA: See `55b365e82119aa939c89e8fd245f7ad3312131ea`

src/openzaak/utils/tasks.py
SonnyBA: Regarding checking the content of the document, this is checked [here](https://github.com/open-zaak/open-zaak/pull/1654/files#diff-57eb99f9dc886fbdf3d28e3a5b45c37d37566f11957dc93682fc13fe71122b4aR433).

src/openzaak/utils/management/commands/create_export.py
```diff
@@ -0,0 +1,213 @@
...
+                indicatie_gebruiksrecht,
+                instance.verschijningsvorm or "",
+                instance.ondertekening["soort"],
+                instance.ondertekening["datum"],
+                instance.integriteit["algoritme"],
+                instance.integriteit["waarde"],
+                instance.integriteit["datum"],
+                informatie_object_url,
+                zaak_uuid,
+                '"foobar,foobar"',
```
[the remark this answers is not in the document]
SonnyBA: Yes this whole management command will probably be removed before merging this PR, see [previous comment](https://github.com/open-zaak/open-zaak/pull/1654#discussion_r1599690707).

src/openzaak/utils/management/commands/create_export.py
```diff
@@ -0,0 +1,213 @@
...
+        batch_size = options["batch_size"]
+        dummy_path = options["dummy_path"]
+        domain = options["domain"]
+        protocol = options["protocol"]
+        generate_uuids = options["generate_uuids"]
+
+        _dummy_path = Path(dummy_path)
+
+        if not _dummy_path.exists():
+            raise CommandError("Dummy path does not exist")
```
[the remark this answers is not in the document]
SonnyBA: If we want to keep this command, we could rewrite it so that it would make this dummy file for you

src/openzaak/utils/tasks.py
```diff
@@ -0,0 +1,791 @@
...
+            return str(self.instance.uuid)
+        return self._uuid
+
+    @property
+    def bronorganisatie(self) -> str:
+        return self._bronorganisatie
+
+    @property
+    def creatiedatum(self) -> Optional[str]:
+        return self._creatiedatum or None
```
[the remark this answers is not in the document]
SonnyBA: In this case the `EnkelvoudigInformatieObjectSerializer` requires `creatiedatum` to have an appropriate value (or `None`). The properties are basically added to perform mutations on the provided data without modifying them (so that they can be used later on). I also try not to make fields optional as it might lead to required fields be silently empty without noticing it.

src/openzaak/utils/tasks.py
```diff
@@ -0,0 +1,791 @@
...
+    row: list[str], row_index: int, existing_uuids: list[str]
+) -> DocumentRow:
+    expected_column_count = len(DocumentRow.import_headers)
+
+    request_factory = APIRequestFactory()
+    request = request_factory.get("/")
+
+    # TODO: is this possible with an invalid row count?
+    data = [*row[:expected_column_count], row_index]
+    document_row = DocumentRow(*data)
```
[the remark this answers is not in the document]
SonnyBA: It should lead to the row not passing validation of the `EnkelvoudigInformatieObject` serializer

src/openzaak/utils/tasks.py
SonnyBA: The latter, it means the `UUID` of the zaak.

src/openzaak/utils/tasks.py
SonnyBA: Before feeding all the instances to the `bulk_create`, various validation checks are done during `_import_document_row`. Because the creation of all the instances in the function run in a single transaction, all of these instances would fail whenever one of them would have a validation error and no validation was done beforehand. Hence the filtering being doing here (e.g no instance on the row means it did not pass validation) and the validation beforehand in `_import_document_row`.

src/openzaak/utils/views.py
SonnyBA: This is required when using the `400` status code error responses (see `ValidattieFoutSerializer` of `vng_api_common`). We could pick a different status code here (like `409`?) but I think the `400` error response is more appropriate here.

src/openzaak/utils/views.py
SonnyBA: This follows the behavior of the sequence diagram. Do we want to change this behavior (e.g retrieve the status whenever the import did not start yet)?

src/openzaak/components/documenten/tasks.py
```diff
@@ -0,0 +1,377 @@
...
+    import_instance.status = ImportStatusChoices.active
+    import_instance.save(update_fields=["total", "started_on", "status"])
+
+    batch: list[DocumentRow] = []
+    batch_size = settings.IMPORT_DOCUMENTEN_BATCH_SIZE
+
+    zaak_uuids = {str(uuid): id for uuid, id in Zaak.objects.values_list("uuid", "id")}
+    eio_uuids = [
+        str(uuid) for uuid in EnkelvoudigInformatieObject.objects.values_list("uuid")
+    ]
```
annashamray, line 311: I hope it won't give us OOM error. 
Let's test it with million of zaken in the test environment when the PR is merged?

src/openzaak/components/documenten/tasks.py
```diff
@@ -0,0 +1,377 @@
...
+        return document_row
+
+    default_dir = get_default_path(EnkelvoudigInformatieObject.inhoud.field)
+    import_path = default_dir / path.name
+
+    if not default_dir.exists():
+        default_dir.mkdir(parents=True)
+
+    try:
+        shutil.copy2(path, import_path)
```
annashamray, line 199: I am curious how it file copying works with our docker setup. Should we create a special volume for it in our docker file? Have you tested it with docker-compose?

src/openzaak/components/documenten/tasks.py
```diff
@@ -0,0 +1,377 @@
...
+        logger.warning(error_message)
+        document_row.comment = error_message
+        document_row.processed = True
+
+        return document_row
+
+    file_path = document_row.bestandspad
+    path = Path(settings.IMPORT_DOCUMENTEN_BASE_DIR) / Path(file_path)
+
+    if not path.exists() or not path.is_file():
```
annashamray, line 180: I suspect we will have a lot of error with the path access before it's configured correctly. So let's check the existence of this folder in the very beginning of the importing process. 

And perhaps let's stop the process if the path doesn't exist, because there is a high change that all documents will have the same error?

src/openzaak/conf/includes/base.py
```diff
@@ -775,6 +776,10 @@
     "ZAAK_EIGENSCHAP_WAARDE_VALIDATION", default=False
 )
 
+# Import settings
+IMPORT_DOCUMENTEN_BASE_DIR = config("IMPORT_DOCUMENTEN_BASE_DIR", BASE_DIR)
```
annashamray, line 780: The whole process should be documented including these env vars should be documented. Please don't forget that nice sequence diagram, which is very useful. It can be done in the separate PR though (but it's necessary before we release it)

src/openzaak/components/documenten/tests/import/test_import_documents.py
```diff
@@ -0,0 +1,817 @@
...
+        )
+
+        self.addCleanup(self.requests_mock.stop)
+
+        super().setUp()
+
+    def test_simple_import(self):
+        zaken = {1: ZaakFactory(), 2: ZaakFactory()}
+
+        import_data = {
```
annashamray, line 84: Could you please create a csv file with test data and save it into test folder, so we would have an example of what the import endpoind expects

src/openzaak/components/documenten/tests/import/test_import_create.py
```diff
@@ -0,0 +1,126 @@
...
+
+        response = self.client.post(self.url)
+
+        self.assertEqual(Import.objects.count(), 0)
+
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
+
+        response_data = response.json()
+
+        self.assertEqual(response_data["code"], "permission_denied")
```
annashamray, line 113: We don't allow django users a direct access to API

annashamray (changes requested):

src/openzaak/import_data/tests/utils.py
```diff
@@ -0,0 +1,57 @@
...
+        for file in files:
+            if file.is_file():
+                file.unlink()
+            else:
+                shutil.rmtree(file)
+
+    def remove_upload_files(self):
+        from openzaak.components.documenten.models import EnkelvoudigInformatieObject
+
+        upload_dir = get_default_path(EnkelvoudigInformatieObject.inhoud.field)
```
stevenbal, line 35: Are we sure that this is always set? I just want to avoid having `upload_dir` being `/` or `.` and deleting everything in those directories :p

stevenbal (changes requested):

src/openzaak/components/documenten/tasks.py
```diff
@@ -0,0 +1,377 @@
...
+            continue
+
+        for key, value in gegevens_groep_value.items():
+            data[f"{field}_{key}"] = value
+
+    instance = EnkelvoudigInformatieObject(**data)
+    instance.canonical = EnkelvoudigInformatieObjectCanonical()
+
+    if not instance.identificatie:
+        instance.identificatie = generate_unique_identification(
```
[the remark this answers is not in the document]
SonnyBA: It is, but `bulk_create` does not call the models `save` method ([see docs](https://docs.djangoproject.com/en/4.2/ref/models/querysets/#bulk-create))

src/openzaak/components/documenten/tasks.py
SonnyBA: Yeah I think we need to test this with test data that is similar to that of what will be used in production environments

src/openzaak/components/documenten/tests/import/test_import_documents.py
SonnyBA: I actually removed the (manually added) test files as it can take up quite some time to modify tests (e.g keep the values in the import file and the data from the DB in sync, updated header names, reorder headers/columns) whenever a field is added and or renamed in the import file.

Note that the files are still generated though through the `DocumentRowFactory` ([link](https://github.com/open-zaak/open-zaak/pull/1654/files#diff-f1bd1fc46084251b322c10fdc624ca950523e3df317b0aa157c0fdde9fdb4659R218)).

src/openzaak/conf/includes/base.py
SonnyBA: Discussed today, will be added through a separate PR

src/openzaak/components/documenten/openapi.yaml
```diff
@@ -3838,6 +3838,381 @@ paths:
                 resources exact dezelfde ETag hebben, dan zijn deze resources identiek
                 aan elkaar. Je kan de ETag gebruiken om caching te implementeren.
           description: No response body
+  /import/{uuid}/report:
```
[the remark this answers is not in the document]
SonnyBA: Setting `SORT_OPERATIONS` to `False` will have impact on the ordering of other components too. Do we want to apply that change? Setting the option to `False` does sort the endpoints in order they are defined.

Another option is to include an attribute on the endpoints which specifies the order and use this in the `get_api_endpoints` [method](https://github.com/tfranzel/drf-spectacular/blob/master/drf_spectacular/generators.py#L24) (where the default behavior of the sorting also takes place) but it feels a bit overkill to me.

SonnyBA:
@annashamray @stevenbal I preserved the commits for now to make reviewing the changes a bit easier. Before merging the PR I'll squash the commits.

src/openzaak/components/documenten/tasks.py
SonnyBA: Note that the current setup does work out of the box for any setup although all the import metadata files would have to be stored in the `BASE_DIR` (the default of `IMPORT_DOCUMENTEN_BASE_DIR`). Having a dedicated directory for imports does feel a bit more organized though.

annashamray: Let's put everything into one docker-compose file, at least when it's possible, so it would be similar to real deployment configuration

src/openzaak/import_data/serializers.py
```diff
@@ -0,0 +1,67 @@
...
+from rest_framework.serializers import HyperlinkedModelSerializer
+
+from openzaak.import_data.models import Import, ImportStatusChoices
+
+
+class ImportSerializer(HyperlinkedModelSerializer):
+    status = SerializerMethodField(source="get_status")
+
+    @extend_schema_field(OpenApiTypes.STR)
+    def get_status(self, instance):
```
annashamray, line 17: Could you please explain why SerializerMethodField is needed here?

annashamray (changes requested):
Please add configuration needed for bulk_import in `docker-compose.yml`

src/openzaak/components/documenten/api/viewsets.py
```diff
@@ -431,6 +443,136 @@ def _zoek(self, request, *args, **kwargs):
...
+@extend_schema_view(
+    retrieve=extend_schema(
+        operation_id="enkelvoudiginformatieobject_import_report",
+        summary=_("Het reportage bestand van een IMPORT downloaden."),
+        description=mark_experimental(
+            "Het reportage bestand downloaden van een IMPORT. Dit bestand is alleen "
+            "beschikbaar indien de IMPORT is afgerond (ongeacht het resultaat). "
+            "Deze actie is niet beschikbaar wanneer de `CMIS_ENABLED` optie is "
+            "ingeschakeld. De volgende kolommen zijn te vinden in het rapportage "
+            f"bestand: {_get_report_headers()}. Voor deze actie is een APPLICATIE"
```
stevenbal, line 553: Currently there is no space between `APPLICATIE` and `nodig`
```suggestion
            f"bestand: {_get_report_headers()}. Voor deze actie is een APPLICATIE "
```

stevenbal (approved):
Minor comment remaining from me about a description in the API spec, other than that it looks good to me (once Anna's comments are addressed)

stevenbal:
@SonnyBA After reviewing the docs I have a question though:

How will it work in practice to get the files associated with each document where they need to be? If I understand it correctly, the idea was to make these endpoints so that it can be done without having to ask devops to run the import. But the actual files need to be on the server and mounted as a volume to the container before Open Zaak can copy them to its internal storage, right? And we only upload a metadata file with paths, not the actual files themselves

src/openzaak/components/documenten/api/viewsets.py
```diff
@@ -431,6 +443,132 @@ def _zoek(self, request, *args, **kwargs):
...
+            "ingeschakeld. Voor deze actie is een APPLICATIE nodig met "
+            "`heeft_alle_autorisaties` ingeschakeld."
+        ),
+        request=OpenApiTypes.NONE,
+    )
+)
+class EnkelvoudigInformatieObjectImportView(ImportCreateview):
+    import_type = ImportTypeChoices.documents
+
+    required_scopes = {}
```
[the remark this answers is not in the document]
SonnyBA: Its for the schema generation:
```
  File "/home/sonny/development/open-zaak/src/openzaak/utils/schema.py", line 84, in get_auth
    scopes = get_required_scopes(self.view.request, self.view)
  File "/home/sonny/development/open-zaak/env/lib/python3.10/site-packages/vng_api_common/permissions.py", line 26, in get_required_scopes
    raise ImproperlyConfigured(
django.core.exceptions.ImproperlyConfigured: The View(Set) must have a `required_scopes` attribute

```

src/openzaak/import_data/serializers.py
SonnyBA: For some reason, without this change, the schema generator thinks `ImportStatus` conflicts with the existing `StatusEnum` and generates a new enum, see [diff](https://gist.github.com/SonnyBA/2325c1c25060f765d960cb202d2f6000).

src/openzaak/components/documenten/tasks.py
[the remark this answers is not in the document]
SonnyBA: Good catch! Should be fixed through https://github.com/open-zaak/open-zaak/pull/1654/commits/62071a34e600be91984af837e66de9dd75160f56.

src/openzaak/components/documenten/tests/import/test_import_documents.py
SonnyBA: See https://github.com/open-zaak/open-zaak/pull/1654/commits/3fff740c8c55fb48a7bc5135c0a6c9747a4366ce

annashamray (changes requested):
Thank you for adding csv files to the tests, I used them to test 

Does import work for you with docker-compose? 

I've managed to start the process but the endpoint to retrieve a report runs forever and doesn't return the response
Does it work for you?

src/openzaak/import_data/models.py
```diff
@@ -0,0 +1,140 @@
...
+        return {
+            cls.documents: ComponentTypes.drc,
+        }
+
+    @classmethod
+    def get_component_from_choice(cls, choice):
+        return cls.component_mapping[choice]
+
+
+class Import(models.Model):
```
annashamray, line 55: Could you please add this model into `fixtures/default_admin_index.json` ? So it will appear in the admin

src/openzaak/import_data/models.py
```diff
@@ -0,0 +1,140 @@
...
+    uuid = models.UUIDField(default=uuid.uuid4)
+    status = models.CharField(choices=ImportStatusChoices.choices, max_length=30)
+    import_type = models.CharField(
+        verbose_name=_("Type import"), choices=ImportTypeChoices.choices, max_length=30
+    )
+
+    import_file = PrivateMediaFileField(
+        verbose_name=_("Import metadata bestand"),
+        upload_to="import/import-files/",
```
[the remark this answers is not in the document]
annashamray: I've just realized that it would be inside `private-media` folder, so no need for a separate volume

annashamray:
The thing that we overlooked is sending notifications about imported documents
Tbh I don't know what to do with it

@SonnyBA Could you please create an issue for it with `triage` label? We will discuss it in the team

src/openzaak/components/documenten/tasks.py
```diff
@@ -0,0 +1,395 @@
...
+def _import_document_row(
+    row: list[str],
+    row_index: int,
+    identifier: str,
+    existing_uuids: list[str],
+    zaak_uuids: dict[str, int],
+) -> DocumentRow:
+    expected_column_count = len(DocumentRow.import_headers)
+
+    request_factory = APIRequestFactory()
```
annashamray, line 49: I have an error after uploading csv file in celery container:

```
[2024-06-14 14:14:26,300: ERROR/ForkPoolWorker-4] Task openzaak.components.documenten.tasks.import_documents[8d584fcb-590a-4d45-8242-dcfd70dca187] raised unexpected: DisallowedHost("Invalid HTTP_HOST header: 'testserver'. You may need to add 'testserver' to ALLOWED_HOSTS.")
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/site-packages/celery/app/trace.py", line 451, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/usr/local/lib/python3.10/site-packages/celery/app/trace.py", line 734, in __protected_call__
    return self.run(*args, **kwargs)
  File "/app/src/openzaak/import_data/utils.py", line 207, in wrapped
    return func(*args, **kwargs)
  File "/app/src/openzaak/components/documenten/tasks.py", line 338, in import_documents
    document_row = _import_document_row(
  File "/app/src/openzaak/components/documenten/tasks.py", line 88, in _import_document_row
    if not eio_serializer.is_valid():
  File "/usr/local/lib/python3.10/site-packages/rest_framework/serializers.py", line 227, in is_valid
    self._validated_data = self.run_validation(self.initial_data)
  File "/usr/local/lib/python3.10/site-packages/rest_framework/serializers.py", line 426, in run_validation
    value = self.to_internal_value(data)
  File "/usr/local/lib/python3.10/site-packages/rest_framework/serializers.py", line 483, in to_internal_value
    validated_value = field.run_validation(primitive_value)
  File "/usr/local/lib/python3.10/site-packages/django_loose_fk/drf.py", line 159, in run_validation
    url = super().run_validation(*args, **kwargs)
  File "/usr/local/lib/python3.10/site-packages/rest_framework/fields.py", line 751, in run_validation
    return super().run_validation(data)
  File "/usr/local/lib/python3.10/site-packages/rest_framework/fields.py", line 548, in run_validation
    self.run_validators(value)
  File "/usr/local/lib/python3.10/site-packages/rest_framework/fields.py", line 560, in run_validators
    validator(value, self)
  File "/app/src/openzaak/utils/validators.py", line 134, in __call__
    super().__call__(value, serializer_field)
  File "/app/src/openzaak/utils/serializer_fields.py", line 68, in __call__
    super().__call__(url, serializer_field)
  File "/usr/local/lib/python3.10/site-packages/django_loose_fk/drf.py", line 88, in __call__
    host = serializer_field.context["request"].get_host()
  File "/usr/local/lib/python3.10/site-packages/django/http/request.py", line 150, in get_host
    raise DisallowedHost(msg)
django.core.exceptions.DisallowedHost: Invalid HTTP_HOST header: 'testserver'. You may need to add 'testserver' to ALLOWED_HOSTS.
```

I think it's because of this factory. Maybe it would be better to pass `request` to the celery task? Or at least the host

SonnyBA:
See https://github.com/orgs/maykinmedia/projects/15/views/1?pane=issue&itemId=67477191 for handling notifications

annashamray:
@SonnyBA It doesn't work for me, perhaps request object can't be a task argument
`upload` endpoint fails with error ,

```
2024-06-17 10:17:40,839 ERROR vng_api_common.views  Object of type Request is not JSON serializable
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/site-packages/kombu/serialization.py", line 39, in _reraise_errors
    yield
  File "/usr/local/lib/python3.10/site-packages/kombu/serialization.py", line 210, in dumps
    payload = encoder(data)
  File "/usr/local/lib/python3.10/site-packages/kombu/utils/json.py", line 68, in dumps
    return _dumps(s, cls=cls or _default_encoder,
  File "/usr/local/lib/python3.10/json/__init__.py", line 238, in dumps
    **kw).encode(obj)
  File "/usr/local/lib/python3.10/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/local/lib/python3.10/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/usr/local/lib/python3.10/site-packages/kombu/utils/json.py", line 58, in default
    return super().default(o)
  File "/usr/local/lib/python3.10/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type Request is not JSON serializable

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.10/site-packages/rest_framework/views.py", line 506, in dispatch
    response = handler(request, *args, **kwargs)
  File "/app/src/./openzaak/components/documenten/api/viewsets.py", line 524, in create
    import_documents.delay(import_instance.pk, request)
  File "/usr/local/lib/python3.10/site-packages/celery/app/task.py", line 425, in delay
    return self.apply_async(args, kwargs)
  File "/usr/local/lib/python3.10/site-packages/celery/app/task.py", line 575, in apply_async
    return app.send_task(
  File "/usr/local/lib/python3.10/site-packages/celery/app/base.py", line 788, in send_task
    amqp.send_task_message(P, name, message, **options)
  File "/usr/local/lib/python3.10/site-packages/celery/app/amqp.py", line 510, in send_task_message
    ret = producer.publish(
  File "/usr/local/lib/python3.10/site-packages/kombu/messaging.py", line 166, in publish
    body, content_type, content_encoding = self._prepare(
  File "/usr/local/lib/python3.10/site-packages/kombu/messaging.py", line 254, in _prepare
    body) = dumps(body, serializer=serializer)
  File "/usr/local/lib/python3.10/site-packages/kombu/serialization.py", line 209, in dumps
    with _reraise_errors(EncodeError):
  File "/usr/local/lib/python3.10/contextlib.py", line 153, in __exit__
    self.gen.throw(typ, value, traceback)
  File "/usr/local/lib/python3.10/site-packages/kombu/serialization.py", line 43, in _reraise_errors
    reraise(wrapper, wrapper(exc), sys.exc_info()[2])
  File "/usr/local/lib/python3.10/site-packages/kombu/exceptions.py", line 21, in reraise
    raise value.with_traceback(tb)
  File "/usr/local/lib/python3.10/site-packages/kombu/serialization.py", line 39, in _reraise_errors
    yield
  File "/usr/local/lib/python3.10/site-packages/kombu/serialization.py", line 210, in dumps
    payload = encoder(data)
  File "/usr/local/lib/python3.10/site-packages/kombu/utils/json.py", line 68, in dumps
    return _dumps(s, cls=cls or _default_encoder,
  File "/usr/local/lib/python3.10/json/__init__.py", line 238, in dumps
    **kw).encode(obj)
  File "/usr/local/lib/python3.10/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/local/lib/python3.10/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/usr/local/lib/python3.10/site-packages/kombu/utils/json.py", line 58, in default
    return super().default(o)
  File "/usr/local/lib/python3.10/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
kombu.exceptions.EncodeError: Object of type Request is not JSON serializable
```

docker/nginx/default.conf
```diff
@@ -2,7 +2,12 @@ server {
     listen       80;
     server_name  localhost;
 
-    location /openzaak {
-        proxy_pass   http://web:8000;
+    location /private-media {
+        alias /private-media;
```
annashamray, line 6:
```suggestion
        internal;
        alias /private-media;
```

As discussed let's make it more secure

src/openzaak/import_data/admin.py
```diff
@@ -0,0 +1,25 @@
+# SPDX-License-Identifier: EUPL-1.2
+# Copyright (C) 2019 - 2024 Dimpact
+from django.contrib import admin
+
+from openzaak.import_data.models import Import
+
+
+@admin.register(Import)
+class ImportAdmin(admin.ModelAdmin):
```
annashamray, line 9: please add `PrivateMediaMixin` so urls to download files and reports will work. `EnkelvoudigInformatieObjectAdmin` can be served as inspiration

annashamray (changes requested):
I confirm that API works with docker-compose! 
Let's make it a little more secure and we are good to go!

annashamray (approved):
image
I'll approve because the process works in API and admin urls also work
But when I upload documents without identificatie, all of them are assigned the same identificatie (you can see it in the printscreen) 
We don't have constraints for it for documenten, so I guess it's not critical, but I think it should be taken care of, because it will raise questions. 
So I suggest to merge this PR and fix it in the follow-up, therefore approve

SonnyBA:
@annashamray the last commit should fix your mentioned issue:

image

I'm merging this PR so that the [documentation build](https://github.com/open-zaak/open-zaak/pull/1679) can pass because of the links to the specification

PR #1658, opened 2024-05-16
:sparkles: [#1619] changed user agent to 'Open Zaak'

### LINKED ISSUE #1619: Remove version number of User Agent from outgoing requests (zgw-consumers / requests)
### Thema / Theme

Other

Taiga AMS #13

Currently outgoing requests have an automatic User Agent which is set by requests (eg: 'python-requests/2.25.0'). According to client this is blocked by the default Azure WAF configuration.

This was flagged as an information disclosure issue, all be it one with a limited security impact in my eyes for the use case of Open Zaak's outgoing requests. But even so, communicating version numbers in requests and responses in general can facilitate attackers so is frowned upon in general.

My recommendation would be to both set this on a per-component basis (UA: 'Open Zaak') and to remove the version number in our default-project requests wrapper (use UA: 'python-requests'). I would advise to not make this configurable as there's no need to.

Do note that 'python-requests' (also without version number) is explicitly set as a mod-security rule. So using a default like 'python' would avoid future problems for new components

https://github.com/SpiderLabs/owasp-modsecurity-crs/blob/v3.3/dev/rules/scripting-user-agents.data

https://stackoverflow.com/questions/52497739/how-to-set-requests-user-agent-header-globally

Low-prio as client has a workaround, but this is likely to come up in future audits or integration discussions

_No response_

_No response_

bart-maykin:
Fixes #1619

**Changes**

changed User-Agent from requests lib to "Open Zaak"

### DIFF
--- a/src/openzaak/utils/apps.py
+++ b/src/openzaak/utils/apps.py
@@ -1,10 +1,12 @@
 # SPDX-License-Identifier: EUPL-1.2
 # Copyright (C) 2019 - 2020 Dimpact
 from django.apps import AppConfig
+from django.conf import settings
 from django.db import models
 from django.db.models.signals import post_migrate
 
 from django_loose_fk.virtual_models import HANDLERS, FKHandler
+from requests import utils
 from rest_framework import serializers
 
 
@@ -22,6 +24,8 @@ class UtilsConfig(AppConfig):
         )
         from .signals import update_admin_index
 
+        utils.default_user_agent = default_user_agent
+
         post_migrate.connect(update_admin_index, sender=self)
 
         # register FKOrServiceUrlField drf field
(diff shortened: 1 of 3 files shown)


Coperh (approved):
Yeah, seems fine.

Fix the Isort error and you can merge.

src/openzaak/utils/apps.py
```diff
@@ -33,3 +37,12 @@ def ready(self):
 
         # register one-to-one field (for multi-table inheritance of Zaak)
         HANDLERS[models.OneToOneField] = FKHandler
+
+
+def default_user_agent(name=settings.USER_AGENT):
```
stevenbal, line 42: The `name` argument won't really be used, so you could just return `settings.USER_AGENT` directly (or do `utils.default_user_agent = lambda settings.USER_AGENT`, but it's not a big deal, so I'll approve

stevenbal (approved):

PR #1677, opened 2024-06-11
:sparkles:[#1256] show conflicting identification number in error mes…

### LINKED ISSUE #1256: API create/update: Show conflicting identification number when returning a "identificatie-niet-uniek" error

Organisatie / Organization: Maykin and gemeente SWF
Naam / Name: Alex and @EduardWitteveen  

This applies at least for the Document API, but most likely also for the Zaak API and Besluiten API.

Currently it isn't clear from the ValidationError raised via the API what the conflicting identification is from a "identificatie-niet-uniek" error. Without looking into Sentry for the identifier and without logging on the client side it isn't possible to determine for which case this error was raised.

Error message as currently logged:

ValidationError {'identificatie': [ErrorDetail(string='Deze identificatie bestaat al voor deze bronorganisatie', code='identificatie-niet-uniek')]}

It would be best to have the error message show the conflicting identification when returning the error via the API:

ValidationError {'identificatie': [ErrorDetail(string='Deze identificatie (SWF-12345) bestaat al voor deze bronorganisatie', code='identificatie-niet-uniek')]}

This allows both the client to investigate directly using only the errormessage, and in OpenZaak the conflicting zaak can easily be found in the application log. 

This is an issue to improve maintainability and improve the developer-friendlyness of the APIs.

Coperh:
…sage

Fixes #1256

**Changes**

[Describe the changes here]

### DIFF
--- a/src/openzaak/utils/validators.py
+++ b/src/openzaak/utils/validators.py
@@ -15,7 +15,11 @@ from rest_framework.validators import (
 )
 from vng_api_common.oas import obj_has_shape
 from vng_api_common.utils import get_resource_for_path, get_uuid_from_path
-from vng_api_common.validators import IsImmutableValidator, URLValidator
+from vng_api_common.validators import (
+    IsImmutableValidator,
+    UniekeIdentificatieValidator as _UniekeIdentificatieValidator,
+    URLValidator,
+)
 
 from openzaak.api_standards import APIStandard
 from openzaak.components.documenten.models import EnkelvoudigInformatieObject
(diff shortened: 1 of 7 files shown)


stevenbal (approved):

src/openzaak/utils/validators.py
```diff
@@ -268,3 +272,18 @@ def __call__(self, url: str):
...
+        "Deze identificatie ({identificatie}) bestaat al binnen de organisatie"
+    )
+
+    @property
+    def message(self):
+        return self.partial_message.format(identificatie=self.identificatie)
+
+    def __call__(self, attrs: dict, serializer):
+        self.identificatie = attrs.get(self.identificatie_field)
+        return super().__call__(attrs, serializer)
```
sergei-maertens, line 289: IIRC this is subject to a race condition because the same validator instance can and will be used for different requests (see https://www.django-rest-framework.org/community/3.11-announcement/#validator-default-context)

PR #1679, opened 2024-06-12
[#1596] add import documentation

SonnyBA:
Partially fixes #1654 

**Changes**

Adds documentation about the bulk import functionality for `EnkelvoudigInformatieObject`'s

Should be merged after https://github.com/open-zaak/open-zaak/pull/1654

### DIFF
--- a/docs/installation/reference/import.rst
+++ b/docs/installation/reference/import.rst
@@ -1,0 +1,176 @@
+.. _installation_reference_containers:
+
+Importing documenten
+=======================
+
+Open Zaak data can be supplied in various ways. One can ofcourse make API calls
+for example to create `EnkelvoudigInformatieObject`'s but whenever a large amount
+of data needs to be created in a relatively short time period this can become
+a not-so-pleasant experience. To prevent having to deal with these kinds of situations
+Open Zaak users can make use of the import functionality for `EnkelvoudigInformatieObject`'s.
+Open Zaak exposes several API endpoints to help aid the user to import larger amounts
+of `EnkelvoudigInformatieObject`'s.
+
+Process
+-------
+
+The import process consists of several steps with each step having a API endpoint.
+The process consists of the following steps:
+
+1. Creating the `Import`
+2. Uploading the `Import` metadata file
+3. (Optionally) Retrieving the status of the `Import`
+4. (Optionally) Retrieving the report file of the `Import`
+
+**Permissions**
+
+Importing `EnkelvoudigInformatieObjecten` is not possible for all authenticated users
+for Open Zaak. The reasoning behind this is that not all users for example have
+the rights to create `EnkelvoudigInformatieObjecten` with certain `InformatieObjectType`'s.
+Various permissions exist within Open Zaak and therefore to make use of the import
+functionality, users only with the `heeft_alle_autorisaties` set to `True` are allowed
+to perform any import operation.
+
+**Creating an Import**
+
+The first step in the import process is that of making an `Import` resource.
+This is done through a `POST` request. The request body of this request can be empty
+for this step.
+
+Creating an `Import` is only possible whenever no other `Import` instance exist
+with the statusses `pending` or `active`.
+
+Whenever no other import is pending or active, the endpoint will provide the user
+three urls; an url to upload an import metadata file, an url to retrieve the status
+of an import and another url for downloading a report of the import. These three
+urls provide the user the ability to progress further in the import process.
+
+The next step in the import process, after creating an `Import` instance, is providing
+a metadata file which starts the actual importing of `EnkelvoudigInformatieObject`'s.
+
+**Starting an Import**
+
+After creating an `Import` instance, users can provide an import metadafile. This
+should be a csv which consists of rows with data needed to create an `EnkelvoudigInformatieObject`.
+The data this csv should contain is rougly the same as that is needed for creating
+an `EnkelvoudigInformatieObject` through a "regular" API call with some exceptions.
+For more details about the format and the requirements of the csv file, the API
+documentation should be consulted. The request for this endpoint should be a `POST`
+request containing the csv data in its request body.
+
+Whenever the csv contains invalid and or missing headers the import process will
+not be started. Whenever headers are missing the error response will contain those
+headers.
+
+One column which is required for each row in the csv file is the path to the
+file which will be imported and will be assosiated to the `EnkelvoudigInformatieObject`.
+This should be a relative path from the path configured in `IMPORT_DOCUMENTEN_BASE_DIR`.
+
+Before starting an import, the directory configured through `IMPORT_DOCUMENTEN_BASE_DIR`,
+is accessed to check if this directory exists (and if it actually is a directory).
+Whenever this is not the case the import also will not start.
+
+Just like in the step to create the `Import` instance, no other `pending` or `active`
+import instances can be active before starting the specified `Import`.
+
+Whenever this request passes validation (with the above mentioned checks) the
+actual import process is started through a background task.
+
+**Retrieving the status**
+
+Whenever an `Import` instance has the status `pending`, `active`, `finished` or
+`error` and the import process is started, the status of that `Import` can be
+retrieved. This endpoint can be called through a `GET` request. The data of the
+response contains information for example about the total amount of rows the
+import metadata file has and the amount of rows the `Import` at that time has
+processed. See the API documentation of this endpoint for more information.
+
+**Retrieving a report file**
+
+Whenever an `Import` instance has the status `finished` or `error` an report
+file can be downloaded of the process. This report is a csv almost identical to
+that of the provided metadata file with the additional columns which specify wether
+an row is imported or not and if there any comments about the row.
+
+Import behavior
+----------------
+
+The actual import process, which is a background task, imports each row in
+batches (configured through `IMPORT_DOCUMENTEN_BATCH_SIZE`). During each batch
+several error situations can occure. For example an existing
+`EnkelvoudigInformatieObject` can exist with the specified `uuid`. This can be
+considered a validation error and will not cause other rows in the specific batch
+not to be imported.
+
+Another situation that can occure is for example a database connection loss. This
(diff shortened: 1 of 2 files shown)


Coperh (changes requested):
A few typos and a few suggested changes.

I think it is a bit wordy. 

Lack examples or should somehow link to the API docs. Not sure if that is in documentarian.

I think all times you use `Whenever`,  you should probably just use `when`

docs/installation/reference/import.rst
```diff
@@ -0,0 +1,177 @@
...
+
+Whenever the csv contains invalid and or missing headers the import process will
+not be started. Whenever headers are missing the error response will contain those
+headers.
+
+One column which is required for each row in the csv file is the path to the
+file which will be imported and will be assosiated to the `EnkelvoudigInformatieObject`.
+This should be a relative path from the path configured in `IMPORT_DOCUMENTEN_BASE_DIR`.
+
+Before starting an import, the directory configured through `IMPORT_DOCUMENTEN_BASE_DIR`,
```
stevenbal, line 70: Could you add `IMPORT_DOCUMENTEN_BASE_DIR` and `IMPORT_DOCUMENTEN_BATCH_SIZE` to the environment variable configuration (`docs/installation/config/env_config.rst`) and add references for them here? 

```rst
.. _import_documenten_base_dir:
* ``IMPORT_DOCUMENTEN_BASE_DIR``: ...
```

and then refer to them in this file like this:

```rst
:ref:`IMPORT_DOCUMENTEN_BASE_DIR <import_documenten_base_dir>`
```

Not 100% sure if this example I gave works right away, but more on this can be found here https://docs.readthedocs.io/en/stable/guides/cross-referencing-with-sphinx.html#cross-referencing-using-roles

I also think I would maybe reference these envvars in a separate section before ``Process`` (maybe something like ``Prerequisites``) and mention that these need to be set before this functionality can be used (though I'm not sure if they need to be set explicitly or if the defaults work too), instead of first mentioning them in the process description.

docs/installation/reference/import.rst
```diff
@@ -0,0 +1,177 @@
...
+is accessed to check if this directory exists (and if it actually is a directory).
+Whenever this is not the case the import also will not start.
+
+Just like in the step to create the `Import` instance, no other `pending` or `active`
+import instances can be active before starting the specified `Import`.
```
stevenbal, line 75: I'm not 100% sure when an `Import` actually gets the status `pending` or `active`, is this after the file is uploaded?
```suggestion
Just like in the step to create the `Import` instance, it is not possible to start an ``Import`` if there is another ``Import`` in progress.
```

docs/installation/reference/import.rst
```diff
@@ -0,0 +1,177 @@
...
+process (the background task). In this situation the database cannot be reached and
+the data `Import` instance will be out-of-sync. Logging however is done and the
+report file will have comments for all rows in that specified batch.
+
+During the import, files are copied to the import path that the "regular" endpoint
+for `EnkelvoudigInformatieObject`'s would do so too. Files are however overwritten
+whenever they already exist.
+
+Configuration
+--------------
```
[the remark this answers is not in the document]
stevenbal: I agree it would be better to move this section before `Process`.

Though I'm not sure what to do with the help texts for the envvars. Should we put them here or add them in `env_config.rst` and refer to them here? @annashamray thoughts?

docs/installation/reference/import.rst
```diff
@@ -0,0 +1,177 @@
...
+
+Another situation that can occure is for example a database connection loss. This
+causes an situation where the import process cannot proceed and will stop the import
+process (the background task). In this situation the database cannot be reached and
+the data `Import` instance will be out-of-sync. Logging however is done and the
+report file will have comments for all rows in that specified batch.
+
+During the import, files are copied to the import path that the "regular" endpoint
+for `EnkelvoudigInformatieObject`'s would do so too. Files are however overwritten
+whenever they already exist.
```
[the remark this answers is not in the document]
stevenbal: I would maybe rephrase this to something like:

```
If a row does not cause any validation errors, the file associated with that row will be copied to Open Zaak's storage. If the file already exists there, it will be overwritten.
```

docs/installation/reference/import.rst
```diff
@@ -0,0 +1,177 @@
...
+
+Configuration
+--------------
+
+Settings related to the import functionaly are:
+- `IMPORT_DOCUMENTEN_BASE_DIR`
+- `IMPORT_DOCUMENTEN_BATCH_SIZE`
+
+`IMPORT_DOCUMENTEN_BASE_DIR` is used to determine the absolute import path for each
```
stevenbal, line 123: Had a question related to this which I posted on the PR with the code: https://github.com/open-zaak/open-zaak/pull/1654#issuecomment-2165285289

docs/installation/reference/import.rst
```diff
@@ -0,0 +1,177 @@
...
+report file will have comments for all rows in that specified batch.
+
+During the import, files are copied to the import path that the "regular" endpoint
+for `EnkelvoudigInformatieObject`'s would do so too. Files are however overwritten
+whenever they already exist.
+
+Configuration
+--------------
+
+Settings related to the import functionaly are:
```
stevenbal, line 119:
```suggestion
Environment variables related to the import functionality are:
```

docs/installation/reference/import.rst
```diff
@@ -0,0 +1,177 @@
...
+
+`IMPORT_DOCUMENTEN_BASE_DIR` is used to determine the absolute import path for each
+row in the import metadata file. All file paths specified in the import metadata
+file should be relative to the directory specified for this setting.
+By default this is the same directory as the projects directory (`BASE_DIR`) and
+can either be configured through an environment variable or directly through the
+django's settings file being used.
+
+`IMPORT_DOCUMENTEN_BATCH_SIZE` is used to split the amount of data which will be
+processed at a time. This setting was added for performance reasons and can be
```
stevenbal, line 131:
```suggestion
`IMPORT_DOCUMENTEN_BATCH_SIZE` is the number of rows that will be processed at a time. This setting was added for performance reasons and can be
```

docs/installation/reference/import.rst
```diff
@@ -0,0 +1,177 @@
...
+tweaked accordingly through an environment variable or directly through the django's
+setting file being used.
```
stevenbal, line 133: I wouldn't mention this, because this is not documentation for developers
```suggestion
```

stevenbal (changes requested):

docs/installation/reference/import.rst
```diff
@@ -0,0 +1,193 @@
...
+``pending``.
+
+See the `API documentation for creating an Import`_ for more details.
+
+**Starting an Import**
+
+After creating an `Import` instance, users can upload an import metada file. This
+should be a CSV which consists of rows with data needed to create an
+``EnkelvoudigInformatieObject``. The data this CSV should contain is roughly the
+same as the data needed for creating an ``EnkelvoudigInformatieObject`` through
```
Coperh, line 76:
```suggestion
should be a CSV which consists of rows with the data needed to create an
``EnkelvoudigInformatieObject``. This data is roughly the
same as the data needed for creating an ``EnkelvoudigInformatieObject`` through
```

I think this sounds more _"proper"_ (not just any data) but not really necessary.

Also an edit to make the next line more concise

docs/installation/reference/import.rst
```diff
@@ -0,0 +1,193 @@
...
+``Import`` instance is changed to ``active``.
+
+See the `API documentation for starting an Import`_ for more details.
+
+**Retrieving the status**
+
+The status of the ``Import`` can be retrieved when an import process is started
+and has the status ``pending``, ``active``, ``finished`` or ``error``. This endpoint
+can be called through a `GET` request. The data of the response contains
+information for example about the total amount of rows the import metadata file
```
Coperh, line 110:
```suggestion
information, for example, about the total amount of rows the import metadata file
```

docs/installation/reference/import.rst
```diff
@@ -0,0 +1,193 @@
...
+not be started and the error response will contain any missing headers.
+
+The ``bestandspad`` column, which is required for each row in the CSV file,
+is the path to the file which will be imported and will be assosciated to the
+``EnkelvoudigInformatieObject``. This should be a relative path from the directory
+configured in ``IMPORT_DOCUMENTEN_BASE_DIR``.
+
+The import will only start if the ``IMPORT_DOCUMENTEN_BASE_DIR`` setting is set
+correctly. Unknown directories or the ``IMPORT_DOCUMENTEN_BASE_DIR`` leading to a
+path which is actually a file are examples of incorrect configurations. An
```
Coperh, line 92:
```suggestion
a file instead of a directory are examples of incorrect configurations. An
```

Coperh (changes requested):
A typo and a few other small suggestions but otherwise good.

SonnyBA:
The build is failing at this point because the actual API documentation for the bulk import is not available yet. Do we want to change these URLs to an existing URL (use a dummy URL to let this build pass) or should we merge it like this?

annashamray (changes requested):
Please don't use `test.openzaak.nl` urls as links to OAS
It's a test deployment for internal (mostly) purposes. If it's ZGW standard we can use links to ZGW standards (like here - https://open-zaak.readthedocs.io/en/stable/api/index.html)
But since it's an experimental feature available only in Open Zaak could you use redocly urls with openzaak documenten OAS? 

It could also help with fixing the CI build

docs/installation/reference/import.rst
```diff
@@ -0,0 +1,192 @@
+.. _installation_reference_import:
+
+===========================
+Importing documents in bulk
+===========================
```
annashamray, line 5: This comment is optional. 
But what do you think about adding a sequence diagram from GH issue to the documentation?

Coperh (approved):
Fine on my end

annashamray (approved):
:+1:

PR #1681, opened 2024-06-13
 ✨ [#1648] Option to keep same UUIDs when importing Zaaktype

### LINKED ISSUE #1648: Ensure UUIDs remain the same when importing zaaktypes and other entities
### Thema / Theme

Admin

Currently we have the option to keep the UUIDs during importing, however this feels weird and complicates #1611 significantly.

With this issue we ensure that UUIDs are always kept the same when importing zaaktypes and other related entities (scoped to the Catalogi API)

_No response_

_No response_

stevenbal:
Fixes #1648 

**Changes**

* Option to keep same UUIDs when importing Zaaktype

### DIFF
--- a/src/openzaak/components/catalogi/tests/admin/test_import_export_zaaktype.py
+++ b/src/openzaak/components/catalogi/tests/admin/test_import_export_zaaktype.py
@@ -4,6 +4,7 @@ import io
 import zipfile
 from datetime import datetime
 from unittest.mock import patch
+from uuid import UUID
 
 from django.contrib.auth.models import Permission
 from django.contrib.sites.models import Site
@@ -13,7 +14,6 @@ from django.urls import reverse
 from django.utils.html import format_html
 from django.utils.translation import gettext as _
 
-import requests_mock
 from django_webtest import TransactionWebTest, WebTest
 from maykin_2fa.test import disable_admin_mfa
 from zgw_consumers.constants import APITypes, AuthTypes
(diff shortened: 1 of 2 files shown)


src/openzaak/components/catalogi/tests/admin/test_import_export_zaaktype.py
```diff
@@ -1871,178 +1752,6 @@ def setUp(self):
         conf.allowed_years = [2020]
         conf.save()
 
-    def test_import_zaaktype_already_exists(self, *mocks):
```
stevenbal, line 1874: I removed this because I think it's no longer relevant after the change to always preserve UUIDs

src/openzaak/components/catalogi/tests/admin/test_import_export_zaaktype.py
```diff
@@ -2149,101 +1858,6 @@ def test_import_zaaktype_invalid_eigenschap(self, *mocks):
         self.assertEqual(ZaakType.objects.count(), 0)
         self.assertEqual(Eigenschap.objects.count(), 0)
 
-    def test_import_zaaktype_informatieobjectype_overlapping(self, *mocks):
```
stevenbal, line 2152: I'm not sure what this test was testing to begin with, it's no longer possible now to export/import the same data without removing the existing data, so I removed it

stevenbal:
@annashamray @joeribekker tt turns out that for catalogi import it was possible to use the original UUIDs or generate new UUIDs, but for zaaktype imports it always generated new UUIDs. This change does mean that it's no longer possible to use export/import as a way of "copying" zaaktypes on the same environment. Do you think this is an issue?

src/openzaak/components/catalogi/admin/utils.py
```diff
@@ -204,9 +204,6 @@ def import_zaaktype_for_catalogus(
                     for old, new in besluittypen_uuid_mapping.items():
                         data = data.replace(old, str(new.uuid))
```
annashamray, line 205: Is it needed now? (and the similar lines above with `ZaakTypeInformatieObjectType` )

joeribekker:
Hmm, Alex pushed this too hard I think.

basically, the feature to use the same UUIDs was already present (but with an option) and that option is now removed. While I think we should just extend this feature (or fix it) to include zaaktype UUIDs which were not working in the first place. right?

joeribekker:
So, we agree 💯

stevenbal:
> I'd recomend to leave the catalogus generate_new_uuids option as-is, and to add generate_new_uuids to the ZaakTypeImportForm

will do :+1:

src/openzaak/components/catalogi/tests/admin/test_import_export_zaaktype.py
stevenbal: re-added them, because it's now optional to generate new UUIDs

src/openzaak/components/catalogi/admin/utils.py
```diff
@@ -204,8 +213,9 @@ def import_zaaktype_for_catalogus(
...
 
-                for old, new in uuid_mapping.items():
-                    data = data.replace(old, new)
+                if generate_new_uuids:
```
annashamray, line 216: I think this condition should be moved a little above and also includes `iotypen_uuid_mapping` and `besluittypen_uuid_mapping` ?

annashamray (approved):
Looks good

src/openzaak/components/catalogi/admin/utils.py
stevenbal: I thought so too, but this is not the case, because even when keeping the same UUIDs for created instances, you can still link with existing instances that have different UUIDs and in that case you do want to use those mappings (same comment as https://github.com/open-zaak/open-zaak/pull/1681#discussion_r1647360243)

PR #1685, opened 2024-06-18
:zap:[#1256] remove potential race condition

### LINKED ISSUE #1256: API create/update: Show conflicting identification number when returning a "identificatie-niet-uniek" error

Coperh:
Fixes #1256

Depends on maykinmedia/commonground-api-common#31

### DIFF
--- a/src/openzaak/utils/validators.py
+++ b/src/openzaak/utils/validators.py
@@ -15,11 +15,7 @@ from rest_framework.validators import (
 )
 from vng_api_common.oas import obj_has_shape
 from vng_api_common.utils import get_resource_for_path, get_uuid_from_path
-from vng_api_common.validators import (
-    IsImmutableValidator,
-    UniekeIdentificatieValidator as _UniekeIdentificatieValidator,
-    URLValidator,
-)
+from vng_api_common.validators import IsImmutableValidator, URLValidator
 
 from openzaak.api_standards import APIStandard
 from openzaak.components.documenten.models import EnkelvoudigInformatieObject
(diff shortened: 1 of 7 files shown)


stevenbal (approved):
Solution seems good

@annashamray this should probably be moved to `commonground-api-common` though?

src/openzaak/utils/validators.py
```diff
@@ -276,14 +276,48 @@ def __call__(self, url: str):
...
+            )
+            .exists()
+        )
+
+        # only part changed
+        if combination_exists:
+            raise serializers.ValidationError(
+                {
+                    self.identificatie_field: self.message.format(
+                        identificatie=identificatie
```
annashamray, line 319: Am I correct that the only change of the validator is this line of code? 
In this case please move this change to the `commonground-api-common` like Steven suggested

annashamray (changes requested):

PR #1689, opened 2024-06-19
:sparkles: [#1637] added environmental variable to disable 2fa

### LINKED ISSUE #1637: Add maykin_2fa

bart-maykin:
Fixes #1637

Moved environmental variable to disable 2FA to the base.py

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,11 @@
 Changelog
 =========
 
+1.13.0 (2024-06-14)
+
+**New features**
+* [#1637] Added 2FA which can be disabled by the environment variable``DISABLE_2FA``.
+
 1.12.1 (2024-04-05)
 -------------------
 
--- a/src/openzaak/conf/includes/base.py
+++ b/src/openzaak/conf/includes/base.py
@@ -693,6 +693,11 @@ MAYKIN_2FA_ALLOW_MFA_BYPASS_BACKENDS = [
     "mozilla_django_oidc_db.backends.OIDCAuthenticationBackend",
 ]
 
+# if DISABLE_2FA is true, fill the MAYKIN_2FA_ALLOW_MFA_BYPASS_BACKENDS with all
+# configured AUTHENTICATION_BACKENDS and thus disabeling the entire 2FA chain.
+if config("DISABLE_2FA", default=False):  # pragma: no cover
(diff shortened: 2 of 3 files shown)


src/openzaak/conf/includes/base.py
```diff
@@ -693,6 +693,13 @@
     "mozilla_django_oidc_db.backends.OIDCAuthenticationBackend",
 ]
 
+ENABLE_2FA = config("ENABLE_2FA", default=False)
+
+# if ENABLE_2FA is true, fill the MAYKIN_2FA_ALLOW_MFA_BYPASS_BACKENDS with all
+# configured AUTHENTICATION_BACKENDS and thus disabeling the entire 2FA chain
+if ENABLE_2FA:  # pragma: no cover
```
annashamray, line 700: I don't get it, why do you disable 2FA with `ENABLE_2FA` env var? 
Could you just move this piece of code from `conf/dev.py` to `conf/includes/base.py`?

annashamray (changes requested):

src/openzaak/conf/includes/base.py
```diff
@@ -693,6 +693,11 @@
     "mozilla_django_oidc_db.backends.OIDCAuthenticationBackend",
 ]
 
+# if DISABLE_2FA is true, fill the MAYKIN_2FA_ALLOW_MFA_BYPASS_BACKENDS with all
+# configured AUTHENTICATION_BACKENDS and thus disabeling the entire 2FA chain
+if config("DISABLE_2FA", default=True):  # pragma: no cover
```
annashamray, line 698:
```suggestion
if config("DISABLE_2FA", default=False):  # pragma: no cover
```

Let's make it more secure by default

CHANGELOG.rst
```diff
@@ -1,6 +1,11 @@
 Changelog
 =========
 
+1.13.0 (2024-06-14)
+
+**New features**
+* [#1637] Added 2FA which is disabled by the environment variable``DISABLE_2FA`` by default.
```
annashamray, line 7:
```suggestion
* [#1637] Added 2FA which is enabled by default. Use the environment variable``DISABLE_2FA`` to turn it off if needed
```

annashamray (changes requested):

annashamray (approved):

PR #1693, opened 2024-06-19
[#1692] add document bulk import delete endpoint

### LINKED ISSUE #1692: Toevoegen van DELETE endpoint voor imports

1.13.0

Na het mergen van https://github.com/open-zaak/open-zaak/commit/2c503c38f2b6c9a01bbc93185c95224cdd57dd6e ontbreekt er nog een DELETE endpoint voor imports (zie https://github.com/open-zaak/open-zaak/issues/1596#issuecomment-2058603085). Daarnaast ontbreekt er ook een manier om imports ouder dan 7 dagen te verwijderen.

_No response_

_No response_

SonnyBA:
Fixes #1692 

**Changes**

Adds a delete endpoint for `Import`'s and adds a celery task to periodically delete `Import`'s older than 7 days.


docs/installation/reference/import.rst
```diff
@@ -126,6 +127,15 @@ a row was imported successfully and if there any comments about the row.
 
 See the `API documentation`_ for more details.
 
+**Deleting an Import**
+
+When an ``Import`` instance has the status ``finished``, ``error`` or ``pending``
+it can be deleted. Deletion of ``Import`` instances that are older than 7 days
+and have one of the above mentioned statusses is done in the background through
+a daily ocurring task.
```
stevenbal, line 135:
```suggestion
and have one of the above mentioned statuses is done in the background through
a daily occurring task.
```

src/openzaak/conf/includes/base.py
```diff
@@ -221,6 +222,14 @@
     "openzaak.config.bootstrap.demo.DemoUserStep",
 ]
 
+# Note that by default UTC times are used here (see `nowfun` kwarg)
+CELERY_BEAT_SCHEDULE = {
+    "daily-remove-imports": {
+        "task": "openzaak.import_data.tasks.remove_imports",
+        "schedule": crontab(hour="9"),
+    }
+}
```
stevenbal, line 231: Open Zaak doesn't have celery beat yet, so the required configuration to set that up will have to be added as well

src/openzaak/import_data/tests/test_tasks.py
```diff
@@ -0,0 +1,69 @@
...
+            ImportFactory(
+                status=ImportStatusChoices.error,
+                finished_on=datetime(2023, 12, 25, 5, tzinfo=pytz.utc),
+                total=0,
+            ),
+        )
+
+        active_import = ImportFactory(
+            status=ImportStatusChoices.active,
+            finished_on=datetime(2023, 12, 25, 6, tzinfo=pytz.utc),
```
stevenbal, line 38: Is is possible in practice for an Import to have status `active` and also have `finished_on` set?

src/openzaak/import_data/views.py
```diff
@@ -217,3 +217,25 @@ def retrieve(self, request, *args, **kwargs):
...
+    def get_queryset(self):
+        return Import.objects.filter(import_type=self.import_type)
+
+    def destroy(self, request, *args, **kwargs):
+        instance = self.get_object()
+
+        if instance.status not in ImportStatusChoices.deletion_choices:
+            message = (
+                _("Import cannot be deleted due to current status: %s")
+                % instance.status
```
stevenbal, line 236: It's not really an issue in this case, but in general I think it is better to use `.format` with keyword arguments for translatable strings. That way, if there are multiple variables used is it possible to know which is which and you can move them around in the translated version if needed

```suggestion
                _("Import cannot be deleted due to current status: {status}").format(status=instance.status)
```

SonnyBA: I usually use positional string interpolation for string formatting with translations for only one variable. [This should always be compatible with translations ](interpolation)(be it lazy or not).

src/openzaak/import_data/tests/test_tasks.py
SonnyBA: This should not be possible but I included it here in the test to check no active imports could be removed (even when they incorrectly have `finished_on` set)

src/openzaak/conf/includes/base.py
SonnyBA: This is added through https://github.com/open-zaak/open-zaak/pull/1693/commits/387901e7728bda8403af42f0d5cf975d95632bb0. Although `django-celery-beat` [is not mandatory for using periodic tasks](https://docs.celeryq.dev/en/stable/userguide/periodic-tasks.html#using-custom-scheduler-classes) (like the one which is being added through this PR) I think it is the most practical way to add it. The alternative is to use the default scheduler (which removes the need for adding `django-celery-beat`) which saves the schedule in a file on the filesystem (which could be removed when a docker setup is used and not configured with any persistent storage solution for it).

src/openzaak/import_data/tasks.py
```diff
@@ -0,0 +1,25 @@
...
+from django.utils import timezone
+
+from openzaak import celery_app
+from openzaak.import_data.models import Import, ImportStatusChoices
+
+logger = logging.getLogger(__name__)
+
+
+@celery_app.task()
+def remove_imports(days_back=7):
```
annashamray, line 15: let's make `days_back` a django setting which can be configured with an env var. Since it's a celery beat task I don't think we will call it with parameters

annashamray (changes requested):
Looks good, just some minor remarks

annashamray (approved):
Looks good

In the future could you please answer my comments or mark them as resolved, so I would know what to check in re-reviewing

stevenbal (approved):

Issue #17, opened 2019-07-29
As developer, I want to have an indication of performance for scenarios


joeribekker:
... so I can see where optimizations need to be made.

**Description**
Run performance test scenario's and report the results.

joeribekker:
Note that as per #78 there is one additional test scenario: list of zaken filtered by zaaktype.

joeribekker:
@annashamray  I think this issue is done seeing: https://bitbucket.org/maykinmedia/open-zaak-performance-test/src/master/results/ ?

annashamray:
> @annashamray I think this issue is done seeing: https://bitbucket.org/maykinmedia/open-zaak-performance-test/src/master/results/ ?

@joeribekker I totally agree

PR #1700, opened 2024-06-27
[#17] add jobs for checking for changes for OAS files

SonnyBA:
Partially fixes https://github.com/maykinmedia/open-api-framework/issues/17

**Changes**

Adds CI steps to check if no missing OAS spec changes are encountered.


stevenbal (approved):
@SonnyBA could you squash this into a single commit?

annashamray (approved):
I think it belongs more to code-quality than to ci check (because it looks like the existing migration check), but it's up to you

PR #1709, opened 2024-07-05
:construction_worker: [#1708] add checking OAS in the GH actions

### LINKED ISSUE #1699: The ZGW OpenAPI specifications contain 'min_length' and 'max_length' properties which are prohibited according to the schema

1.13.0

The ZGW OpenAPI specifications contain `min_length` and `max_length` properties which are prohibited according to the schema. E.g. https://github.com/open-zaak/open-zaak/blob/main/src/openzaak/components/zaken/openapi.yaml but this also goes for the other `openapi.yaml` specs.

When we try to generate Java stub client code from such a spec using https://github.com/OpenAPITools/openapi-generator we get schema validation errors such as:

`Schema validation: Property 'min_length' is not allowed`

I think these properties should be renamed to `minLength` and `maxLength` instead.

1. Download an openapi.yaml file. E.g. https://github.com/open-zaak/open-zaak/blob/main/src/openzaak/components/zaken/openapi.yaml
2. Use https://github.com/OpenAPITools/openapi-generator to generate Java client code from this spec.
3. Schema validation errors occur such as: `Schema validation: Property 'min_length' is not allowed`

No schema validation errors. Older versions of OpenZaak did not have this issue. We are upgrading from 1.10.3 to 1.13.0. Not sure in which version this issue was introduced.

### LINKED ISSUE #1702: DRC OpenAPI spec schema validation error

1.13.0

The new DRC API OpenAPI spec contains a schema validation error for the `/import/{uuid}/upload`: endpoint. The schema response type is set to null, which is not allowed:

``` 
content:
        application/json:
         schema: null         
```

For now we implemented a workaround by disabling the schema validation:

```
   // this OpenAPI spec contains a schema validation error: `schema: null`
   // so we disable the schema validation for this spec until this is fixed in a future version of this spec
   validateSpec.set(false)
```

1. Generate Java stub code from the DRC OpenAPI spec using https://github.com/OpenAPITools/openapi-generator
2. A schema validation error is shown. The Java stub code is not generated

No schema validation error. Java code is generated correctly.

### LINKED ISSUE #1708: As developer I want to check OAS in GH actions

Should be added to all APIs: 

* [x] Open Zaak https://github.com/open-zaak/open-zaak/pull/1709
* [x] Open Notificaties https://github.com/open-zaak/open-notificaties/pull/172
* [x] Open Klant https://github.com/maykinmedia/open-klant/pull/202

Objects API and Objecttypes API are fine

annashamray:
Fixes #1708 (Open Zaak part)
Fixes #1699 
Fixes #1702 

**NOTE**
I've fixed all linter warnings except pagination ones, which appear because of custom responses for expand endpoints. I've tried to fix these warnings, which resulted in CI failing (commit https://github.com/open-zaak/open-zaak/pull/1709/commits/70fc9ae6a756162a227467f819fe71b6c44c5993) I couldn't fix this error, because I couldn't reproduce it locally. So after spending a lot of time on it I reverted the change. 
If you can fix it please do :)


src/openzaak/components/catalogi/openapi.yaml
```diff
@@ -11361,105 +11214,6 @@ components:
       - informatieobjectcategorie
       - omschrijving
       - vertrouwelijkheidaanduiding
-    Notificatie:
```
stevenbal, line 11364: Makes me wonder why this and the callbacks above were here in the first place :sweat_smile:

annashamray: They appeared in the OAS after moving to drf_spectacular. We hid them in `drf_yasg`, but I forgot to hide them also in the `drf-spectacular`

PR #1715, opened 2024-07-16
:bug:[#1707] fix error when no latest version object

### LINKED ISSUE #1707: AttributeError /zaken/api/v{version}/zaken/{uuid} 'NoneType' object has no attribute 'uuid'

1.11.0

Sentry maykin-media/issues/363194

```
AttributeError: 'NoneType' object has no attribute 'uuid'
  File "rest_framework/views.py", line 506, in dispatch
    response = handler(request, *args, **kwargs)
  File "rest_framework/mixins.py", line 82, in partial_update
    return self.update(request, *args, **kwargs)
  File "notifications_api_common/viewsets.py", line 195, in update
    with conditional_atomic(self.notifications_wrap_in_atomic_block)():
  File "django/db/transaction.py", line 290, in __exit__
    connection.set_autocommit(True)
  File "django/db/backends/base/base.py", line 420, in set_autocommit
    self.run_and_clear_commit_hooks()
  File "django/db/backends/base/base.py", line 653, in run_and_clear_commit_hooks
    func()
  File "vng_api_common/caching/etags.py", line 135, in __call__
    return self.callback()
  File "vng_api_common/caching/etags.py", line 180, in calculate_new_value
    self.instance.calculate_etag_value()
  File "vng_api_common/caching/models.py", line 35, in calculate_etag_value
    self._etag = calculate_etag(self)
  File "vng_api_common/caching/etags.py", line 92, in calculate_etag
    rendered = renderer.render(serializer.data, "application/json")
  File "rest_framework/serializers.py", line 548, in data
    ret = super().data
  File "rest_framework/serializers.py", line 246, in data
    self._data = self.to_representation(self.instance)
  File "rest_framework/serializers.py", line 515, in to_representation
    ret[field.field_name] = field.to_representation(attribute)
  File "openzaak/components/documenten/api/fields.py", line 32, in to_representation
    kwargs={"uuid": value.uuid},
```

from the shell of a project talking to our test env open-zaak:

```python
from zgw_consumers.models import Service
from zgw_consumers.constants import APITypes
from zgw_consumers.client import build_client

service = Service.objects.get(api_type=APITypes.zrc)
zrc_client = build_client(service)
with zrc_client:
     response = zrc_client.patch("zaken/d5aca3e4-14f9-4345-8c17-f03668b32a37", headers={"Accept-Crs": "EPSG:4326","Content-Crs": "EPSG:4326",}, json={"archiefactiedatum": "2026-01-01"})
```

Either a 400 or success

Coperh:
Fixes #1707 

- Prevents saving `EnkelvoudigInformatieObjectCanonical` without a version. 
- Requires (Besluit/Zaak)InformatieObjectect's related canonical to have a Version
- Adds script to delete Canonical without related objects,
- Runs said script on version in migration


src/openzaak/components/documenten/api/fields.py
```diff
@@ -27,6 +27,10 @@ def to_representation(self, value):
             return super().to_representation(value)
 
         value = value.latest_version
+
+        if value is None:
```
bart-maykin, line 31: Why are we returning the super().to_representation() without the value?

If there is no good reason then I would suggest to return the super at the end of the method, and have a check to see if we want to return the reverse of "enkelvoudiginformatieobject-detail".

src/openzaak/components/documenten/tests/admin/test_eiocanonical_inline_audittrail.py
```diff
@@ -105,7 +105,10 @@ def test_eio_add(self):
 
         get_response = self.app.get(self.change_url)
 
-        form = get_response.form
+        response = get_response.form.submit()
+        self.assertEqual(response.status_code, 200)
+
+        form = response.form
```
Coperh, line 111: I have no idea how to get the form error list for the inline, it does does display here

src/openzaak/components/documenten/api/fields.py
Coperh: Idk, it probably makes more sense to just return an empty string or None, but this situation is not supposed to happen in the first place.

bart-maykin: If it isn't supposed to happen shouldn't we raise an exception instead or atleast log it? I might be thinking too much into it though.

src/openzaak/components/documenten/tests/admin/test_eiocanonical_inline_audittrail.py
annashamray: You can try `response.context["errors"]` to see all errors in the page. If you want to access inlines you can try `response.context["inline_admin_formsets"]`

annashamray (changes requested):
Could you please add script to remove canonical objects without latest_version from the DB. It can be used both in data migration and in the management command

src/openzaak/components/documenten/management/commands/delete_empty_eioc.py
```diff
@@ -0,0 +1,103 @@
+# SPDX-License-Identifier: EUPL-1.2
+# Copyright (C) 2024 Dimpact
```
Coperh, line 2: Mostly copied from detect_duplicate_documents

src/openzaak/components/documenten/management/commands/delete_empty_eioc.py
```diff
@@ -0,0 +1,103 @@
...
+        msg = _("Found {count} empty records.").format(count=len(duplicates))
+        self.stdout.write(msg)
+
+        if not options["interactive"]:
+            delete_empties(duplicates)
+            self.stdout.write(self.style.SUCCESS(_("Deleted all the empties.")))
+            return
+
+        while True:
+            option = self.get_option()
```
annashamray, line 52: Looks good, but let's make it less interactive :) 
The idea is that this command can be used in some deployment script without user input. 

Options look nice, so maybe add `interactive` flag to the command if you want to keep the them? 

And could you please create a regression test with this command, which is based on this issue with following steps:
* create canonical object without latest_version
* relate zaak to it
* try to patch the zaak (it should fail)
* call the command in the non-interactive mode
* try to patch the zaak again (success)

annashamray (changes requested):

src/openzaak/components/documenten/tests/management/test_delete_empty_eioc.py
```diff
@@ -0,0 +1,64 @@
...
+
+        latest.delete()
+
+        # zio_url = reverse(zio)
+        # with self.assertRaises(AttributeError):
+        #     self.client.get(zio_url, **ZAAK_WRITE_KWARGS)
+        #
+        # bio_url = reverse(bio)
+        # with self.assertRaises(AttributeError):
+        #     self.client.get(bio_url)
```
Coperh, line 48: These fail in the CI but work locally and I don't know why.

A lot of the initial issues I could not really recreate either.

src/openzaak/components/documenten/management/commands/delete_empty_eioc.py
Coperh: I've just removed the interactivity

src/openzaak/components/documenten/tests/management/test_delete_empty_eioc.py
```diff
@@ -0,0 +1,64 @@
...
+
+        eioc = EnkelvoudigInformatieObjectCanonicalFactory()
+        latest = eioc.latest_version
+        zaak = ZaakFactory()
+        ZaakInformatieObjectFactory(informatieobject=eioc, zaak=zaak)
+
+        besluit = BesluitFactory()
+        BesluitInformatieObjectFactory(informatieobject=eioc, besluit=besluit)
+
+        latest.delete()
```
Coperh, line 40: How about handling deleting? Should I prevent it if there is only one left? What if you want to delete a version? Should it only delete if the canonical is deleted?

src/openzaak/components/documenten/management/commands/delete_empty_eioc.py
```diff
@@ -0,0 +1,60 @@
...
+from openzaak.components.documenten.models import EnkelvoudigInformatieObjectCanonical
+from openzaak.components.zaken.models import ZaakInformatieObject
+
+
+class Command(BaseCommand):
+    help = "Check for any empty document identities (no related document)."
+
+    def handle(self, *args, **options):
+        if settings.CMIS_ENABLED:
+            self.stdout.write(_("This command does not run with CMIS enabled."))
```
stevenbal, line 20: I don't think we have to mark the strings in this file as translatable because this is just for admins, let's keep them in English

src/openzaak/components/documenten/management/commands/delete_empty_eioc.py
```diff
@@ -0,0 +1,60 @@
...
+        delete_empties(related_ziots)
+        delete_empties(related_biots)
+        delete_empties(duplicates)
+
+
+@transaction.atomic
+def delete_empties(empties: List) -> None:
+    for document_id in empties:
+        document_id.delete()
+    return
```
stevenbal, line 60: I think this can be simplified:
```suggestion
with transaction.atomic():
    related_ziots.delete()
    related_biots.delete()
    duplicates.delete()
```

src/openzaak/components/documenten/management/commands/delete_empty_eioc.py
```diff
@@ -0,0 +1,60 @@
...
+from django.db import transaction
+from django.utils.translation import gettext_lazy as _
+
+from openzaak.components.besluiten.models import BesluitInformatieObject
+from openzaak.components.documenten.models import EnkelvoudigInformatieObjectCanonical
+from openzaak.components.zaken.models import ZaakInformatieObject
+
+
+class Command(BaseCommand):
+    help = "Check for any empty document identities (no related document)."
```
stevenbal, line 16:
```suggestion
    help = "Delete any empty Document identities and delete related Besluit/ZaakInformatieObjecten"
```

src/openzaak/components/documenten/management/commands/delete_empty_eioc.py
```diff
@@ -0,0 +1,60 @@
...
+            self.stdout.write(_("This command does not run with CMIS enabled."))
+            return
+
+        msg = _("Checking {count} records ...").format(
+            count=EnkelvoudigInformatieObjectCanonical.objects.count()
+        )
+        self.stdout.write(msg)
+
+        duplicates = EnkelvoudigInformatieObjectCanonical.objects.filter(
+            enkelvoudiginformatieobject__isnull=True
```
stevenbal, line 29: Do we know where these canonicals without version come from? Are they created by admins?

src/openzaak/components/besluiten/models.py
```diff
@@ -324,3 +325,15 @@ def unique_representation(self):
...
+
+    def save(self, *args, **kwargs):
+
+        if (
+            self._informatieobject is not None
+            and self._informatieobject.latest_version is None
+        ):
+            raise ValidationError(
+                "Related InformatieObject must have at least one version"
+            )
```
stevenbal, line 337: This does raise a 500 in case you save it via the admin, but tbh people shouldn't be doing this via the admin anyway, so I'll allow it :p

src/openzaak/components/documenten/tests/admin/test_eiocanonical_inline_audittrail.py
```diff
@@ -136,3 +137,36 @@ def test_eio_add(self):
 
         new_data = audittrail.nieuw
         self.assertEqual(new_data["identificatie"], "12345")
+
+    def test_eio_add_no_version(self):
```
stevenbal, line 141: This test file is specifically for audittrail stuff, so I don't think this is the right place for this test.

src/openzaak/components/documenten/tests/admin/test_eiocanonical_inline_audittrail.py
```diff
@@ -136,3 +137,36 @@ def test_eio_add(self):
...
+            version_form.errors["creatiedatum"], [_("This field is required.")]
+        )
+        self.assertEqual(version_form.errors["titel"], [_("This field is required.")])
+        self.assertEqual(version_form.errors["auteur"], [_("This field is required.")])
+        self.assertEqual(version_form.errors["taal"], [_("This field is required.")])
+        self.assertEqual(
+            version_form.errors["bestandsomvang"], [_("This field is required.")]
+        )
+        self.assertEqual(version_form.errors["inhoud"], [_("This field is required.")])
+        self.assertEqual(
```
stevenbal, line 163: It might also be good to add an assertion to verify that no object was created as a sanity check

src/openzaak/components/documenten/tests/management/test_delete_empty_eioc.py
stevenbal: I'm not sure if it's possible via the API to delete a specific version, I would say that we should reflect the API behavior as much as possible, so you could check what happens now in these cases if you try to delete the EIO via the API

src/openzaak/components/documenten/tests/management/test_delete_empty_eioc.py
stevenbal: What error do you get in CI?

stevenbal (changes requested):

src/openzaak/components/documenten/tests/management/test_delete_empty_eioc.py
Coperh: The attribute error is not raise when it should be raised

src/openzaak/components/documenten/management/commands/delete_empty_eioc.py
Coperh: They must deleted individually and not with a manager. But i put them all in the same transaction

src/openzaak/components/documenten/tests/management/test_delete_empty_eioc.py
stevenbal: @Coperh did you check the API behavior here?

annashamray: I agree with Steven, I think we can't delete the only document version without canonical object. 
You can modify this test by removing creating a document (and have something like `EnkelvoudigInformatieObjectCanonicalFactory(latest_version=None)`
I think this is what happened in the original issue. 

But it's optional, since the end result is the same

annashamray: Please ignore my previous comment, I forget that the validation was introduced to prevent it

src/openzaak/components/documenten/tests/management/test_delete_empty_eioc.py
stevenbal: might be related to the value of `CMIS_ENABLED`? In that case we should add `skipIf` or something like that I guess

src/openzaak/components/documenten/tests/management/test_delete_empty_eioc.py
```diff
@@ -0,0 +1,68 @@
...
+from ...models import EnkelvoudigInformatieObjectCanonical
+from ..factories import EnkelvoudigInformatieObjectCanonicalFactory
+
+
+class DeleteEmptyEIOCTests(JWTAuthMixin, APITestCase):
+    heeft_alle_autorisaties = True
+
+    def setUp(self):
+        super().setUp()
+        self.client.raise_request_exception = False
```
Coperh, line 32: If the tests pass, the issue was that with DEBUG set to true, it will raise an exception instead of returning a 500 error

annashamray (approved):
Good job!

stevenbal (approved):

PR #1734, opened 2024-08-13
Feature/1732 generate data extra

### LINKED ISSUE #1732: Add selectielijstklasse and archief datum to test data generation.
### Thema / Theme

Zaken API

The (generate_data?) management command creates test data but seems to result in a set of zaken that lacks a (valid) selectielijstklasse and an archiefdatum. 

This allows more realistic test data to be generated benefitting the development of OAB.

_No response_

annashamray:
Fixes #1732

### DIFF
--- a/src/openzaak/management/commands/generate_data.py
+++ b/src/openzaak/management/commands/generate_data.py
@@ -9,6 +9,7 @@ from django.db import models, transaction
 from django.utils import timezone
 
 import factory.fuzzy
+import vcr
 from vng_api_common.constants import VertrouwelijkheidsAanduiding
 
 from openzaak.components.besluiten.models import Besluit, BesluitInformatieObject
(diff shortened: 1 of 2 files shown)


src/openzaak/management/commands/generate_data.py
```diff
@@ -64,6 +65,19 @@
     ZaakInformatieObjectFactory,
     ZaakObjectFactory,
 )
+from openzaak.selectielijst.models import ReferentieLijstConfig
+
+
+@vcr.use_cassette("src/openzaak/tests/cassettes/selectielijst_resultaten.yaml")
```
bart-maykin, line 71: Any reason why we are using vcr to create a cassette? I don't see us using this for anything atm.

bart-maykin (commented):
Seems good although I don't see any tests.
Did you forget to push those by any chance?

src/openzaak/management/commands/generate_data.py
annashamray: I've read vcr documentation one more time and noticed that they recommend to use it with tests. There is no example o use vcr outside of tests. So after some consideration I decided not to use it. So let's just call Selectielijst API resource for real in this command

annashamray:
@bart-maykin I've updated a test and added assertions for Zaak attributes

bart-maykin (approved):

PR #1740, opened 2024-08-16
:bug: [#1737] Fix testsuite tmp folders not being deleted

### LINKED ISSUE #1737: Temp folders are created when running tests with -k options

1.13.0

When you run a command `src/manage.py test src -k test_some_test_name` several temp folders are created
It doesn't happen when you run just ``src/manage.py test src` 

I think the reason is this line, which decorates document import tests:
```
@override_settings(IMPORT_DOCUMENTEN_BASE_DIR=get_temporary_dir())
```

So temp folders are created during class initialization, but removed during tearDownClass which is never called when you use -k option
I think it would be best solution to move creation of temp folders to `setUpClass` method

stevenbal:
Fixes #1737

**Changes**

* Fix testsuite tmp folders not being deleted


annashamray (approved):
Works for me

annashamray:
Please merge it after fixing failing tests in CI

PR #1747, opened 2024-08-23
:arrow_up: bump open-api-framework to 0.8.0 and its related deps

annashamray:


Coperh (approved):
Looks good

Coperh:
Something I remembered actually, OZ does not use the OAF middleware, so the content security policy will not actually be enabled since it needs the CSP middleware.

annashamray:
@Coperh Good catch! I'll update the middleware

PR #1752, opened 2024-08-28
:arrow_up: [#50] updated deployment packages

bart-maykin:
Fixes [#50](https://github.com/maykinmedia/open-api-framework/issues/50)

**Changes**

- certifi `2024.7.22` to `2024.7.4`
- idna `3.4` to `3.7`
- jinja2 `3.1.2` to `3.1.4`
- urllib3 `1.26.18` to `1.26.19`

### DIFF
--- a/deployment/requirements.txt
+++ b/deployment/requirements.txt
@@ -14,7 +14,7 @@ attrs==23.1.0
     #   referencing
 cachetools==5.3.2
     # via google-auth
-certifi==2023.7.22
+certifi==2024.7.4
     # via
     #   kubernetes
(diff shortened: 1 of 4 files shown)


annashamray (commented):
I think `cryptography` package also may be updated here

bart-maykin:
@annashamray 
> I think `cryptography` package also may be updated here

Yes cryptography will also be upgraded. But I will do that in a separate pr. Because we need to update zgw-consumers for this as well.

annashamray (approved):

PR #1768, opened 2024-09-12
Feature/config steps disabled

annashamray:
Fixes https://github.com/maykinmedia/django-setup-configuration/issues/10 (OZ part)


Coperh (approved):
Yep looks fine

PR #1769, opened 2024-09-12
autorisatie with manual iotypen

### LINKED ISSUE #1718: AutorisatieForm in admin does not show selected informatieobjecttypen

latest / 6bc94b31168af217c3e9b4024c9ac3790e04738d

In the `AutorisatieForm`, selected informatieobjecttypes do not show after saving them

If I save an Autorisatie via the admin with the following settings (I'm not sure if the actual component/scopes matter, but this is how I ran into it locally):

component: Documenten API
scopes: notificaties.consumeren
extra params, select an Informatieobjecttype manually
choose any confidentiality

And then navigate back to the AutorisatiesForm, I see that the option to manually select a type is select, but no type is checked in the selectboxes below it, nor is it part of external types.

The selected type should show in the form

I tested this locally and on the testenv with version 1.11.0 and I could not reproduce it there, so it has to be introduced somewhere between 1.11.0 and `6bc94b31168af217c3e9b4024c9ac3790e04738d`

annashamray:
Fixes #1718

### DIFF
--- a/src/openzaak/components/autorisaties/forms.py
+++ b/src/openzaak/components/autorisaties/forms.py
@@ -587,13 +587,10 @@ class AutorisatieBaseFormSet(forms.BaseFormSet):
                         ):
                             raise ValidationError(
                                 error_msg.format(
-                                    component=data["component"], catalogus=catalogus
+                                    component=data["component"],
+                                    catalogus=_type.catalogus,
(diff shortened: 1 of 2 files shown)


Coperh (approved):
Everything works correctly

PR #1772, opened 2024-09-13
:wrench:[maykinmedia/open-api-framework#68] fix CSP errors

Coperh:
Fixes maykinmedia/open-api-framework#68

Also fixes:  maykinmedia/mozilla-django-oidc-db#116

### DIFF
--- a/src/openzaak/conf/includes/api.py
+++ b/src/openzaak/conf/includes/api.py
@@ -25,6 +25,9 @@ OPENZAAK_API_CONTACT_EMAIL = "support@maykinmedia.nl"
 OPENZAAK_API_CONTACT_URL = "https://www.maykinmedia.nl"
 
 SPECTACULAR_SETTINGS = {
+    "SWAGGER_UI_DIST": "SIDECAR",
+    "SWAGGER_UI_FAVICON_HREF": "SIDECAR",
+    "REDOC_DIST": "SIDECAR",
     # info object
     "TITLE": "Open Zaak API",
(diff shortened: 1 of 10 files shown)


docs/conf.py
```diff
@@ -84,6 +84,7 @@
     r"http://localhost:\d+/",
     r"http://127\.0\.0\.1:\d+/",
     r"https://.*sentry\.openzaak\.nl.*",
+    r"https://miniwebtool.com",  # cannot be fetched with cURL or non-browser
```
Coperh, line 87: Some anything spam thing probably prevents cURL from requesting this site, gets a 403 response

annashamray: The link to miniwebtool should already be removed from the docs, let's not include it in the list of trusted urls

src/openzaak/conf/includes/api.py
```diff
@@ -25,6 +25,9 @@
 OPENZAAK_API_CONTACT_URL = "https://www.maykinmedia.nl"
 
 SPECTACULAR_SETTINGS = {
+    "SWAGGER_UI_DIST": "SIDECAR",
+    "SWAGGER_UI_FAVICON_HREF": "SIDECAR",
```
annashamray, line 29: I think these two settings are excessive since we don't provide swagger ui views, but I'll leave it up to you

annashamray (changes requested):

docs/conf.py
Coperh: Removed it from the rest of the documentation (and some comments too)

annashamray (approved):

PR #1781, opened 2024-09-19
:bug: [#1780] Fix setting NOTIFICATIONS_DISABLED via envvar

### LINKED ISSUE #1780: wrong order of settings makes NOTIFICATIONS_DISABLED always False

1.14.0

`NOTIFICATION_DISABLED` setting can't be managed with the env var now.

From OZ slack (discovered by @sergei-maertens ):
```
openzaak@7df63db34406:/app$ printenv | grep NOTI
NOTIFICATIONS_DISABLED=true
openzaak@7df63db34406:/app$ src/manage.py shell
>>> from django.conf import settings
>>> settings.NOTIFICATIONS_DISABLED
False
```

The reason is in the import order of settings:

```
import openzaak.conf.docker
from openzaak.conf.docker.production import *
from openzaak.conf.includes.base import *
from notifications_api_common.settings import *
from open_api_framework.conf.base import *  # defines NOTIFICATIONS_DISABLED
from openzaak.conf.includes.api import *
from vng_api_common.conf.api import *  # fucks up NOTIFICATIONS_DISABLED
```

_No response_

_No response_

stevenbal:
Fixes #1780 

**Changes**

* Fix setting NOTIFICATIONS_DISABLED via envvar

### DIFF
--- a/src/openzaak/conf/includes/api.py
+++ b/src/openzaak/conf/includes/api.py
@@ -1,11 +1,8 @@
 # SPDX-License-Identifier: EUPL-1.2
 # Copyright (C) 2019 - 2022 Dimpact
-from vng_api_common.conf.api import *  # noqa - imports white-listed
+from vng_api_common.conf.api import BASE_REST_FRAMEWORK, ZDS_CLIENT_CLASS  # noqa: F401
 
 from openzaak.api_standards import APIStandard
-
-# Remove the reference - we don't have a single API version.
-del API_VERSION  # noqa
 
 AUTORISATIES_API_VERSION = "1.0.0"
 BESLUITEN_API_VERSION = "1.1.0"
@@ -55,8 +52,6 @@ SPECTACULAR_SETTINGS = {
     "DEFAULT_GENERATOR_CLASS": "openzaak.utils.generators.SchemaGenerator",
 }
 
-
-GEMMA_URL_INFORMATIEMODEL_VERSIE = "1.0"
 
 COMMONGROUND_API_COMMON_GET_DOMAIN = "openzaak.utils.get_openzaak_domain"
 


stevenbal:
@annashamray I tried writing a test for this, but I'm not sure if this is possible, because the settings we use (`ci.py`) also hardcodes `NOTIFICATIONS_DISABLED`

annashamray:
@stevenbal Should we just import particular settings from vng_api_common without the wildcard? 
I think it would be more readable then importing everything and then deleting

stevenbal:
yeah you're right, I'll change it

annashamray:
Or maybe just use OAF settings in OZ, and OAF itself can reuse vng_api_common settings? 

I'm trying to think of reducing the number of setting imports in OZ but still having the benefits of reusable settings in OAF

annashamray:
But I leave it to your judgement

stevenbal:
I think this might be sort of in the grey area between OAF and vng-api-common, settings like `COMMONGROUND_API_COMMON_GET_DOMAIN` seem pretty specific to vng-api-common and I'm not sure if we want to manage them in OAF :thinking:

src/openzaak/conf/includes/api.py
```diff
@@ -1,12 +1,20 @@
...
 # Copyright (C) 2019 - 2022 Dimpact
-from vng_api_common.conf.api import *  # noqa - imports white-listed
+from vng_api_common.conf.api import (  # noqa: F401
+    BASE_REST_FRAMEWORK,
+    BASE_SWAGGER_SETTINGS,
+    COMMON_SPEC,
+    ZDS_CLIENT_CLASS,
+    GEMMA_URL_TEMPLATE,
+    GEMMA_URL_COMPONENTTYPE,
+    GEMMA_URL_INFORMATIEMODEL,
```
annashamray, line 10: Correct me if I'm wrong but these settings are used only in generate_swagger command which is outdated and not used in OZ (and will be removed together with drf-yasg)

src/openzaak/conf/includes/api.py
```diff
@@ -1,12 +1,20 @@
 # SPDX-License-Identifier: EUPL-1.2
 # Copyright (C) 2019 - 2022 Dimpact
-from vng_api_common.conf.api import *  # noqa - imports white-listed
+from vng_api_common.conf.api import (  # noqa: F401
+    BASE_REST_FRAMEWORK,
+    BASE_SWAGGER_SETTINGS,
```
annashamray, line 5: I think we don't use it in OZ anymore, it's replaced with drf_spectacular settings

src/openzaak/conf/includes/api.py
```diff
@@ -1,12 +1,20 @@
...
+    COMMON_SPEC,
+    ZDS_CLIENT_CLASS,
+    GEMMA_URL_TEMPLATE,
+    GEMMA_URL_COMPONENTTYPE,
+    GEMMA_URL_INFORMATIEMODEL,
+    REDOC_SETTINGS,
```
annashamray, line 11: same comment as for BASE_SWAGGER_SETTINGS

src/openzaak/conf/includes/api.py
```diff
@@ -1,12 +1,20 @@
 # SPDX-License-Identifier: EUPL-1.2
 # Copyright (C) 2019 - 2022 Dimpact
-from vng_api_common.conf.api import *  # noqa - imports white-listed
+from vng_api_common.conf.api import (  # noqa: F401
+    BASE_REST_FRAMEWORK,
+    BASE_SWAGGER_SETTINGS,
+    COMMON_SPEC,
```
annashamray, line 6: Do we use this setting in OZ?

annashamray (changes requested):
Let's use this opportunity to clean up settings a little bit

annashamray:
@stevenbal Ok, let's just reorganize settings in OZ for now
And could you please take a look if the same thing with NOTIFICATION_DISABLED happens in Objects API?

src/openzaak/conf/includes/api.py
stevenbal: I checked it, it's used by `use_external_components` command in vng_api_common, but I don't see that being used in OZ, so I'll remove this setting

stevenbal:
@annashamray I checked it, the same issue occurs there as well: https://github.com/maykinmedia/objects-api/pull/452

annashamray (approved):
:+1:

PR #1790, opened 2024-10-01
:memo: Remove broken OIDC link

Coperh:


annashamray (approved):
It looks like NLX documentation is not available anymore, I wonder if it's temporary or not

stevenbal:
@annashamray it was moved, I've found https://docs.fsc.nlx.io/fsc-in-production/setup-oidc, but I'm not 100% sure if this was the same page as before

PR #1795, opened 2024-10-09
✨ [#1783] Add zaaktype__not_in filter to ZaakZoek

### LINKED ISSUE #1783: zaaktype__not_in filter on zaak search
### Thema / Theme

Zaken API

In some cases we want to get all zaken, but not of a few specific types.
DH Usecase:
- A party started registering zaken in OpenZaak, great, but they did not expect them to be visible in our customer portal
- Now we want to exclude zaken from those zaaktypes from our searches so we can still query properly instead of filtering them out client side

Exceptions in queries can be made on prod, without weird client side magic

Maybe zaken__out, or maybe a negative mark on the search like  `zaaktype__in ["!my.zaak.nl/UUID1", "!my.zaak.nl/UUID2"]` though this might be overkill as a (positive) zaaktype__in already excludes the rest.

stevenbal:
Fixes #1783

**Changes**

* Add zaaktype__not_in filter to ZaakZoek

### DIFF
--- a/src/openzaak/components/zaken/api/serializers/zaken.py
+++ b/src/openzaak/components/zaken/api/serializers/zaken.py
@@ -41,6 +41,7 @@ from openzaak.utils.api import (
 )
 from openzaak.utils.auth import get_auth
 from openzaak.utils.exceptions import DetermineProcessEndDateException
+from openzaak.utils.help_text import mark_experimental
 from openzaak.utils.serializer_fields import FKOrServiceUrlField
 from openzaak.utils.validators import (
     LooseFkIsImmutableValidator,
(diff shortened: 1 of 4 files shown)


src/openzaak/components/zaken/api/serializers/zaken.py
```diff
@@ -547,6 +547,11 @@ class ZaakZoekSerializer(serializers.Serializer):
         required=False,
         help_text=_("Array van zaaktypen."),
     )
+    zaaktype__not_in = serializers.ListField(
```
stevenbal, line 550: @joeribekker @annashamray in https://github.com/open-zaak/open-zaak/issues/1783#issuecomment-2370776330 it's mentioned that this filter should be added to both `zaak_zoek` and `zaak_list`, but `zaaktype__in` is only available on `zaak_zoek` (https://github.com/VNG-Realisatie/gemma-zaken/issues/2072#issuecomment-1198244150), so shouldn't we keep `zaaktype__not_in` in line with this?

stevenbal: also, should I mark it as experimental?

joeribekker: Yes and yes

annashamray (approved):
Nice lookup!

PR #1803, opened 2024-10-30
fix resultaattype.resultaattypeomschrijving for generation data

### LINKED ISSUE #1798: Test data Resultaattype gives 500 when accessing from admin

1.13.0-archief.0

When generating test data with the management command, the resultaattypen have no field "omschrijving". This leads to the error:

```
...
File "openzaak/selectielijst/admin_fields.py", line 93, in get_resultaattype_omschrijving_readonly_field
    return omschrijving["omschrijving"]
KeyError: 'omschrijving'
```

When accessing them in the admin.

Sentry: maykin-media/issues/374838

Go to the page: `/admin/catalogi/resultaattype/3/change/` on the test environment (full URL in sentry)

_No response_

annashamray:
Fixes #1798

### DIFF
--- a/src/openzaak/tests/management/test_generate_data.py
+++ b/src/openzaak/tests/management/test_generate_data.py
@@ -3,12 +3,17 @@
 from unittest.mock import patch
 
 from django.apps import apps
+from django.contrib.sites.models import Site
 from django.core.management import CommandError, call_command
+from django.urls import reverse
 
 import requests_mock
+from django_webtest import WebTest
+from maykin_2fa.test import disable_admin_mfa
 from rest_framework.test import APITestCase
 
-from openzaak.components.catalogi.models import ZaakType
+from openzaak.accounts.tests.factories import SuperUserFactory
+from openzaak.components.catalogi.models import ResultaatType, ZaakType
 from openzaak.components.zaken.models import Zaak
 from openzaak.selectielijst.models import ReferentieLijstConfig
 from openzaak.selectielijst.tests import mock_selectielijst_oas_get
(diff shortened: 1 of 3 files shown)


src/openzaak/tests/management/test_generate_data.py
```diff
@@ -38,10 +43,28 @@ def test_generate_data_yes(self, m):
...
+                    "opmerking": "",
+                },
+                {
+                    "url": f"{config.api_root}resultaattypeomschrijvingen/7cb315fb-4f7b-4a43-aca1-e4522e4c73b3",
+                    "omschrijving": "Afgehandeld",
+                    "definitie": "Afgehandeld",
+                    "opmerking": "",
+                },
+            ],
+        )
```
Coperh, line 67: I dont think this gets called. Is it supposed to be?

Coperh (approved):
Looks good I think.

Just one question about a mock not being used

src/openzaak/tests/management/test_generate_data.py
Coperh: I commented it out and the test passes

annashamray: Hmmm I've commented it and this test fails for me in this line:
https://github.com/open-zaak/open-zaak/blob/fe5b36e72843f94f5d5aea23438ef178f2e593cd/src/openzaak/management/commands/generate_data.py#L263
 
To fill in `resultaattype.resultaattypeomschrijving` we add extra request to Selectielijst API in the `generate_data` command, because we need the url from there.

Coperh: It should be fine then. IDK why it works locally

PR #1810, opened 2024-11-04
[#1741] delete jwtsecret when deleting applicatie in the admin

### LINKED ISSUE #1741: Deleting an Application via the admin via a bulk-delete doesn't delete the underlying Autorisatie

1.12.3 and 1.11 (test.openzaak.nl)

Taiga Utrecht 264

Having trailing entities when deleting an Application was an issue earlier in #1515 , this issue is new but related.

When deleting an indivudal Application things go well, however when you do a bulk-delete with the Application the Autorisatie (or JWTSecret?) object remains. Even though the Application credentials can no longer be used, the client-id's of the Application's Autorisatie can't be reused.

This issue can be easily reproduced by doing a bulk-delete of an Application and then attempting to use the deleted client-id again in a new Application:

image

image

_No response_

_No response_

annashamray:
Fixes #1741

### DIFF
--- a/src/openzaak/components/autorisaties/admin.py
+++ b/src/openzaak/components/autorisaties/admin.py
@@ -121,7 +121,7 @@ class ApplicatieAdmin(admin.ModelAdmin):
     def delete_queryset(self, request, queryset):
         """Given a queryset, delete it from the database."""
         client_ids = queryset.values_list("client_ids", flat=True)
-        client_ids = [client_id[0] for client_id in client_ids]
+        client_ids = sum(list(client_ids), [])
 
         secrets = JWTSecret.objects.filter(identifier__in=client_ids)
         secrets.delete()
--- a/src/openzaak/components/autorisaties/admin.py
+++ b/src/openzaak/components/autorisaties/admin.py
@@ -117,6 +117,17 @@ class ApplicatieAdmin(admin.ModelAdmin):
         secrets.delete()
         super().delete_model(request, obj)
 
+    @transaction.atomic
+    def delete_queryset(self, request, queryset):
+        """Given a queryset, delete it from the database."""
+        client_ids = queryset.values_list("client_ids", flat=True)
+        client_ids = [client_id[0] for client_id in client_ids]
+
+        secrets = JWTSecret.objects.filter(identifier__in=client_ids)
+        secrets.delete()
+
+        super().delete_queryset(request, queryset)
+
     @admin.display(
         description=_("Ready?"),
         boolean=True,
--- a/src/openzaak/components/autorisaties/tests/admin/test_applications_admin.py
+++ b/src/openzaak/components/autorisaties/tests/admin/test_applications_admin.py
@@ -146,6 +146,9 @@ class ApplicationsTests(WebTest):
         JWTSecret.objects.create(identifier="foo1", secret="bar1")
         JWTSecret.objects.create(identifier="foo2", secret="bar2")
         JWTSecret.objects.create(identifier="foo3", secret="bar3")
+        jwt_no_delete = JWTSecret.objects.create(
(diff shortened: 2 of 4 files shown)


src/openzaak/components/autorisaties/admin.py
```diff
@@ -117,6 +117,17 @@ def delete_model(self, request, obj):
         secrets.delete()
         super().delete_model(request, obj)
 
+    @transaction.atomic
+    def delete_queryset(self, request, queryset):
+        """Given a queryset, delete it from the database."""
+        client_ids = queryset.values_list("client_ids", flat=True)
+        client_ids = [client_id[0] for client_id in client_ids]
```
Coperh, line 124:
```suggestion
        client_ids = [client_id for client_id  in client_id_field for client_id_field  in client_ids]
```
Change to something like this


This also apparently works. but I've never used it.
```
import itertools
a = [['a','b'], ['c']]
print(list(itertools.chain.from_iterable(a)))
```

annashamray: Thanks for catching it!
I've used `sum`, `itertools.chain` looks also interesting

Coperh (approved):
Yep, looks good

src/openzaak/components/autorisaties/tests/admin/test_applications_admin.py
```diff
@@ -135,3 +135,32 @@ def test_create_applicatie_inline_credentials_without_secret(self):
...
+
+        # confirmation page
+        form = response.form
+        form["action"] = "delete_selected"
+        form["post"] = "yes"
+
+        response = form.submit()
+
+        self.assertEqual(response.status_code, 302)
+        self.assertFalse(JWTSecret.objects.exists())
```
stevenbal, line 166: You could maybe add a JWTSecret that you don't expect to be deleted, to ensure the `delete_queryset` doesn't accidentally delete all JWTSecrets

src/openzaak/components/autorisaties/admin.py
```diff
@@ -117,6 +117,17 @@ def delete_model(self, request, obj):
         secrets.delete()
         super().delete_model(request, obj)
 
+    @transaction.atomic
+    def delete_queryset(self, request, queryset):
+        """Given a queryset, delete it from the database."""
+        client_ids = queryset.values_list("client_ids", flat=True)
+        client_ids = sum(list(client_ids), [])
```
stevenbal, line 124: Is this line necessary?

stevenbal (changes requested):

src/openzaak/components/autorisaties/admin.py
annashamray: Yes, it flats the list
for example, if we delete 2 applicaties, which have client_ids: ['id1', id2'] and ['id3'], then the line123 will give the result:
client_ids = [ ['id1', id2'], ['id3'] ], so we need to flat it further

src/openzaak/components/autorisaties/tests/admin/test_applications_admin.py
annashamray: Sure, I've updated the test

Coperh:
I just remembered, but do we want a management command to delete orphaned JWT secrets?

annashamray:
@Coperh hmmm do we need a management command or a one-time data migration? If deleting jwt secrets work as intended there won't be any orphans anymore, right? 
But there is always a change of some bug of course...

stevenbal:
@annashamray I'd probably add a data migration for it, otherwise people will have to run the management command manually

annashamray:
@stevenbal @Coperh I've added a data migration to remove orphaned jwt secrets

stevenbal:
@annashamray looks good :+1:

PR #1842, opened 2024-12-23
:sparkles: [#1825] add yaml config for authorization configuration step

annashamray:
Fixes #1825

~~Depends on https://github.com/maykinmedia/commonground-api-common/pull/59~~
~~Depends on https://github.com/maykinmedia/commonground-api-common/pull/56~~
~~Depends on https://github.com/open-zaak/open-zaak/pull/1844~~

PR #1844, opened 2024-12-30
Notificaion Setup  Configuraiton

### LINKED ISSUE #1823: Add support for Notification configuration + Channel creation via django-setup-configuration

Coperh:
Fixes #1823

Actually Fixes it


src/openzaak/conf/includes/base.py
```diff
@@ -290,12 +290,11 @@
...
-    "openzaak.config.bootstrap.notifications.AuthNotificationStep",
-    "openzaak.config.bootstrap.notifications.NotificationsAPIConfigurationStep",
-    "openzaak.config.bootstrap.selectielijst.SelectielijstAPIConfigurationStep",
-    "openzaak.config.bootstrap.demo.DemoUserStep",
-    "openzaak.config.bootstrap.authorizations.AuthorizationConfigurationStep",
+    "django_setup_configuration.config.sites.steps.SitesConfigurationStep",
+    "zgw_consumers.contrib.setup_configuration.steps.ServiceConfigurationStep",
+    "notifications_api_common.contrib.setup_configuration.steps.NotificationConfigurationStep",
+    # "openzaak.config.setup_configuration.steps.SelectielijstAPIConfigurationStep",
+    # "openzaak.config.setup_configuration.steps.DemoUserStep",
```
Coperh, line 297: IDK if demo user is actually used and I think selectielijst is covered in the authorization or external PRs

Coperh:
Docker stuff was copied from objects API, should be checked by someone familiar with docker

- Removes old Bootstrap directory
- Converts demo user and Selectielijst steps to new format, but are disabled by default
- Move to common Sites, ZGW and notifcation api common steps
- Delete a load of stuff

src/openzaak/conf/includes/base.py
annashamray: I think DemoConfig can be definetely removed since it's covered partly by `commonground-api-common.JWTSecretsConfigurationStep` and will be wrapped up in #1842

docker/setup_configuration/data.yml
```diff
@@ -0,0 +1,25 @@
...
+  items:
+  - domain: example.com
+    name: Example site
+
+zgw_consumers_config_enable: true
+zgw_consumers:
+  services:
+  - identifier: notifications-api
+    label: Notificaties API
+    api_root: http://notificaties.local/api/v1/
```
annashamray, line 12: I'd add the service with such name to docker-compose so you would have an example of working configuration of ON and OZ

src/openzaak/config/tests/setup_configuraiton/files/auth.yaml
annashamray, line 1: I think this file is used for testing AuthorizationConfigurationStep which was removed, so I don't think if it's still needed here

src/openzaak/config/tests/setup_configuraiton/files/auth_invalid.yaml
annashamray, line 1: Same comment as before: this file was used for testing AuthorizationConfigurationStep

src/openzaak/config/tests/setup_configuraiton/files/domain_mapping.yaml
annashamray, line 1: Same comment: this file was used for testing AuthorizationConfigurationStep

annashamray (commented):
Please take a look at docker/data files (they are also included in the docker volume), I don't think they are needed anymore

The whole docker/ folder needs some love, but this is perhaps an issue for another time

src/openzaak/config/setup_configuration/models.py
```diff
@@ -0,0 +1,33 @@
...
+from vng_api_common.models import JWTSecret
+from zgw_consumers.models import Service
+
+from openzaak.selectielijst.models import ReferentieLijstConfig
+
+
+class DemoConfig(ConfigurationModel):
+
+    demo_client_id = DjangoModelRef(JWTSecret, "identifier")
+    demo_secret = DjangoModelRef(JWTSecret, "secret")
```
stevenbal, line 15: Like Anna remarked, the democonfig stuff can be removed

src/openzaak/config/tests/setup_configuraiton/test_selectielijst_configuration.py
```diff
@@ -0,0 +1,25 @@
...
+    def test_configure(self):
+
+        execute_single_step(
+            SelectielijstAPIConfigurationStep, yaml_source=str(CONFIG_YAML)
+        )
+
+        config = ReferentieLijstConfig.get_solo()
+        self.assertEqual(
+            config.service.api_root, "https://selectielijst.example.com/api/v1/"
+        )
```
stevenbal, line 25: Could you add assertions for all fields on the config here to verify that they're correctly set here? It might also be good to add another test in which you explicitly specify `allowed_years` and `default_year` in the yaml file to verify that that works as well. You can add the Service via a factory in `setUpTestData`, since that should be created by the step from `zgw-consumers`

src/openzaak/config/setup_configuration/steps/selectielijst.py
```diff
@@ -0,0 +1,54 @@
...
+
+class SelectielijstAPIConfigurationStep(BaseConfigurationStep[SelectielijstAPIConfig]):
+    """
+    Configure the Open Zaak client to request Selectielijst API
+
+    1. Create service for Selectielijst API
+    2. Set up configuration for Selectielijst API client
+
+    Normal mode doesn't change the allowed and default years
+    If they are changed, run this command with 'overwrite' flag
```
stevenbal, line 20: This no longer applies in the new version of django-setup-configuration, so I think you can remove these lines

src/openzaak/config/setup_configuration/steps/selectielijst.py
```diff
@@ -0,0 +1,54 @@
...
+        service, created = Service.objects.update_or_create(
+            api_root=model.api_root,
+            defaults={
+                "label": "Selectielijst API",
+                "slug": model.api_root,
+                "api_type": APITypes.orc,
+                "oas": model.oas,
+                "auth_type": AuthTypes.no_auth,
+            },
+        )
```
stevenbal, line 40: zgw-consumers has a step for this (https://zgw-consumers.readthedocs.io/en/latest/setup_config.html), so I think it would be better to reuse that step, on the `SelectielijstAPIConfig` model add a reference to the identifier https://github.com/open-zaak/open-notificaties/blob/main/src/nrc/setup_configuration/models.py#L10 and do something like this in the step to get the service: https://github.com/open-zaak/open-notificaties/blob/main/src/nrc/setup_configuration/authorization.py#L9

That way we don't duplicate the code for creating `Services` and if any changes are made to the `Service` model, it can be tackled in the library's step right away and no code in the projects has to be changed

src/openzaak/config/setup_configuration/steps/selectielijst.py
```diff
@@ -0,0 +1,54 @@
...
+        if not created:
+            service.oas = model.SELECTIELIJST_API_OAS
+            service.save(update_fields=["oas"])
+
+        # 2. Set up configuration
+        config = ReferentieLijstConfig.get_solo()
+        if (
+            getattr(getattr(config, "service", None), "api_root", None)
+            != model.api_root
+        ):
```
stevenbal, line 50: We want to make sure the service is always set when `setup_configuration` is run (it should always overwrite the database state), so I think you can remove the if statement.

src/openzaak/conf/includes/base.py
```diff
@@ -290,12 +290,11 @@
 # Django setup configuration
 #
 SETUP_CONFIGURATION_STEPS = [
-    "openzaak.config.bootstrap.site.SiteConfigurationStep",
-    "openzaak.config.bootstrap.notifications.AuthNotificationStep",
-    "openzaak.config.bootstrap.notifications.NotificationsAPIConfigurationStep",
-    "openzaak.config.bootstrap.selectielijst.SelectielijstAPIConfigurationStep",
-    "openzaak.config.bootstrap.demo.DemoUserStep",
-    "openzaak.config.bootstrap.authorizations.AuthorizationConfigurationStep",
+    "django_setup_configuration.config.sites.steps.SitesConfigurationStep",
```
stevenbal, line 293: The command `python src/manage.py setup_configuration --yaml-file docker/setup_configuration/data.yml` currently crashes: 

```
CommandError: Unable to import step django_setup_configuration.config
```

```suggestion
    "django_setup_configuration.contrib.sites.steps.SitesConfigurationStep",
```

Also `zgw-consumers` has to be upgraded to >=0.36.0 to be able to use the `ServiceConfigurationStep`

docker/setup_configuration/data.yml
```diff
@@ -0,0 +1,25 @@
+sites_config_enable: true
```
stevenbal, line 1: Could you also add data for the `Selectielijst` (an extra step service in `zgw_consumers` and the config) here?

stevenbal (changes requested):

src/openzaak/conf/includes/base.py
Coperh: I thought we had a test but I remember that I deleted it.

Maybe an action that runs `python src/manage.py setup_configuration --yaml-file docker/setup_configuration/data.yml` would be useful

stevenbal: > Maybe an action that runs python src/manage.py setup_configuration --yaml-file docker/setup_configuration/data.yml would be useful

I think as a sanity check that would be good, maybe it could be similar to the `quickstart` job in objects: https://github.com/maykinmedia/objects-api/blob/master/.github/workflows/quick-start.yml

src/openzaak/config/setup_configuration/steps/selectielijst.py
```diff
@@ -0,0 +1,40 @@
...
+
+def get_service(slug: str) -> Service:
+    """
+    Try to find a Service and re-raise DoesNotExist with the identifier
+    to make debugging easier
+    """
+    try:
+        return Service.objects.get(slug=slug)
+    except Service.DoesNotExist as e:
+        raise Service.DoesNotExist(f"{str(e)} (identifier = {slug})")
```
stevenbal, line 19: let's reraise this as `django_setup_configuration.exceptions.ConfigurationRunFailed`

src/openzaak/config/setup_configuration/steps/selectielijst.py
```diff
@@ -0,0 +1,40 @@
...
+from django_setup_configuration.configuration import BaseConfigurationStep
+from zgw_consumers.models import Service
+
+from openzaak.config.setup_configuration.models import SelectielijstAPIConfig
+from openzaak.selectielijst.models import ReferentieLijstConfig
+
+
+def get_service(slug: str) -> Service:
+    """
+    Try to find a Service and re-raise DoesNotExist with the identifier
```
stevenbal, line 13:
```suggestion
    Try to find a Service and re-raise ConfigurationRunFailed with the identifier
```

stevenbal (changes requested):
@Coperh the files under `docker/data` can be removed and the volume mapping to it in `docker-compose.yml` can be removed as well, since they are related to the deprecated configurationstep.

src/openzaak/conf/includes/base.py
stevenbal: created a separate issue for this https://github.com/maykinmedia/open-api-framework/issues/96

stevenbal (approved):

PR #1849, opened 2025-01-06
:construction_worker: add OAF auto-update action

Coperh:
Partially fixes: maykinmedia/open-api-framework#44


Coperh:
No update to OAF in a while, so it has not really been tested in _production_ since merging in Objects, Open Klant or Referentielijsten.

But I think its low risk so could be added here anyway.

PR #1858, opened 2025-01-15
[#1827] Add relevante zaak relatie aard 'Overig' & toelichting field

### LINKED ISSUE #1827: Zaken API: extend the posible relationships between zaken (experimental)
### Thema / Theme

Zaken API

Currently, the relations between zaken that can be specified with `relevanteAndereZaken` are restricted by an enum (onderwerp, vervolg, bijdrage). We would like to extend this enum with "overig". In addition, to be able to furter specify the relation, an additional attribute `overigeRelatie` (or some other better attribute name) is added. This is a string.

Zaakbehandelaars can freely define other types of relations between zaken that better fit the actual relation that is being made.

The relations that are created with "overig" are loose, which means that zaken do not have to wait on each other to be finished, and there are no consequences for records management.

This proposal is not conform the standard which makes it experimental.

I wil propose the same change to the standard at the gemma-zaken repo as well.

Floris272:
closes #1827 

**Changes**

- Adds  relevante zaak relatie aard 'Overig'
- Adds toelichting CharField & overige_relatie Charfield to  RelevanteZaakRelatie
- Adds validation to make overige_relatie required when aard_relatie is overig.

**Questions**
- ~~Not sure about adding the toelichting field to the API.~~
- ~~Should the toelichting field be required when relatie aard is set to overig?~~
- The validation for the admin was added to the admin form, is this the correct location? 
- Looking at other validation i found some stuff that is only added the API and not the admin 
    - for example `RolIndicatieMachtigingValidator` adds `indicatieMachtiging must be set to 'gemachtigde' when a representee is specified` but on the role admin (or model) there does not seem to be any similar validation. Should this be added?


Hugo-ter-Doest:
Additionally, I proposed an attribute `overigeRelatie` to specify a name for the relation. `overigeRelatie` is required, `toelichting` is optional.

src/openzaak/components/zaken/admin/zaken.py
```diff
@@ -419,6 +420,13 @@ def clean(self):
                 "selecteer een zaak of vul een externe URL in."
             )
 
+        if cleaned_data.get(
+            "aard_relatie"
+        ) == AardZaakRelatie.overig and not cleaned_data.get("overige_relatie"):
+            raise forms.ValidationError(
+                "Het veld overige relatie is verplicht als de relatie aard 'overig' is."
```
stevenbal, line 427:
```suggestion
                "Het veld `overigeRelatie` is verplicht als de relatie aard `overig` is."
```

src/openzaak/components/zaken/api/validators.py
```diff
@@ -620,3 +620,22 @@ def __call__(self, attrs: dict):
             and indicatie_machtiging != IndicatieMachtiging.gemachtigde
         ):
             raise serializers.ValidationError(self.message, code=self.code)
+
+
+class OverigeRelevanteZaakRelatieValidator:
+    code = "overigerelatie-required"
+    message = _("overige relatie is verplicht als de relatie aard 'overig' is.")
```
stevenbal, line 627:
```suggestion
    message = _("`overigeRelatie` is verplicht als de relatie aard overig is.")
```

src/openzaak/components/zaken/models/zaken.py
```diff
@@ -548,6 +548,27 @@ class RelevanteZaakRelatie(models.Model):
         ),
     )
 
+    overige_relatie = models.CharField(
+        max_length=100,
+        verbose_name=_("overige relatie"),
+        blank=True,
+        null=True,
```
stevenbal, line 555: Generally it's better to not make `CharField`s (or other string based fields) nullable, because this means that there can be two different values for "no data": an empty string `""` and `null` (see https://docs.djangoproject.com/en/4.2/ref/models/fields/#null)
```suggestion
```

src/openzaak/components/zaken/models/zaken.py
```diff
@@ -548,6 +548,27 @@ class RelevanteZaakRelatie(models.Model):
...
+        help_text=mark_experimental(
+            "naam van de overige relatie. Verplicht bij relatie aard 'Overige'."
+        ),
+    )
+
+    toelichting = models.CharField(
+        max_length=255,
+        verbose_name=_("toelichting"),
+        blank=True,
+        null=True,
```
stevenbal, line 565:
```suggestion
```

src/openzaak/components/zaken/models/zaken.py
```diff
@@ -548,6 +548,27 @@ class RelevanteZaakRelatie(models.Model):
         ),
     )
 
+    overige_relatie = models.CharField(
+        max_length=100,
+        verbose_name=_("overige relatie"),
+        blank=True,
+        null=True,
+        help_text=mark_experimental(
+            "naam van de overige relatie. Verplicht bij relatie aard 'Overige'."
```
stevenbal, line 557:
```suggestion
            "Naam van de overige relatie. Verplicht bij relatie aard `overige`."
```

src/openzaak/components/zaken/models/zaken.py
```diff
@@ -548,6 +548,27 @@ class RelevanteZaakRelatie(models.Model):
...
+    )
+
+    toelichting = models.CharField(
+        max_length=255,
+        verbose_name=_("toelichting"),
+        blank=True,
+        null=True,
+        help_text=mark_experimental(
+            "Een toelichting op de aard van de relatie tussen beide ZAKEN. "
+            "(Vooral bedoeld in combinatie met relatie aard 'Overig')"
```
stevenbal, line 568:
```suggestion
            "(vooral bedoeld in combinatie met relatie aard `overig`)"
```

src/openzaak/components/zaken/tests/admin/test_relevantezaakrelatie_model_form.py
```diff
@@ -48,3 +48,25 @@ def test_relevantezaakrelatie_form_clean_throws_exception_if_relevant_zaak_and_r
...
+        form = RelevanteZaakRelatieForm()
+        form.cleaned_data = {
+            "_relevant_zaak": 1,
+            "aard_relatie": "overig",
+            "overige_relatie": "overig",
+        }
+        try:
+            form.clean()
+        except forms.ValidationError:
+            self.fail("Exception was raised in clean function when it should not have")
```
stevenbal, line 72: I think the try except is not strictly necessary, because if the `.clean()` does raise an error, the test will fail due to that error anyway

stevenbal (changes requested):
This needs to be rebased on main to fix the merge conflicts

@Floris272 regarding your questions in the description:

The validation should be implemented in both the API and admin, this is a known issue (for all components) because the admin interface operates directly on the database (and doesn't use the API). Ideally in the future we would have an admin interface that uses the API as well, so that we don't have to duplicate this validation, but I'm not sure if we will get to that soon

src/openzaak/components/zaken/models/zaken.py
```diff
@@ -552,6 +552,25 @@ class RelevanteZaakRelatie(models.Model):
         ),
     )
 
+    overige_relatie = models.CharField(
+        max_length=100,
+        verbose_name=_("overige relatie"),
+        blank=True,
+        help_text=mark_experimental(
+            "Naam van de overige relatie. Verplicht bij relatie aard `overige`."
```
stevenbal, line 560: Missed this on the previous review
```suggestion
            "Naam van de overige relatie. Verplicht bij relatie aard `overig`."
```

src/openzaak/components/zaken/constants.py
```diff
@@ -87,6 +89,7 @@ class AardZaakRelatie(models.TextChoices):
             "Aan het bereiken van de uitkomst van de andere zaak levert de onderhanden zaak een bijdrage."
         ),
     )
+    overig = ("overig", mark_experimental("Overig"))
```
stevenbal, line 92: The string passed to `mark_experimental` must explicitly be marked as translatable in order for it to be picked up by `makemessages`
```suggestion
    overig = ("overig", mark_experimental(_("Overig")))
```

stevenbal (approved):
Two minor comments remaining, looks good otherwise

PR #1860, opened 2025-01-20
:sparkles:[#1836] add zaak identity reservation endpoint

### LINKED ISSUE #1836: Introduce a case number reservation endpoint
### Thema / Theme

Zaken API

Currently, the ZGW-API lacks a dedicated service for reserving a unique case number before the actual case creation, unlike the functionality offered by the StUF standard. As a result, a component handling a customer’s request cannot immediately provide the definitive case number to the client. Instead, Open Forms displays its own internally generated ID, leading to the creation of two separate identifiers: the interim Open Forms ID and, later, the Open Zaak ID. This scenario conflicts with the service guidelines of the municipality of Rotterdam, which require that all customer communications reference a single unique case identifier. The current approach introduces confusion and detracts from the clarity and trustworthiness of the client-facing experience.

By introducing a service call to Open Zaak that allows for reserving a unique case number upfront, the customer can be immediately informed of the final, definitive case identifier. This not only removes the need for an interim ID but also ensures uniformity, consistency, and transparency in all communications. The improved clarity will streamline the customer journey, boost trust in the service’s reliability, and ultimately enhance the overall customer experience.

The proposed solution involves implementing a new endpoint in Open Zaak dedicated to reserving case numbers. Once a number is reserved, it cannot be reused, ensuring the integrity and uniqueness of case identifiers. Adopting this approach aligns with practices previously available in the StUF standard, maintains standardization across the ecosystem, and respects the municipality’s guidelines for customer communication by relying on a single, authoritative case reference.

Coperh:
Fixes #1836


src/openzaak/components/zaken/api/serializers/zaken.py
```diff
@@ -167,6 +168,37 @@ def update(self, instance, data):  # pragma:nocover
...
+            },
+        }
+
+    def create(self, validated_data):
+
+        current_date = date.today()
+
+        return self.Meta.model.objects.generate(
+            validated_data["bronorganisatie"],
+            current_date,
```
Coperh, line 195: I'm not sure if we should bother allowing the date, since it only seems to use the year.
Would someone want to reserve  a zaak for a future year or is the date use elsewhere?

Also current_date variable is not necessary and I'll remove it with the translation

src/openzaak/components/zaken/api/viewsets.py
```diff
@@ -363,6 +366,31 @@ def create(self, request, *args, **kwargs):
...
+        "zaak_reserve_identity",
+        summary="Reserve an zaak Identity.",
+        description=(
+            "Reserve a zaak identity without creating a case."
+            "This identity then cannot be generated by another reservation or zaak creation."
+            "The zaak first created with the reserved identity will be the owner and the indentity cannot be resued."
+        ),
+        request=ReserveZaakIdentificatieSerializer,
+        responses={
+            status.HTTP_200_OK: ReserveZaakIdentificatieSerializer(),
```
Coperh, line 379: This should be 201 actually. Again I'll change with the translation

Coperh:
Also I forgot to add model tests

src/openzaak/components/zaken/openapi.yaml
```diff
@@ -10745,6 +10745,195 @@ paths:
               schema:
                 $ref: '#/components/schemas/Fout'
           description: Precondition failed
+  /zaken/reserve_identiteit:
```
alextreme, line 10748: reserveer_zaaknummer

src/openzaak/components/zaken/models/zaken.py
```diff
@@ -458,6 +458,14 @@ def save(self, *args, **kwargs):
                 date=self.registratiedatum,
             )
             self.identificatie = self.identificatie_ptr.identificatie
+        elif not self.identificatie_ptr_id:
+
+            reserved_identificatie = ZaakIdentificatie.objects.filter(
```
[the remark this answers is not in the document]
Coperh: I used to filter to avoid an object does not exist error instead of catching it.

identificatie_ptr_id is only when the zaak is created. Its a weird one-to-one field with `parent_link` so it is never none after creation.

The query returns no results if an identificatie is given when creating an object the old way. If the ZaakIdentificatie exists i.e. it  was created with the reserve endpoint or manually, it should return in the queryset.

src/openzaak/components/zaken/tests/test_zaken.py
```diff
@@ -714,6 +715,83 @@ def test_create_zaak_with_incorrect_local_zaaktype(self):
...
+            "startdatum": "2025-02-04",
+        }
+
+        create_response = self.client.post(
+            reverse("zaak-list"), zaak_data, **ZAAK_WRITE_KWARGS
+        )
+        self.assertEqual(create_response.status_code, status.HTTP_201_CREATED)
+
+        self.assertEqual(Zaak.objects.count(), 1)
+        self.assertEqual(ZaakIdentificatie.objects.count(), 2)
```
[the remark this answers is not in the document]
Coperh: Added test to show the Zaak's id object is different from the origin one created

src/openzaak/components/zaken/api/viewsets.py
```diff
@@ -363,6 +367,31 @@ def create(self, request, *args, **kwargs):
 
         return super().create(request, *args, **kwargs)
 
+    @extend_schema(
+        "zaak_reserveer_zaaknummer",
+        summary="Reserve a zaak number.",
```
stevenbal, line 372:
```suggestion
        summary="Reserveer a zaaknummer",
```

src/openzaak/components/zaken/tests/test_zaken.py
```diff
@@ -714,6 +715,88 @@ def test_create_zaak_with_incorrect_local_zaaktype(self):
...
+
+        zaak = Zaak.objects.get()
+        self.assertNotEqual(zaak.identificatie, reserved_zaak_id)
```
stevenbal, line 798: You're now comparing the identificatie string (e.g. `ZAAK-2025-0000000001`) to a `ZaakIdentificatie` object. I think this should probably be two assertions, because the zaaknummer is the same, but the `ZaakIdentificatie` object is different (because of the different organisatie):

```python
self.assertEqual(zaak.identificatie, zaak_id)
self.assertNotEqual(zaak.identificatie_ptr, reserved_zaak_id)
```

stevenbal (approved):
Just the comment about the assertion, looks good otherwise :+1:

src/openzaak/components/zaken/api/viewsets.py
Coperh: I assume it should be _een_ and not _a_

stevenbal: you're right, missed that

PR #1863, opened 2025-01-21
add PUT endpoint for rollen

### LINKED ISSUE #1835: Zaken API: PATCH en PUT for rol
### Thema / Theme

Zaken API

Is it possible to add support foor PATCH and PUT calls for the `rol` object in the zaken API?
In our use case we need to update the `betrokkene` field in the rol object. This field is used to store a url reference to a partij in open-klant, which is used the the OutputManagementComponent (OMC) for sending notification to NotifyNL. Currently we do a DELETE and POST, but for obvious reasons we (Gemeente Nijmegen) do not want to do this in production.

This is already implemented during the hackathon in commits:
- https://github.com/open-zaak/open-zaak/commit/ed4b997b621806acff5bc886f98f2fd80de4a2a0
- https://github.com/open-zaak/open-zaak/commit/dd8851112a6db8929efc4363c3ca2143ba03c5ab

_No response_

_No response_

annashamray:
Fixes #1835

### DIFF
--- a/src/openzaak/components/zaken/api/mixins.py
+++ b/src/openzaak/components/zaken/api/mixins.py
@@ -84,9 +84,9 @@ class UpdateOnlyModelMixin:
     """
 
     def update(self, request, *args, **kwargs):
-        partial = kwargs.pop("partial", False)
+        kwargs.pop("partial", False)
         instance = self.get_object()
-        serializer = self.get_serializer(instance, data=request.data, partial=partial)
+        serializer = self.get_serializer(instance, data=request.data, partial=False)
         serializer.is_valid(raise_exception=True)
         self.perform_update(serializer)
 
--- a/src/openzaak/components/zaken/api/mixins.py
+++ b/src/openzaak/components/zaken/api/mixins.py
@@ -5,6 +5,7 @@ from typing import Optional
 from django.db import models
 
 from rest_framework import serializers
+from rest_framework.response import Response
 
 from openzaak.components.zaken.api.scopes import SCOPE_ZAKEN_GEFORCEERD_BIJWERKEN
 from openzaak.components.zaken.models import Zaak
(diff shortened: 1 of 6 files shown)


src/openzaak/components/zaken/api/mixins.py
```diff
@@ -74,3 +75,27 @@ def perform_destroy(self, instance: models.Model) -> None:
         zaak = instance.zaak
         self._check_zaak_closed(zaak)
         super().perform_destroy(instance)
+
+
+class UpdateOnlyModelMixin:
+    """
+    Update mixin, which supports PUT but doesn't support PATCH
+    Copied from DRF
```
Coperh, line 83: How is PUT supported but not PATCH? Does patch use PATCH update and since its missing, it is not supported?

src/openzaak/components/zaken/api/mixins.py
```diff
@@ -74,3 +75,27 @@ def perform_destroy(self, instance: models.Model) -> None:
...
+    """
+
+    def update(self, request, *args, **kwargs):
+        partial = kwargs.pop("partial", False)
```
Coperh, line 87: What does this line/ partial do?

src/openzaak/components/zaken/api/serializers/zaken.py
```diff
@@ -1021,6 +1021,34 @@ def create(self, validated_data):
...
         return rol
 
+    @transaction.atomic
+    def update(self, instance, validated_data):
+        group_data = validated_data.pop("betrokkene_identificatie", None)
+        contactpersoon_rol = validated_data.pop("contactpersoon_rol", None)
+
+        # delete the existing betrokkene_identificatie instance
+        if not instance.betrokkene:
+            instance.betrokkene_identificatie.delete()
```
Coperh, line 1031: What if you switch from `betrokkene_identificatie` to `betrokkene`? Is it possible and will the `betrokkene_identificatie` still be there?

Coperh (changes requested):
A few comments plus there should be tests for the put endpoint

src/openzaak/components/zaken/api/mixins.py
annashamray: Yes
When the viewset is registered with DRF router it creates something like urlpattern and binds the HTTP methods with default action names, but only if these acions are provided in the viewset, therefore when the router doesn't doesn't find `partial_update` method for `RolViewSet` it will not create an endpoint for the PUT method. 

You can find it in DRF `SimpleRouter.get_method_map` code

src/openzaak/components/zaken/api/mixins.py
annashamray: Just a copy paste from DRF, it uses in the serializer for required field validation
But you are right, here it should always be False, I'll change it

src/openzaak/components/zaken/api/serializers/zaken.py
annashamray: It's possible to switch from `betrokkene_identificatie` to `betrokkene` and I've added a testcase for it (`RolTestCase.test_update_rol_from_betrokkene_identificatie_to_betrokkene`)

Coperh (approved):

PR #1867, opened 2025-01-24
Feature/1855 iotype filter on zaaktype

### LINKED ISSUE #1855: Filter informatieobjecttype by relation to zaaktype
### Thema / Theme

Catalogi API

Currently, for the statustypen and resultaattypen endpoints, it is possible to use the query param `zaaktype` to retrieve the statustypen and resultaattypen that are related to the zaaktype.

For the informatieobjecttypen, one has to filter the zaaktype-informatieobjecttypen on zaaktype, then with the URLs of the respective informatieobjecttypen retrieve the data of the informatieobjecttypen. This involves many calls to retrieve the informatieobjecttypen.

It would be great to be able to filter the informatieobjecttypen on zaaktype too.

Retrieve the informatieobjecttypen related to a zaaktype in one call.

This would be useful for open archiefbeheer in the context of this PR: https://github.com/maykinmedia/open-archiefbeheer/pull/604

annashamray:
Fixes #1855

### DIFF
--- a/src/openzaak/components/catalogi/api/filters.py
+++ b/src/openzaak/components/catalogi/api/filters.py
@@ -8,11 +8,13 @@ from django.db import models
 from django.utils.translation import gettext_lazy as _
 
 from django_filters import rest_framework as filters
+from drf_spectacular.utils import OpenApiTypes, extend_schema_field
 from vng_api_common.filters import URLModelChoiceFilter
 from vng_api_common.utils import get_help_text, get_resource_for_path
 
 from openzaak.utils.filters import CharArrayFilter
 from openzaak.utils.filterset import FilterSet
+from openzaak.utils.help_text import mark_experimental
 
 from ..models import (
     BesluitType,
@@ -219,10 +221,19 @@ class InformatieObjectTypeFilter(FilterSet):
         method=geldigheid_filter,
         help_text=DATUM_GELDIGHEID_HELP_TEXT,
     )
+    zaaktype = extend_schema_field(OpenApiTypes.URI)(
+        URLModelChoiceFilter(
+            field_name="zaaktypeinformatieobjecttype__zaaktype",
+            queryset=ZaakType.objects.all(),
+            help_text=mark_experimental(
+                get_help_text("catalogi.ZaakTypeInformatieObjectType", "zaaktype")
+            ),
+        )
+    )
 
     class Meta:
         model = InformatieObjectType
-        fields = ("catalogus", "status", "omschrijving", "datum_geldigheid")
+        fields = ("catalogus", "status", "omschrijving", "datum_geldigheid", "zaaktype")
 
 
 class BesluitTypeFilter(FilterSet):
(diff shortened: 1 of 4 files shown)


src/openzaak/components/catalogi/api/filters.py
```diff
@@ -219,10 +221,19 @@ class InformatieObjectTypeFilter(FilterSet):
         method=geldigheid_filter,
         help_text=DATUM_GELDIGHEID_HELP_TEXT,
     )
+    zaaktype = extend_schema_field(OpenApiTypes.URI)(
+        URLModelChoiceFilter(
+            field_name="zaaktypeinformatieobjecttype__zaaktype",
+            queryset=ZaakType.objects.all(),
+            help_text=mark_experimental(
```
stevenbal, line 228: This help_text doesn't seem to be picked up in the schema: 
image

src/openzaak/components/catalogi/tests/test_informatieobjecttype.py
```diff
@@ -740,6 +741,28 @@ def test_filter_geldigheid(self):
         self.assertEqual(len(data), 1)
         self.assertEqual(data[0]["url"], f"http://testserver{reverse(iotype)}")
 
+    @override_settings(ALLOWED_HOSTS=["openzaak.nl"])
+    def test_filter_zaaktype(self):
```
stevenbal, line 745: Could you add a test to check the behavior in case the zaaktype doesn't exist? E.g. the URL does not point to a zaaktype in the db

stevenbal (changes requested):

src/openzaak/components/catalogi/api/filters.py
annashamray: I face the same problem here as discussed in this issue - https://github.com/tfranzel/drf-spectacular/issues/1267

If `extend_schema_field` is not used then the schema is generate with `type: integer`
But if I used it then the description is not picked up (the current situation)
I can't find any elegant way to solve it. DO you have any ideas?

stevenbal: Hmm that's unfortunate, I tried some stuff as well, the only way I could get it to work was by explicitly adding the parameter to `@extend_schema_view` on the viewset (https://github.com/open-zaak/open-zaak/blob/main/src/openzaak/components/catalogi/api/viewsets/informatieobjecttype.py#L31):

```python
@extend_schema_view(
    list=extend_schema(
        summary="Alle INFORMATIEOBJECTTYPEn opvragen.",
        description="Deze lijst kan gefilterd wordt met query-string parameters.",
        parameters=[
            OpenApiParameter(
                name="zaaktype",
                type=OpenApiTypes.URI,
                location=OpenApiParameter.QUERY,
                description=mark_experimental(
                    get_help_text("catalogi.ZaakTypeInformatieObjectType", "zaaktype")
                ),
                required=False,
            ),
        ]
    ),
   ...
)
```

annashamray: Added in the explicit way

src/openzaak/components/catalogi/tests/test_informatieobjecttype.py
annashamray: The test is added

stevenbal (approved):

PR #1870, opened 2025-01-28
[#1831] Roltoelichting allowed to be an empty string

### LINKED ISSUE #1831: Zaken API: roltoelichting isn't conform the VNG referencecomponent and should be allowed to be an empty string

1.16.0

Taiga UWV #6

Currently with OpenZaak 1.16 the roltoelichting field in the Zaken API rollen endpoint must be filled in with a non-empty string:

image

OAS:

```
        roltoelichting:
          type: string
          minLength: 1
          maxLength: 1000
```

However, in the Zaken API referencecomponent of the VNG (version 1.5.1) roltoelichting is required but (based on the OAS and Redoc) can be an empty string. The above restriction also doesn't exist in the referencecomponent:

image

It also doesn't make sense to make roltoelichting required as it's usually not necessary, I see dummy input or even incorrect input (eg. a hardcoded value which doesn't correspond with the Roltype) in a few OZ installs.

The solution would be to allow empty strings for this field

_No response_

_No response_

danielmursa-dev:
Fixes #1831


src/openzaak/components/zaken/openapi.yaml
```diff
@@ -11225,7 +11223,6 @@ components:
             , vanuit het belang van BETROKKENE in haar ROL bij een ZAAK.
       required:
       - betrokkeneType
-      - roltoelichting
```
Coperh, line 11228: Is it no longer required? IDK what the spec says.

Coperh (commented):
Seems fine but I am unsure if `roltoelichting` is still meant to be required

src/openzaak/components/zaken/openapi.yaml
annashamray: It's required in the standard - https://github.com/VNG-Realisatie/zaken-api/blob/920ecf3b972dc7270a124ef55f60ba0ece69d588/src/openapi.yaml#L13074

annashamray (commented):
As Alex mentioned in the US there is no sense to make it required and allow a blank string, so I don't mind deviating from the standard here. But mentioning it in the help text would be nice

src/openzaak/components/zaken/models/zaken.py
```diff
@@ -805,7 +805,10 @@ class Rol(ETagMixin, APIMixin, models.Model):
         editable=False,
         db_index=True,
     )
-    roltoelichting = models.TextField(max_length=1000)
+    roltoelichting = models.TextField(
+        max_length=1000,
+        blank=True,
```
stevenbal, line 810: As @annashamray suggested, let's mention in the help text that this deviates from the standard
```suggestion
        blank=True,
        help_text=mark_experimental(_("Toelichting bij de rol (dit veld wijkt af van de standaard, omdat een lege string `\"\"` is toegestaan).")),

```

stevenbal (changes requested):

annashamray (approved):

stevenbal (approved):

PR #1872, opened 2025-01-29
[#1820] Check zaaktype in admin operations

### LINKED ISSUE #1820: Add validation when creating/updating a zaak via the admin
### Thema / Theme

Admin

Taiga UWV 14

Via the admin it is possible to set a status for a zaak which doesn't correspond with the zaaktype/statustype configuration.

Likewise, via the admin it is possible to add a rol with a roltype that doesn't correspond with the zaaktype

_No response_

_No response_

danielmursa-dev:
Fixes #1820


src/openzaak/components/zaken/models/zaken.py
```diff
@@ -635,6 +636,17 @@ def __str__(self):
     def unique_representation(self):
         return f"({self.zaak.unique_representation()}) - {self.datum_status_gezet}"
 
+    def clean(self):
```
Coperh, line 639: Could this be in the admin form instead of the model? 
This I think will be checked when creating an object the normal way I think while not in the API, which is what we want?

Coperh (commented):
I think the code and tests look fine but I have a few questions I don't know the answer to.

I think @annashamray or @stevenbal could answer better

src/openzaak/components/zaken/validators.py
```diff
@@ -0,0 +1,23 @@
+# SPDX-License-Identifier: EUPL-1.2
+# Copyright (C) 2019 - 2020 Dimpact
```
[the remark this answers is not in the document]
stevenbal: Not sure, I think this is for the Foundation for Public code, maybe @joeribekker knows?

src/openzaak/components/zaken/validators.py
```diff
@@ -0,0 +1,23 @@
...
+        self.resource = resource or url_field
+
+    def __call__(self, attrs):
+        url = attrs.get(self.url_field)
+        zaak = attrs.get(self.zaak_field)
+        if not url or not zaak:
+            return
+
+        if url.zaaktype != zaak.zaaktype:
+            raise ValidationError(self.message, code=self.code)
```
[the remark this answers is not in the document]
stevenbal: I also think it makes more sense to raise Django ValidationErrors from validators, that way they can be used outside of a DRF context as well, it's up to DRF to reraise it as a DRF ValidationError as @danielmursa-dev mentioned

src/openzaak/components/zaken/models/zaken.py
```diff
@@ -635,6 +636,17 @@ def __str__(self):
...
+        super().clean()
+        if (self._statustype or self._statustype_url) and self.zaak:
```
stevenbal, line 641: In what cases would neither `self._statustype` or `self._statustype_url` be defined?

src/openzaak/components/zaken/models/zaken.py
stevenbal: I think this is indeed only checked when using `ModelForms`, but since the validator is also part of the `validators` in each serializer, the validation will also be applied to the API, so I think it's okay to leave this as is

src/openzaak/components/zaken/models/zaken.py
```diff
@@ -897,6 +909,17 @@ def save(self, *args, **kwargs):
 
         super().save(*args, **kwargs)
 
+    def clean(self):
```
stevenbal, line 912: Could you also add this for the other models where this is used if it applies? E.g. I see that the validator is also used for ZaakEigenschap, Resultaat and ZaakObject

src/openzaak/components/zaken/tests/admin/test_rol_audittrail.py
```diff
@@ -160,3 +166,27 @@ def test_delete_rol(self):
         old_data = audittrail.oud
 
         self.assertEqual(old_data["roltoelichting"], "some desc")
+
+    def test_invalid_check_zaaktype(self):
```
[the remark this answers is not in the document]
stevenbal: Let's add a new file `test_rol_admin.py`

stevenbal (changes requested):

src/openzaak/components/zaken/models/zaken.py
annashamray: `StatusAdmin` uses `StatusForm` which checks that either `_statustype` or `_statustype_base_url` should be defined. So do all other admin forms. So now we have `clean()` defined both on admin level and model level. 
Should we use it only in one place? For example only in the admin forms? 

@stevenbal What do you think?

danielmursa-dev: I think I have found a solution, in `FkOrServiceUrlField` there is `_add_check_constraint`. It was added to check if both elements cannot be null, and this constraint is executed when running the `full_clean' of the model.

stevenbal: @annashamray I think it's a bit cleaner to have `clean` only on the admin forms.

@danielmursa-dev could you move this validation to the related admin form (e.g. `StatusForm.clean`, etc.)?

stevenbal (changes requested):
Only one mark remaining to move the check to `clean` in admin forms, looks good otherwise

src/openzaak/components/zaken/models/zaken.py
danielmursa-dev: @annashamray @stevenbal I tried to move and do the validation in the `AdminForm`, but it fails because the validator requires the instance (status_type example) already validated and checked. 

The `statustype` field is an `FkOrServiceUrlField()`, so it has two sources from which it can take the instance, and in the admin form, you only have validations on the input fields `FK` or the `URL` (actually two urls, base and relative, where base is a service), you can't get the final instance of the statustype, so I can think of 3 solutions at the moment:
- do the validation in the form, but only for the FK field object
- modify the validator, so that you get the instance inside it
- leave the validation in the clean of the model

What do you think?

stevenbal: let's keep it as it is now then, since moving it to the form makes it more annoying to deal with the `FKOrServiceUrlFields`

stevenbal (approved):

PR #1876, opened 2025-01-30
:sparkles: [#1809] Add experimental Rol.tijdvakGeldigheid attribute

### LINKED ISSUE #1809: Implementing validity period for ROL in ZRC-API
### Thema / Theme

Zaken API

Currently, the ZGW-API lacks support for capturing the validity period for parties involved in a case (ROL). This means that changes in an involved party's role are not historically tracked, risking the loss of key information over a case's lifecycle. In the StUF-ZKN standard, this is addressed using tijdvakGeldigheid, which includes beginGeldigheid (start) and eindGeldigheid (end) to indicate validity.

This construction is especially relevant for parties with roles such as "medewerer" or "organisatorische_eenheid," ensuring that the appropriate handler for a given phase of the case can be identified for customer service purposes.

**Request**
An extension of the ROL-schema, with tijdvakGeldigheid, so I can accurately track all involved parties (mvp for "medewerker" and "organisatorische_eenheid") over the case lifecycle. The validity period (beginGeldigheid and eindGeldigheid) can be specified for each involved party. 

For betrokkene "medewerker" and "organisatorische_eenheid" with the roltype "behandelaar", the system validates logical date sequences and prevents overlapping periods. Only one can be active at a given time.

- Historical: Track changes in party involvement over time, including role updates.
- Accurate: Organizations can generate precise reports on party involvement for accountability and audits.
- Standards alignment: Aligns with StUF-ZKN

_No response_

stevenbal:
Fixes #1809

**Changes**

* experimental Rol.beginGeldigheid and Rol.eindeGeldigheid attribute


stevenbal:
I created https://github.com/open-zaak/open-zaak/issues/1878, because I noticed that this is an existing issues with `GegevensGroepType` that can break `PUT` requests. I think this especially becomes an issue if we add this experimental field, that should be backwards compatible

src/openzaak/components/zaken/api/validators.py
```diff
@@ -639,3 +640,64 @@ def __call__(self, attrs, serializer):
...
+            or betrokkene_type != self.betrokkene_type
+            or roltype.omschrijving_generiek != RolOmschrijving.behandelaar
+        ):
+            return
+
+        begin_geldigheid = tijdvak_geldigheid.get("begin_geldigheid")
+        eind_geldigheid = tijdvak_geldigheid.get("eind_geldigheid")
+
+        # TODO workaround for https://github.com/open-zaak/open-zaak/issues/1878
+        if not begin_geldigheid:
```
annashamray, line 679: Should we raise an validation error if `begin_geldigheid` is empty but `eind_geldigheid` is not empty?

src/openzaak/components/zaken/tests/utils.py
```diff
@@ -188,12 +188,14 @@ def get_statustype_response(statustype: str, zaaktype: str, **overrides) -> dict
...
-def get_roltype_response(roltype: str, zaaktype: str):
+def get_roltype_response(
+    roltype: str, zaaktype: str, omschrijving_generiek: str = "adviseur"
+):
     return {
         "url": roltype,
         "zaaktype": zaaktype,
         "omschrijving": "some role",
-        "omschrijvingGeneriek": "adviseur",
+        "omschrijvingGeneriek": omschrijving_generiek,
```
annashamray, line 198: adding `**overrides` is also an option here

src/openzaak/components/zaken/api/validators.py
```diff
@@ -639,3 +640,64 @@ def __call__(self, attrs, serializer):
...
+        tijdvak_geldigheid = get_from_serializer_data_or_instance(
+            "tijdvak_geldigheid", attrs, serializer
+        )
+
+        # Validation does not apply if no tijdvakGeldigheid is supplied, to make it
+        # a backwards compatible experimental feature
+        if (
+            not tijdvak_geldigheid
+            or betrokkene_type != self.betrokkene_type
+            or roltype.omschrijving_generiek != RolOmschrijving.behandelaar
```
annashamray, line 671: I think it is like extension of `RolOccurenceValidator` but with dates?

annashamray (dismissed):
The main question I have is for business design of this feature, it looks like it overlaps with the existing business rules or Rol and Roltype resources without regarding them

What we have now in the Zaken and Catalogi API standards:
* each Roltype has writable `begin_geldigheid` and `einde_geldigheid` attributes 
* only one rol for one zaak can has a resultaattype `initiator` or `zaakcoordinator`

What should we do with these rules with relation to this new feature?
1. How do Roltype and Rol validity dates depend on each other? Is it possible for Rol date range to be bigger than the related roltype date range? 
2. What should we do with `initiator` and `zaakcoordinator` occurance restriction? Should we also introduce the condition with overlapping dates for them? 
3. Do we really need rol validity dates? Can we use the existing roltype date attributes? And use them for the `behandelaar` occurrence validation? 

Sorry if it's not a good timing for such questions, since the PR is already done, but I still think they should be addressed.

src/openzaak/components/zaken/api/validators.py
stevenbal: I forgot to remove this validator, it was mentioned in refinement that we won't do validation on this, just record allow the validity period to be recorded https://github.com/open-zaak/open-zaak/issues/1809#issuecomment-2647716834

src/openzaak/components/zaken/api/validators.py
stevenbal: as mentioned I'll remove this validator

stevenbal:
1. I'll have to look into this, I'm not sure if there is any validation currently on Roltype.beginGeldigheid / eindeGeldigheid
2. As mentioned we won't put restrictions based on the validity period in the API itself, if clients want this they should enforce it themselves https://github.com/open-zaak/open-zaak/issues/1809#issuecomment-2647716834
3. I don't know this for certain, but Roltype.beginGeldigheid / eindeGeldigheid to me would indicate that Rollen of this type can be created in this timeframe, whereas Rol.tijdvakGeldigheid says something about when this specific Rol was in place for a Zaak (e.g. if there is a Zaak that between 2020-01-01 and 2020-06-01 had behandelaar A and after 2020-06-01 had behandelaar B, you can indicate that using tijdvakGeldigheid)

src/openzaak/components/zaken/api/serializers/zaken.py
```diff
@@ -939,7 +940,40 @@ class Meta:
...
-class RolSerializer(PolymorphicSerializer):
+class RolTijdvakGeldigheidSerializer(GegevensGroepSerializer):
+    class Meta:
+        model = Rol
+        gegevensgroep = "tijdvak_geldigheid"
+        # TODO workaround to make sure this is backwards compatible
+        # see: https://github.com/open-zaak/open-zaak/issues/1878
+        extra_kwargs = {
+            "begin_geldigheid": {"allow_null": True},
+            "eind_geldigheid": {"allow_null": True},
```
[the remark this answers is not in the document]
stevenbal: @joeridiederen would it be okay if instead of adding `Rol.tijdvakGeldigheid` we add the fields `Rol.beginGeldigheid` and `Rol.eindeGeldigheid`?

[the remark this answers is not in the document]
stevenbal: changed this to flat attributes as Anna suggested

src/openzaak/components/zaken/tests/utils.py
stevenbal: reverted this change because the validation was removed, so this is no longer needed in the tests

danielmursa-dev (approved):
Good!

PR #1891, opened 2025-02-07
:wrench: Use new issue types in templates instead of labels

stevenbal:
I also added types to the issues that are currently open

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,7 +1,9 @@
 name: Probleem melden / Bug report
 description: Meldt een probleem om ons te helpen verbeteren  / Create a report to help us improve
 title: "Title here"
+type: "Bug"
 labels: ["bug", "triage"]
+projects: ["maykinmedia/15"]
 assignees: []
 body:
   - type: input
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -1,7 +1,9 @@
 name: Probleem melden / Bug report
 description: Meldt een probleem om ons te helpen verbeteren  / Create a report to help us improve
 title: "Title here"
-labels: ["bug", "triage"]
+type: "Bug"
+labels: ["triage"]
```
annashamray, line 5: I personally filter on labels in the project board, so I'd keep "bug" label for this purpose

stevenbal: hmm good point, I'll see if I can add the issue type to the display in the project board as well, that way you can do the same filtering but via the `Bug` issue type instead of the label :)

stevenbal: I added the types to the project board display, currently these types aren't assigned yet for repos other than OZ/ON, but I think I'll have this done sometime next week.

I'll leave the bug and enhancement labels in place for now, sometime in the future we might be able to remove them, but there's no rush to that imo

annashamray: Sounds good!

annashamray, line 5:
```suggestion
labels: ["bug", "triage"]
```

annashamray (approved):

.github/ISSUE_TEMPLATE/bug_report.yml
stevenbal: re-added the labels for now

PR #1899, opened 2025-02-11
:construction_worker: [#1712] Run tests in parallel in CI

### LINKED ISSUE #1712: As developer I want to reduce the time of CI action
### Thema / Theme

Other

Run CI action takes arounf 45 minutes now, so there is a big room for improvement here

- [x] Run tests in parallel
- [x] see if setUp can be replaced with setUpTestData
- [ ] Squash migrations to reduce migration time
- [x] See if `pip install ...` can be replaced with `uv pip install ...`
- [x] Check if postgres settings in CI can be optimized

stevenbal:
Closes #1712 partially

Reduces `Run CI` job time from ~53 minutes to  ~21 minutes

I will reopen #1712 after this is merged, because I want to look into some other ways to speed up CI (squashing migrations, cleaning up tests, etc.)

**Changes**

* Run tests in parallel in CI

**Checklist**

- Commit hygiene


danielmursa-dev (commented):
I have run a test locally and it almost always works, except in these cases:

danielmursa-dev (approved):

.github/workflows/ci.yml
```diff
@@ -136,50 +206,78 @@ jobs:
...
       - name: Run tests
         if: ${{ !(matrix.cmisurlmapping == 'True' && matrix.binding == 'BROWSER') }}
         run: |
+          echo "# Profiling stats" >> $GITHUB_STEP_SUMMARY
+          echo "" >> $GITHUB_OUTPUT
+
+          python src/manage.py compilemessages
           python src/manage.py collectstatic --noinput --link
-          coverage run src/manage.py test src
+          coverage run src/manage.py test src --tag cmis
```
annashamray, line 237: Am I correct that CMIS tests are not run in parallel? Afaik CMIS tests are quite slow, perhaps in the future running them in parallel could also speed up the process

src/openzaak/utils/cache.py
```diff
@@ -110,13 +111,28 @@ def requests_cache_enabled(cache_name, backend=None, *args, **kwargs):
 
     original_client = vng_api_common.client.Client
 
+    cache_name = "import_requests"
```
annashamray, line 114: There are two options to set up cache backend for requests_cache now:
* `'import_requests'` constant (like here and in CACHE setting
* `IMPORT_REQUESTS_CACHE_NAME` setting like default in DjangoRequestsCache 

It seems possible that if IMPORT_REQUESTS_CACHE_NAME is configured the default requests-cache backend could appear again. Anyway I think IMPORT_REQUESTS_CACHE_NAME can be removed to reduce this probability

src/openzaak/conf/dev.py
```diff
@@ -9,7 +9,9 @@
 from urllib3.exceptions import SystemTimeWarning
 
 os.environ.setdefault("DEBUG", "yes")
-os.environ.setdefault("ALLOWED_HOSTS", "*")
+if "test" not in sys.argv:
+    os.environ.setdefault("ALLOWED_HOSTS", "*")
```
annashamray, line 13: Why is this needed?

annashamray (changes requested):

.github/workflows/ci.yml
stevenbal: yes, I tried running them in parallel, but since they all operate on the same docker container I wasn't able to get it to work. Also since the total time for them to complete is less than the non-CMIS part of the test suite, I figured it wasn't worth my time to try and optimize this

src/openzaak/utils/cache.py
stevenbal: Hmm yeah I think I might just make it always use the same cache name + backend without making it configurable, to avoid this issue

stevenbal:
@danielmursa-dev does the error on `openzaak.components.documenten.tests.test_upload.SmallFileUpload.test_create_empty_file` also occur if you run that single test, or only if you run everything in parallel? Also if it does, it might be good to check if these same thing happens if you change `SENDFILE_BACKEND` envvar to `django_sendfile.backends.simple`

src/openzaak/conf/dev.py
stevenbal: I forgot to add a comment, setting `ALLOWED_HOSTS` to `*` causes issues with django-loose-fk, I'll see if I can pin down the failing tests and add an override_settings for ALLOWED_HOSTS instead

stevenbal: removed this and added `override_settings` where needed

annashamray (approved):
Looks good!

Issue #1017, opened 2021-08-04
Catalogi import should run asynchronously


sergei-maertens:
The default request-response time-out is set to about 60s. The catalogi import (a full catalogue) runs longer than that if you have some real data (tested with Utrecht proeftuin catalogus).

A solution would be to run the import in a celery task and have the admin/ui poll the completion status of the import task.

Currently the web-interface for export/import is not usable in this form and I had to resort to using the CLI, which is not a pleasant experience on Docker/Kubernetes.

ericherman:
Making this asynchronous is clearly the way to go.

A follow-on could be to consider how to notify of success/failure upon completion, but notification should not be required for the first asynch implementation.

sergei-maertens:
We could simply track import runs in a database table and show success/failure in there, as users may run into validation errors (spent the better part of the day fixing those yesterday). That would be the MVP approach for me.

ericherman:
Sounds good.

The very first version may not even need the logging, but recording what happened shouldn't create a lot of scope creep either. 

Even if you deliver it in one chunk, I suggest transforming any add-on ideas -- especially thoughts like this which are not coming from someone with access to the purse strings -- into distinct deliverables/todo-items/features/whatever to avoid snowballing requirements from the peanut gallery.

sergei-maertens:
Agree, however this feedback is really about actually making the interface useful. Making it async so that the import actually completes is worthless if there are errors in the import data and the end-user cannot see what the errors are :-)

I'll add a checklist with MVP actions

PR #1914, opened 2025-02-20
:bug: [#1907] Make sure requests_cache is not applied globally

### LINKED ISSUE #1907: requests-cache uses SQLite backend in some cases

1.17.0

AMS #24

In some cases `requests-cache` uses the default backend (SQLite) instead of the django cache backend. 
It started happening after the OZ client was refactored for zds_client to ape_pie.APIClient

Possible fix - https://github.com/open-zaak/open-zaak/pull/1899/commits/838c4ca69a11f51458584cdfaa1c2b19fea87775#diff-4883068558e45ee8cd89e8c1e4118f2eedc574b6cedfec18b8a97f9c9b3ec99eR123-R126

**Needs further analysis:**

* how to reproduce and test it locally
* Why OIDC and POST ZIO endpoints are affected by it  

One of this endpoints could possible cause this error:
* POST /admin/catalogi/catalogus/{catalogus_pk}/zaaktype/import/
* POST /zaken/api/v{version}/zaakinformatieobjecten
* GET /oidc/callback/

```
OperationalError

unable to open database file
```

stevenbal:
Closes #1907

This is a temporary fix for the issue with requests_cache being applied globally. I think the proper way to fully fix this is to run imports in Celery tasks, that way we both ensure that they are run in a separate process and we can run them asynchronously, which is a better user experience if imports take some time to complete (https://github.com/open-zaak/open-zaak/issues/1017)

Running things in Celery does mean we have to properly provide feedback to the user (is the task still running, did it fail, etc.) and I do not want to rush that functionality for the sake of fixing this issue, so that's why I want to apply this bandaid fix first.

**Changes**

* Make sure requests_cache is not applied globally

**Checklist**

- Commit hygiene

### DIFF
--- a/src/openzaak/components/catalogi/admin/utils.py
+++ b/src/openzaak/components/catalogi/admin/utils.py
@@ -13,7 +13,7 @@ from dateutil.relativedelta import relativedelta
 from rest_framework.test import APIRequestFactory
 from rest_framework.versioning import URLPathVersioning
 
-from openzaak.utils.cache import requests_cache_enabled
+from openzaak.utils.cache import run_in_process_with_caching
 
 from ..api import serializers
 from ..models import BesluitType, Catalogus, InformatieObjectType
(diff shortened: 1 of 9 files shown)


stevenbal:
Not sure why codecov is not picking up the `wrapper` part inside `run_in_process_with_caching`, because I have explicitly added tests for this

Issue #1908, opened 2025-02-17
Removal of CMIS from Documents API?


alextreme:
### Thema / Theme

Zaken API

### Omschrijving / Description

To be discussed after Joeri's holiday

alextreme:
Removal of CMIS support has been communicated via the mailinglist and is to be completed in Q3 2025.

PR #1931, opened 2025-03-07
[#1856] upgrade django to 5.1

### LINKED ISSUE #1856: Open Zaak Django 5.1/5.2 upgrade
### Thema / Theme

Other

From #1852 for trying out Postgresql connection pooling:

This requires an upgrade of Open API Framework to work with Django 5.1. After that Open Zaak can be updated to use the new OAF. Complexity however in the zgw-consumers / commonground-api-common libraries, which will also require upgrading. Estimated at 3-4 weeks

---

Django 5.2 will be out in april so we will have to upgrade the various components in due time

Tasks:
* [ ] update OAF - https://github.com/maykinmedia/open-api-framework/pull/113
* [ ] update django-setup-configuration - https://github.com/maykinmedia/django-setup-configuration/pull/66
* [ ] update OZ

Floris272:
Closes #1856 

**Changes**

- Upgrades django to 5.1.6

**Checklist**

- Commit hygiene


Floris272:
image

@stevenbal `inline-unsafe` seems to already have been added to `CSP_SCRIPT_SRC` by open-api-framework.

stevenbal:
Locally I only got this to work if I do `CSP_INCLUDE_NONCE_IN.remove("script-src")` in `base.py` and add `"https://*.earthdata.nasa.gov"` to `CSP_IMG_SRC`. I tried adding the nonce to the script by overriding the widget template that adds this inline script, but I don't think this template has access to the `request` context var (which is needed to set `<script nonce="{{ request.csp_nonce }}">`) https://github.com/django/django/blob/main/django/contrib/gis/templates/gis/openlayers.html

I also looked at the custom widget we currently use and I don't think we need to override `get_context` or `deserialize` anymore. If you save a Zaak that has coordinates, the latitude and longitude are flipped each time you save them with the current widget:

image

image

I checked and if you use the regular `OpenLayersWidget` it does work, although the map isn't very detailed when you zoom in, so not sure if we can use another tile layer. Could you look into that @Floris272?
image

Floris272:
Using the default `OSMWidget` i got a better map:
image

it does seem to use lon,lat which is ESPG:3587 as Anna commented in the AuthorityAxisOrderOLWidget.

It can be switched using:
```python
class customOSMWidget(OSMWidget):
    map_srid = 4326
```

This is needed right?

alextreme:
Blocking this issue until #1908 is determined. Either we upgrade and have to keep CMIS intact or we remove CMIS as per 1908, however @joeribekker needs to make a decision

Floris272:
Test `test_besluitinformatieobjecten_cmis.BesluitInformatieObjectCMISAPITests.test_filter_by_informatieobject` fails because deep inside the django code for related objects in the class `MultiColSource` a warning was changed to an error.

image


(The cmis objects are not saved to the database and do not have a pk)

The test passed with 4.2 but did not actually work because the filter just returns all objects. this not caught because the filter is only tested with one created object.

Floris272:
During refinement it was decided to skip the test for now since CMIS support is currently not being updated and will be removed in the future. #1908

Floris272:
@stevenbal 
I changed UnsaltedMD5PasswordHasher to MD5PasswordHasher. not sure if this is the fastest option for the tests.

The AuthorityAxisOrderOLWidget now works the same as on the main branch:
- it expects coords in 4326 LAT LONG
- it swaps the the values (LONG LAT)
- the coords are transformed to 3587 for the map
- deserialize works the other way around: 3587 -> 4326 and swap

The confusing thing is that is in tests zaakgeometrie is created using LONG LAT order. in `CreateZaakTests.test_create_zaak` the following is used:

```
            "zaakgeometrie": {
                "type": "Point",
                "coordinates": [4.910649523925713, 52.37240093589432],
            },
```
creating an object with this will result in the widget not showing the correct location (because the coords are swapped)

some other context:
- In the API the headers Content-Crs and Accept-Crs are required and needs to be set to EPSG:4326 so the data should follow this.
- https://epsg.org/crs_4326/WGS-84.html 4326 defines the order as LAT LONG

src/openzaak/components/zaken/tests/admin/test_zaak_admin.py
```diff
@@ -297,3 +299,28 @@ def test_filter_on_betrokkene_rsin(self):
...
+        """
+        Tests that lat long coords (ESPG 4326) are swapped within the map.
+
+        The map uses ESPG 3857 so the coords need to transform
+        before they can be checked against the expected value.
+        """
+        lat = 52.38554000854492
+        long = 4.842977046966553
+
+        zaak = ZaakFactory.create(zaakgeometrie=Point(lat, long))
```
stevenbal, line 313: For my understanding: the GeoJSON spec expects the order `[long, lat]` via the API right? Could you replace the factory with an API call to create a Zaak that way, to make sure that this is shown correctly in the admin?

src/openzaak/conf/includes/base.py
```diff
@@ -350,9 +350,13 @@
...
-CSP_SCRIPT_SRC = CSP_SCRIPT_SRC + ["cdnjs.cloudflare.com"]
-CSP_IMG_SRC = CSP_IMG_SRC + ["cdnjs.cloudflare.com", "*.tile.openstreetmap.org"]
-CSP_STYLE_SRC = CSP_STYLE_SRC + ["cdnjs.cloudflare.com"]
+CSP_SCRIPT_SRC = CSP_SCRIPT_SRC + ["cdnjs.cloudflare.com", "cdn.jsdelivr.net"]
+CSP_IMG_SRC = CSP_IMG_SRC + ["cdnjs.cloudflare.com", "tile.openstreetmap.org"]
+CSP_STYLE_SRC = CSP_STYLE_SRC + ["cdnjs.cloudflare.com", "cdn.jsdelivr.net"]
+
+# TODO is there a better way to fix this?
+CSP_INCLUDE_NONCE_IN.remove("script-src")  # error with redoc.
+CSP_INCLUDE_NONCE_IN.remove("style-src")  # error with GISModelAdmin.
```
stevenbal, line 359: I think the comments for these need to be swapped, it seems that redoc breaks with the style-src nonce and GISModelAdmin with script-src nonce

stevenbal (changes requested):

src/openzaak/components/zaken/tests/admin/test_zaak_admin.py
Floris272: If i  use `[long, lat]` in a api call (or with the Point model) it will be shown incorrectly in the widget. (the same as on main). [see comment](https://github.com/open-zaak/open-zaak/pull/1931#issuecomment-2733259237)

If we want the api to also use long lat, the Content-Crs and Accept-Crs headers may need to be changed?

alextreme:
Removal of CMIS support has been communicated via the mailinglist and is to be completed in Q3 2025. How this affects this issue/PR is to be discussed

src/openzaak/components/zaken/tests/admin/test_zaak_admin.py
stevenbal: @Floris272 it's probably in that order because postgis seems to expect `long, lat` https://postgis.net/documentation/tips/lon-lat-or-lat-lon/

This also aligns with what GeoJSON expects: https://www.rfc-editor.org/rfc/rfc7946#section-3.1.1

So even though we do use the coordinate system EPSG:4326, since we want to adhere to the GeoJSON standard we should use long, lat as well I think

stevenbal (changes requested):
Comment about GeoJSON (the gift that keeps on giving :upside_down_face:) remaining

stevenbal (changes requested):
@Floris272 code looks good, though noticed that django-silk (by accident, because I've set `PROFILE=yes` in my .env) also needs to be upgraded to be compatible with django 5.1, could you upgrade django-silk to 5.3.2?

```
File "/open-zaak/5.1/lib/python3.11/site-packages/silk/models.py", line 8, in <module>
    from django.core.files.storage import get_storage_class
ImportError: cannot import name 'get_storage_class' from 'django.core.files.storage'
```

Floris272:
@stevenbal Done, can i merge the related setup-configuration & api framework prs and create releases?

Floris272:
@stevenbal because sites got removed in open-api-framwork 0.9.4 this needs to wait on  #1947.

stevenbal:
@Floris272 #1947 has been merged, could you rebase this PR on main?

stevenbal (approved):

PR #1947, opened 2025-03-19
:sparkles: Remove django.contrib.sites

danielmursa-dev:
Partially fixed https://github.com/maykinmedia/open-api-framework/issues/59

- [ ] Upgrade open-api-framework
- [ ] Upgrade notifications-api-common
- [ ] Check docs
- [x] Delete django_setup_configuration for site
- [x] Remove from INSTALLED_APPS 
- [x] Remove from Admin
- [x] Revert QRGeneratorView, tests and url


src/openzaak/conf/includes/base.py
```diff
@@ -171,6 +171,9 @@
     ),
     auto_display_default=False,
 )
+if not OPENZAAK_DOMAIN:
```
stevenbal, line 174: I think we should probably deprecate OPENZAAK_DOMAIN and rely on SITE_DOMAIN instead, add a fallback that's the other way around to ensure that the current usage of OPENZAAK_DOMAIN still works, and then in the next major release we can drop it

src/openzaak/components/autorisaties/tests/management/test_sync_zaaktype_autorisatie.py
```diff
@@ -13,44 +12,41 @@
...
-        with override_settings(ALLOWED_HOSTS=[domain]):
-            # Create 5 Autorisaties for non existing ZaakTypen
-            for i in range(5):
-                AutorisatieFactory.create(
-                    zaaktype=f"http://{domain}/catalogi/api/v1/zaaktypen/{str(uuid.uuid4())}",
-                )
+        # Create 5 Autorisaties for non existing ZaakTypen
+        for i in range(5):
+            AutorisatieFactory.create(
+                zaaktype=f"http://testserver/catalogi/api/v1/zaaktypen/{str(uuid.uuid4())}",
```
stevenbal, line 21: Could you replace this with a `reverse` to zaaktype-detail? Same applies for line 34

src/openzaak/utils/checks.py
```diff
@@ -89,19 +89,18 @@ def check_missing_init_files(app_configs, **kwargs):
...
-    if not settings.OPENZAAK_DOMAIN:
-        return []
-
     errors = []
-
-    try:
-        parsed = furl(netloc=settings.OPENZAAK_DOMAIN)
-    except ValueError:
-        invalid_netloc = True
+    if settings.OPENZAAK_DOMAIN:
```
stevenbal, line 93: Could you rewrite this check to check SITE_DOMAIN instead?

stevenbal (changes requested):
I think this might actually be a breaking change, because if this change is in and you deploy Open Zaak without setting SITE_DOMAIN (or OPENZAAK_DOMAIN, which isn't required) the view-config page will raise an error and `register_kanalen` will raise an error.

Might be good to discuss during standup

src/openzaak/components/autorisaties/tests/management/test_sync_zaaktype_autorisatie.py
danielmursa-dev: If I change the full path to the relative path, the tests fail because there is this check to see if zaaktype is not `is_local_url`:
https://github.com/open-zaak/open-zaak/blob/8eec4d26a57819bfe080d046367ea900716f5d9b/src/openzaak/components/autorisaties/management/commands/sync_autorisaties.py#L22

Should I also update the tests?

src/openzaak/conf/includes/base.py
```diff
@@ -171,6 +178,7 @@
     ),
     auto_display_default=False,
 )
+
```
stevenbal, line 181: Unrelated to this line, but `NOTIFICATIONS_API_GET_DOMAIN` needs to be set as well to make sure `src/manage.py register_kanalen` works, should have the same value as the commonground-api-common setting

src/openzaak/conf/includes/base.py
```diff
@@ -97,6 +97,12 @@
     + PLUGIN_INSTALLED_APPS
 )
 
+# `django.contrib.sites` added at the project level because it has been removed at the packages level.
+# This component is deprecated and should be completely removed.
+# To determine the project's domain, use the `SITE_DOMAIN` environment variable.
+if "django.contrib.sites" not in INSTALLED_APPS:
+    INSTALLED_APPS.append("django.contrib.sites")
```
stevenbal, line 104: Can you also add back the SITE_ID envvar here? This setting is required if sites is used https://github.com/maykinmedia/open-api-framework/pull/121/files#diff-8b7db4fa63ef0f1e64c48e8cf49d8beacf93587629abd42f1f6f9cd9a0aa8f06L32

stevenbal (changes requested):
I approved the other two PRs in the libraries. For Open zaak we're almost there, some small things left to make it work. Additionally could you add some tests to `test_view_config.py` to verify the behavior of the `/view-config` page? We want it to use the domain in this order SITE_DOMAIN > OPENZAAK_DOMAIN > django.contrib.sites

stevenbal:
failing docs build is addressed by https://github.com/open-zaak/open-zaak/pull/1966, you don't have to rebase

src/openzaak/components/autorisaties/tests/management/test_sync_zaaktype_autorisatie.py
stevenbal: I meant to keep the full path, but use reverse for the relative part:

```python
f"http://testserver{reverse('zaaktype-detail', kwargs={'uuid': uuid4()})}"
```

stevenbal (approved):

src/openzaak/utils/__init__.py
```diff
@@ -21,12 +21,15 @@ def get_openzaak_domain() -> str:
...
     if settings.OPENZAAK_DOMAIN:
         return settings.OPENZAAK_DOMAIN
 
     warnings.warn(
-        "Deriving the domain from the Site configuration will soon be deprecated, "
-        "please migrate to the OPENZAAK_DOMAIN setting.",
+        "Deriving the domain from the `OPENZAAK_DOMAIN` and `Sites` configuration will soon be deprecated, "
+        "please migrate to the SITE_DOMAIN setting.",
         PendingDeprecationWarning,
     )
```
stevenbal, line 34: This warning should probably be moved above the `if OPENZAAK_DOMAIN` statement

PR #1950, opened 2025-03-21
✅ Add basic performance test for zaken list

stevenbal:
Related issue: https://github.com/maykinmedia/open-api-framework/issues/114

I looked into codspeed, but I couldn't get it to work with their walltime instrument yet (https://github.com/maykinmedia/objects-api/pull/555), to at least get started with improving performance, I added similar tests to the ones I added in objects-api a few weeks back.

Not sure how consistent the results are in CI because we don't run it on dedicated runners, but it at least gives some indication of performance and can be used locally


danielmursa-dev (approved):
Looks good!

PR #1982, opened 2025-04-09
Feature/1930 kenmerken filter

### LINKED ISSUE #1930: Vraag; Hoe OF nummer of andere values querybaar registreren bij een zaak
### Thema / Theme

Zaken API

Open vraag;
Graag leggen we querybare values vast bij een zaak. Een voorbeeld hiervan is het OF- nummer.
We dachten zaak eigenschappen hiervoor te gebruiken maar de API lijkt niet te supporten om een query te doen als 'Geef mij alle zaken met OF nummer x'.
Wat is de beste manier om dit wel te doen?

_No response_

_No response_

Floris272:
Closes #1930 

**Changes**
- Adds `kenmerk__bron` filter
- Adds `kenmerk` filter that can filter on a bron-kenmerk combination. format `[<bron>:<kenmerk>]`

**Checklist**

- Commit hygiene

### DIFF
--- a/docs/api/experimental.rst
+++ b/docs/api/experimental.rst
@@ -73,6 +73,8 @@ Query parameters
     * ``rol__betrokkeneIdentificatie__vestiging__kvkNummer`` **(scheduled for deprecation in Open Zaak version 3.0)**
     * ``rol__machtiging``
     * ``rol__machtiging__loa``
+    * ``kenmerk__bron``
+    * ``kenmerk``
 
 
 Documenten API
--- a/docs/api/experimental.rst
+++ b/docs/api/experimental.rst
@@ -74,7 +74,7 @@ Query parameters
     * ``rol__machtiging``
     * ``rol__machtiging__loa``
     * ``kenmerk__bron``
-    * ``kenmerk``
+    * ``kenmerk`` A bron-kenmerk combination of a zaak. (format: ``<bron>:<kenmerk>``)
 
 
 Documenten API
--- a/src/openzaak/components/zaken/api/filters.py
+++ b/src/openzaak/components/zaken/api/filters.py
@@ -269,7 +269,7 @@ class ZaakFilter(FilterSetWithGroups):
         key_field_name="zaakkenmerk__bron",
         value_field_name="zaakkenmerk__kenmerk",
         help_text=mark_experimental(
-            "Een bron-kenmerk combinatie van de zaak. format: [<bron>:<kenmerk>]"
+            "Een bron-kenmerk combinatie van de zaak. format: <bron>:<kenmerk>"
         ),
     )
 
--- a/src/openzaak/components/zaken/api/filters.py
+++ b/src/openzaak/components/zaken/api/filters.py
@@ -15,6 +15,7 @@ from vng_api_common.utils import get_field_attribute, get_help_text
 from openzaak.components.zaken.api.serializers.zaken import ZaakSerializer
 from openzaak.utils.filters import (
     ExpandFilter,
+    KeyValueFilter,
     MaximaleVertrouwelijkheidaanduidingFilter,
 )
 from openzaak.utils.filterset import FilterGroup, FilterSet, FilterSetWithGroups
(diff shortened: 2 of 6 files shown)


docs/api/experimental.rst
```diff
@@ -73,6 +73,8 @@ Query parameters
     * ``rol__betrokkeneIdentificatie__vestiging__kvkNummer`` **(scheduled for deprecation in Open Zaak version 3.0)**
     * ``rol__machtiging``
     * ``rol__machtiging__loa``
+    * ``kenmerk__bron``
+    * ``kenmerk``
```
stevenbal, line 77: let's also mention the format for this query parameter here

src/openzaak/components/zaken/api/filters.py
```diff
@@ -258,6 +259,20 @@ class ZaakFilter(FilterSetWithGroups):
...
+    kenmerk__bron = filters.CharFilter(
+        field_name="zaakkenmerk__bron",
+        help_text=mark_experimental(get_help_text("zaken.ZaakKenmerk", "bron")),
+    )
+
+    kenmerk = KeyValueFilter(
+        key_field_name="zaakkenmerk__bron",
+        value_field_name="zaakkenmerk__kenmerk",
+        help_text=mark_experimental(
+            "Een bron-kenmerk combinatie van de zaak. format: [<bron>:<kenmerk>]"
```
stevenbal, line 272: Do we need the square brackets for these values? I think `[]` are reserved characters in URLs, so they might have to be encoded for it to work.

Locally this seems to work without square brackets as well actually, so can't we just drop that from the format?

```suggestion
            "Een bron-kenmerk combinatie van de zaak. format: `<bron>:<kenmerk>`"
```

stevenbal (changes requested):

src/openzaak/components/zaken/api/filters.py
Floris272: The validator didn't work, i removed the brackets around the k:v but should the brackets still be disallowed in the key & value themselves? or maybe only allow `[a-z][A-Z]`, `_` & `-`?

Floris272: add didgets for the value

Floris272: `[a-zA-Z0-9_-]+` but then the db fields should also have this as validation.

src/openzaak/components/zaken/tests/test_zaak_filter.py
```diff
@@ -304,3 +304,77 @@ def test_filter_zaaktype_omschrijving(self):
...
+        )
+
+    def test_filter_kenmerken(self):
+        zaak = ZaakFactory.create()
+        ZaakKenmerkFactory(bron="BAG-id", kenmerk="123", zaak=zaak)
+        ZaakKenmerkFactory(bron="GAB-id", kenmerk="456", zaak=zaak)
+
+        ZaakFactory.create()
+
+        with self.subTest("existing combination"):
```
stevenbal, line 353: could you also add a test to verify that this works if you explicitly URLencode the query param? E.g. `?kenmerk=foo%3Abar` should also work, `:` should technically be encoded because it's a reserved character in URLs

src/openzaak/utils/filters.py
```diff
@@ -49,3 +51,36 @@ def __init__(self, *args, **kwargs):
...
     def filter(self, qs, value):
         return qs
+
+
+class KeyValueFilter(filters.CharFilter):
+
+    def __init__(self, key_field_name, value_field_name, *args, **kwargs):
+        kwargs.setdefault(
+            "validators", [RegexValidator(r"^([^:\[\]]+):([^:\[\]]+)$")]
+        )  # key:value where the key and value cannot contain `[`, `]` or `:`
```
stevenbal, line 61: Is it possible to specify this as a class variable to make sure it's always applied? Because if we reuse this filter and pass extra validators, this one won't be applied I think

stevenbal (changes requested):

src/openzaak/components/zaken/tests/test_zaak_filter.py
Floris272: added a test, but it seems that the other tests do encode. `response.request["QUERY_STRING"]` has `%3A`

stevenbal (approved):

PR #1985, opened 2025-04-10
Feature/django cachalot

### LINKED ISSUE #1986: Performance: investigate if caching can be applied to results of database queries
### Thema / Theme

Zaken API

This was previously attempted in https://github.com/open-zaak/open-zaak/pull/1377, but not being used due to unexpected interactions with `ZaakIdentificatie` tabel.

I think it would be worth to invest more time into this to see if it can work

_No response_

_No response_

stevenbal:
Closes #1986

TODO:
- [ ] check if generate_identificatie works for other resources
- [ ] apply more sparingly? on specific viewsets, etc
- [ ] cache specific queries with cacheops of which the result wont change often (authorization/catalogi related stuff)

issues:
- applying it globally seems to cause issues (cache not being invalidated?)
- redis issues because keys arent removed? https://django-cachalot.readthedocs.io/en/latest/limits.html#redis

**Changes**

[Describe the changes here]

**Checklist**

- Commit hygiene

### DIFF
--- a/requirements/base.in
+++ b/requirements/base.in
@@ -10,6 +10,7 @@ requests-cache
 self-certifi
 
 # Framework libraries
+django-cachalot
 django-db-logger
 django-extra-views
 django-loose-fk
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -96,6 +96,7 @@ django==5.1.7
     #   django-admin-index
     #   django-appconf
     #   django-axes
+    #   django-cachalot
     #   django-celery-beat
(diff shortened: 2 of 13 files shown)


stevenbal:
Putting this on hold for now, cachalot seems to be all or nothing in terms of applying caching to specific tables, while I'd rather cache specific queries only. If we start caching all queries on `zaken_zaak`, I don't think that would be too effective, since changing a single parameter like `ordering` or another filter param will result in a newly created cache entry.

PR #1988, opened 2025-04-11
Feature/1920 detect past zaak suspension

### LINKED ISSUE #1920: Detect if a zaak was suspended in the past
### Thema / Theme

Zaken API

It would be beneficial if OpenZaak provides a flag that specifies that a zaak was already suspended in the past. This can be boolean or the date on which it was suspended.

At the moment listing zaaken for never suspended zaak we get:
```
      "opschorting": {
        "indicatie": false,
        "reden": ""
      },
```

For a suspended in the past zaak the API returns:
```
"opschorting": {
  "indicatie": false,
  "reden": "Aanvullende informatie geleverd"
}
```

The only good way to know that a zaak was suspended in the past is to parse audit log / history of a zaak. That requires a lot of custom/client code.

Having a dedicated boolean flag or a date makes processing really easy and unified across clients.

[ZAC](https://github.com/infonl/dimpact-zaakafhandelcomponent) wants to implement the following logic: if a zaak was suspended once it can no longer be suspended

For this we now use the suspension reason field in OpenZaak, but we think that's not really a good way to implement this feature.

Floris272:
Closes #1920

**Changes**
- Adds `eerdere_opschorting` to `Zaak`
- Adds `GegevensGroepTypeWithReadOnlyFields` that adds a read_only parameter to specify fields that should keep their existing value. (the original GegevensGroepType class sets fields to their default value if they are not within the value parameter in `__set__`)

**Checklist**

- Commit hygiene


src/openzaak/components/zaken/tests/admin/test_zaak_admin.py
```diff
@@ -343,3 +343,19 @@ def test_zaak_geometrie_change_from_admin(self):
 
         self.assertAlmostEqual(zaak.zaakgeometrie.x, long)
         self.assertAlmostEqual(zaak.zaakgeometrie.y, lat)
+
+    def test_zaak_opschorting_indicatie_sets_eerdere_opschorting(self):
+        zaak = ZaakFactory.create()
```
stevenbal, line 348: Might be good to explicitly initialize these attributes as `False`, as a sanity check
```suggestion
        zaak = ZaakFactory.create(opschorting_indicatie=False, opschorting_eerdere_opschorting=False)
```

src/openzaak/components/zaken/tests/test_zaak_opschorting.py
```diff
@@ -0,0 +1,76 @@
...
+from .factories import ZaakFactory
+from .utils import ZAAK_WRITE_KWARGS
+
+
+class ZaakOpschortingTests(JWTAuthMixin, APITestCase):
+
+    heeft_alle_autorisaties = True
+
+    def test_eerdere_opschorting_field_cannot_be_changed(self):
+        zaak = ZaakFactory.create()
```
stevenbal, line 19:
```suggestion
        zaak = ZaakFactory.create(opschorting_indicatie=False, opschorting_eerdere_opschorting=False)
```

src/openzaak/utils/descriptors.py
```diff
@@ -0,0 +1,33 @@
...
+        assert (
+            read_only_fields_known
+        ), "The fields in 'read_only' must be a subset of the mapping keys"
+
+    def __set__(self, obj, value: Optional[dict]):
+        if not value:
+            value = {}
+
+        for key in self.read_only:
+            value[key] = getattr(obj, self.mapping[key].name)
```
stevenbal, line 31: What happens if you create a Zaak with `opschorting`? Not a very realistic scenario I know, but I'm wondering if `obj` is something other than `None` in that case?

stevenbal (changes requested):

src/openzaak/utils/descriptors.py
Floris272: Added a test, it seems obj is always filled, looking at the `NestedGegevensMixin` zaak is already saved before the gegevensgroep is set.

stevenbal (approved):

PR #2001, opened 2025-04-18
:bug: [#2011] Show appropriate betalingsindicatieWeergave via serializer

### LINKED ISSUE #2011: `Zaak.betalingsindicatieWeergave` does not have an underlying database field and is always empty

1.20.0 / latest

I noticed that the attribute `Zaak.betalingsindicatieWeergave` is read only, refers to `get_betalingsindicatie_display` as its source, even though this method is not defined anywhere and there also isn't any field on the `Zaak` model to store this information. I checked with the standard and reference implementation and those have the same issue

- https://github.com/VNG-Realisatie/zaken-api/blob/9be7e14b0f08adc7932e3e4cad6bec57c5cbd6b3/src/zrc/api/serializers/core.py#L223-L227
- https://github.com/VNG-Realisatie/zaken-api/blob/9be7e14b0f08adc7932e3e4cad6bec57c5cbd6b3/src/zrc/datamodel/models/core.py#L185

_No response_

I think this field should function properly, from what I gather from the help text, displaying the label of the `BetalingsIndicatie` choices would make sense

stevenbal:
Fixes #2011

this field was read only and was never populated, even if betalingsindicatie was set. Now it has been changed to show the label that corresponds with the betalingsindicatie value that the Zaak has

Baseline: https://github.com/open-zaak/open-zaak/actions/runs/14726552958/job/41330465681#step:5:62
After changes: https://github.com/open-zaak/open-zaak/actions/runs/14726565385/job/41330508336?pr=2001#step:5:62

**Changes**

* Show appropriate betalingsindicatieWeergave via serializer

**Checklist**

- Commit hygiene

### DIFF
--- a/src/openzaak/components/documenten/tests/models/test_last_version.py
+++ b/src/openzaak/components/documenten/tests/models/test_last_version.py
@@ -2,12 +2,15 @@
 # Copyright (C) 2019 - 2020 Dimpact
 from django.test import TestCase, override_settings
 
+from privates.test import temp_private_root
+
 from ..factories import (
     EnkelvoudigInformatieObjectCanonicalFactory,
     EnkelvoudigInformatieObjectFactory,
 )
 
 
+@temp_private_root()
 @override_settings(CMIS_ENABLED=False)
 class LastVersionTests(TestCase):
     def test_canonical_last_version(self):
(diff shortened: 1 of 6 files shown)


Floris272 (approved):

stevenbal:
@danielmursa-dev @Floris272 ended up fixing the issue related to `betalingsindicatieWeergave` anyway, since it was a small fix

danielmursa-dev (approved):

Floris272 (approved):

PR #2029, opened 2025-05-15
:sparkles: [#2018] Implement document number reservation for EnkelvoudigInformatieObject

### LINKED ISSUE #2018: Documentnummer reserveren t.b.v. gekoppelde TSA
### Thema / Theme

Documenten API

T.b.v. de koppeling met iBurgerzaken. iBurgerzaken kan hiermee documenten creëren als (de koppeling met) OpenZaak niet beschikbaar is.

_No response_

_No response_

OlhaZahoruiko:
Closes #2018

- A new ReservedDocument model with identificatie, bronorganisatie, and date_created
- An update to generate_document_identificatie() to avoid already reserved IDs
- Validation logic ensuring that if a document is created with a specific identificatie, it must match an existing reservation
- Automatic removal of the reservation upon document creation

Experimental features/changes

- [x] Any experimental features added in this PR are documented in the docs/api/experimental.rst page


src/openzaak/components/documenten/models.py
```diff
@@ -466,6 +468,20 @@ def locked(self, value: bool) -> None:
         self._locked = value
 
     def save(self, *args, **kwargs) -> None:
+        is_new_instance = self.pk is None
+        if is_new_instance and hasattr(self, "bronorganisatie"):
```
stevenbal, line 472: I think `hasattr(self, "bronorganisatie")` will always evaluate to true, so I think that condition can be removed

src/openzaak/components/documenten/models.py
```diff
@@ -466,6 +468,20 @@ def locked(self, value: bool) -> None:
         self._locked = value
 
     def save(self, *args, **kwargs) -> None:
+        is_new_instance = self.pk is None
```
stevenbal, line 471: `EnkelvoudigInformatieObject` inherits from `InformatieObject` and since the identification is currently generated in `InformatieObject.save`, let's move the changes made in `EnkelvoudigInformatieObject.save` to `InformatieObject.save`.

Let's also add the decorator `@transaction.atomic` to make sure the deletion of ReservedDocuments is rolled back if anything fails while saving the `EnkelvoudigInformatieObject`

```
from django.db import transaction

@transaction.atomic
def save(...):
    ...
```

src/openzaak/components/documenten/tests/models/test_human_readable_identification.py
```diff
@@ -10,18 +10,18 @@
 class EIOTests(TestCase):
     def test_default_human_readable(self):
         eio = EnkelvoudigInformatieObjectFactory.create(
-            identificatie="", creatiedatum=date(2019, 7, 1)
+            identificatie="", creatiedatum=date(2025, 7, 1)
```
stevenbal, line 13: Are these changes needed?

src/openzaak/components/documenten/models.py
```diff
@@ -1292,3 +1308,34 @@ def get_informatieobject(self, permission_main_object=None):
...
+        "heeft vastgelegd.",
+        db_index=True,
+    )
+    date_created = models.DateTimeField(
+        auto_now_add=True,
+        help_text=_("De datum en tijd waarop het document werd gereserveerd."),
+    )
+
+    class Meta:
+        unique_together = ("bronorganisatie", "identificatie")
```
stevenbal, line 1336: `unique_together` might be deprecated in a future release of Django, so it's generally recommended to use `UniqueConstraint` (https://docs.djangoproject.com/en/5.2/ref/models/options/#django.db.models.Options.unique_together)
```suggestion
        constraints = [
            models.UniqueConstraint(
                fields=["bronorganisatie", "identificatie"],
                name="bronorganisatie_identificatie_unique_together",
            ),
        ]
```

src/openzaak/components/documenten/api/utils.py
```diff
@@ -46,3 +48,36 @@ def check_path(url, resource):
         return False
 
     return True
+
+
+def generate_document_identificatie(
+    bronorganisatie: str, document_model: models.Model, date_value: date
```
stevenbal, line 54: Is it necessary to pass the `document_model` as an argument?

src/openzaak/components/documenten/api/serializers.py
```diff
@@ -1112,3 +1114,47 @@ class Meta:
             "url": {"lookup_field": "uuid", "read_only": True},
         }
         validators = [VerzendingAddressValidator()]
+
+
+class ReservedDocumentSerializer(serializers.ModelSerializer):
```
stevenbal, line 1119: To make sure this can also be used via the API, we'll need to have a view that uses this serializer. The reserveer zaken endpoint has one as well: https://github.com/open-zaak/open-zaak/blob/main/src/openzaak/components/zaken/api/viewsets.py#L1431 which is also registered in `urls.py` to make sure it's available at a path.

If possible I'd like to see if we can use a function based view (https://www.django-rest-framework.org/api-guide/views/#function-based-views) for this instead of a class based view (like the reserveer zaaknummer endpoint has now), because it's just a single method (POST), but let's first implement it the same as reserveer zaaknummer and check if we can refactor this later

src/openzaak/components/documenten/tests/test_reserved_document.py
```diff
@@ -0,0 +1,131 @@
...
+from openzaak.components.documenten.models import (
+    EnkelvoudigInformatieObject,
+    ReservedDocument,
+)
+from openzaak.components.documenten.tests.factories import (
+    EnkelvoudigInformatieObjectFactory,
+)
+
+
+class ReservedDocumentTests(TestCase):
```
stevenbal, line 20: When the viewset and urlpattern for this endpoint have been added, these tests can be changed to make actual API calls to the endpoint, instead of calling the serializer methods directly. You can see an example of these kind of tests at `src/openzaak/components/zaken/tests/test_zaken.py` for example.

```suggestion
class ReservedDocumentTests(APITestCase):
```

stevenbal (changes requested):
Going in the right direction :+1: 

@OlhaZahoruiko if you want the PR to be reviewed, could you make sure it's no longer in draft? 
afbeelding

Also I'm not sure if you modified the PR description, but PRs should by default include a checklist, to remind the PR creator and reviewer of some things:

afbeelding

In this case the first two points apply, because this change is experimental (it's not part of the Zaken API standard), so once you add the viewsets it should be marked as such (https://github.com/open-zaak/open-zaak/blob/main/src/openzaak/components/zaken/api/viewsets.py#L1435)

src/openzaak/components/documenten/tests/models/test_human_readable_identification.py
OlhaZahoruiko: No, already removed

src/openzaak/components/documenten/api/viewsets.py
```diff
@@ -981,3 +983,27 @@ def destroy(self, request, *args, **kwargs):
...
         return super().destroy(request, *args, **kwargs)
+
+
+class ReservedDocumentViewSet(mixins.CreateModelMixin, viewsets.GenericViewSet):
+    queryset = ReservedDocument.objects.all()
+    serializer_class = ReservedDocumentSerializer
+
+    @extend_schema(
+        summary="Reserveer een documentnummer",
+        description=(
```
stevenbal, line 994: Since this endpoint is not part of the Documenten API standard, it has to be marked as "experimental" in the API spec and listen in `docs/api/experimental.rst` as a experimental feature. Could you add that?
```suggestion
        from openzaak.utils.help_text import mark_experimental
        ...
        description=mark_experimental(
```

src/openzaak/components/documenten/tests/test_reserved_document.py
```diff
@@ -0,0 +1,198 @@
...
+            bronorganisatie=self.bronorganisatie,
+        )
+
+    def test_reserved_document_str_representation(self):
+        reserved_document = ReservedDocument.objects.create(
+            identificatie="DOCUMENT-2025-1234567890",
+            bronorganisatie="812345678",
+        )
+        expected_str = "812345678 - DOCUMENT-2025-1234567890"
+        self.assertEqual(str(reserved_document), expected_str)
```
stevenbal, line 43: I guess it's fine to have this, but generally I'd say we wouldn't need a test for this

src/openzaak/components/documenten/tests/test_reserved_document.py
```diff
@@ -0,0 +1,198 @@
...
+            "bronorganisatie": self.bronorganisatie,
+        }
+
+        response = self.client.post(self.url, data, format="json")
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+
+        identificatie = response.data.get("identificatie")
+        self.assertIsNotNone(identificatie)
+        self.assertTrue(identificatie.startswith("DOCUMENT-" + str(date.today().year)))
+        self.assertEqual(len(identificatie.split("-")[-1]), 10)
```
stevenbal, line 136: If you use `from freezegun import freeze_time`, you can make this more explicit. You can put a decorator on the entire testcase class like `@freeze_time("2025-01-01T12:00:00")` and then add more readable assertions. I'd prefer to see the exact expected identificatie hardcoded here, e.g. `"DOCUMENT-2025-000000000X"`

src/openzaak/components/documenten/tests/test_reserved_document.py
```diff
@@ -0,0 +1,198 @@
...
+
+        with self.assertRaises(ReservedDocument.DoesNotExist):
+            ReservedDocument.objects.get(
+                identificatie=self.reserved_identificatie,
+                bronorganisatie=self.bronorganisatie,
+            )
+
+        self.assertEqual(eio.identificatie, self.reserved_identificatie)
+
+    def test_create_enkelvoudig_info_object_with_non_reserved_identificatie(self):
```
stevenbal, line 73: I don't think this test making the right assertions, it should only be possible to reserve an identificatie if it has not been reserved yet

src/openzaak/components/documenten/tests/test_reserved_document.py
```diff
@@ -0,0 +1,198 @@
...
+
+from openzaak.components.documenten.api.serializers import ReservedDocumentSerializer
+from openzaak.components.documenten.api.utils import generate_document_identificatie
+from openzaak.components.documenten.models import ReservedDocument
+from openzaak.components.documenten.tests.factories import (
+    EnkelvoudigInformatieObjectFactory,
+)
+
+
+class ReservedDocumentTests(APITestCase):
```
stevenbal, line 18: It's going in the right direction now that there are some tests that use the API endpoint, but there are still some tests that test implementation details (like directly using `ReservedDocument.objects.create` or serializer methods). 

I think we need tests for the following, all of which I think should use the API to test that this behavior applies:
- doing a successful POST on /reserveer-document returns an `identificatie` that is not yet used by other existing reserveddocuments and informatieobjecten
- doing a successful POST on /enkelvoudiginformatieobjecten without passing `identificatie` should generate an `identificatie` that excludes existing reserveddocuments and informatieobjecten
- doing a successful POST on /enkelvoudiginformatieobjecten with a reserved identificatie and bronorganisatie combination should delete the reserveddocument if it existed (it should work both when it was and wasn't reserved)

stevenbal (changes requested):

src/openzaak/components/documenten/api/utils.py
```diff
@@ -46,3 +47,33 @@ def check_path(url, resource):
...
+    model_name = "DOCUMENT"
+
+    year = date_value.year
+    prefix = f"{model_name}-{year}"
+    pattern = prefix + r"-\d{10}"
+
+    issued_ids = ReservedDocument._default_manager.filter(
+        identificatie__startswith=prefix,
+        identificatie__regex=pattern,
+    ).values_list("identificatie", flat=True)
```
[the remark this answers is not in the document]
OlhaZahoruiko: > Shouldn't this use the `InformatieObject` model to get the issued ids instead? Otherwise it's querying on the same table twice I think

Yes, you're right

src/openzaak/components/documenten/tests/test_reserved_document.py
OlhaZahoruiko: But how can I cover this with a test?

src/openzaak/components/documenten/tests/test_reserved_document.py
```diff
@@ -0,0 +1,287 @@
...
+        serializer = ReservedDocumentSerializer(
+            data={
+                "identificatie": not_reserved_id,
+                "bronorganisatie": self.bronorganisatie,
+            }
+        )
+        serializer.initial_data = {"bronorganisatie": self.bronorganisatie}
+
+        validated = serializer.validate_identificatie(not_reserved_id)
+        self.assertEqual(validated, not_reserved_id)
```
stevenbal, line 108: Can be removed if identificatie is made read only
```suggestion
```

src/openzaak/components/documenten/tests/test_reserved_document.py
```diff
@@ -0,0 +1,287 @@
...
+)
+from openzaak.components.documenten.api.scopes import SCOPE_DOCUMENTEN_AANMAKEN
+from openzaak.components.documenten.api.serializers import ReservedDocumentSerializer
+from openzaak.components.documenten.api.utils import generate_document_identificatie
+from openzaak.components.documenten.models import ReservedDocument
+from openzaak.components.documenten.tests.factories import (
+    EnkelvoudigInformatieObjectFactory,
+)
+from openzaak.tests.utils import JWTAuthMixin
+
```
stevenbal, line 22: Could you tag all testcases you've added with `@tag("gh-2018")`?

src/openzaak/components/documenten/tests/test_reserved_document.py
```diff
@@ -0,0 +1,287 @@
...
+        ReservedDocument.objects.create(
+            identificatie=self.another_reserved_identificatie,
+            bronorganisatie=self.bronorganisatie,
+        )
+
+        Autorisatie.objects.create(
+            applicatie=self.applicatie,
+            component=ComponentTypes.drc,
+            scopes=["documenten.aanmaken"],
+        )
```
stevenbal, line 51: If we have a separate testcase for the permission part, we can use `heeft_alle_autorisaties=True` at the class level instead

```suggestion
```

src/openzaak/components/documenten/tests/test_reserved_document.py
```diff
@@ -0,0 +1,287 @@
...
+                "identificatie": self.reserved_identificatie,
+            }
+            response = self.client.post(self.url, data, format="json")
+            self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+
+            with self.assertRaises(ReservedDocument.DoesNotExist):
+                ReservedDocument.objects.get(
+                    identificatie=self.reserved_identificatie,
+                    bronorganisatie=self.bronorganisatie,
+                )
```
stevenbal, line 265:
```suggestion
            self.assertFalse(ReservedDocument.objects.filter(...).exists()
```

stevenbal (changes requested):

PR #2034, opened 2025-05-20
:art: [maykinmedia/open-api-framework#133] Refactor code with ruff

danielmursa-dev:
Fixes #[maykinmedia/open-api-framework#133]

Workflow: https://github.com/maykinmedia/open-api-workflows/pull/23

- [x] Install `ruff` and uninstall `black`, `isort`, `flake8`
- [ ] Update `code-quality` workflow version
- [x] Format code after ruff checks
- [x] Update README, index.rst
- [x] Set `pre-commit` and update docs
- [x] Add `.git-blame-ignore-revs`


bin/check_spdx.py
```diff
@@ -62,11 +60,18 @@ def main(src):
         sys.exit(1)
 
 
-def get_files_in_dir(path: Path, root: Path, gitignore: PathSpec) -> List[Path]:
-    assert root.is_absolute(), f"INTERNAL ERROR: `root` must be absolute but is {root}"
+def get_gitignore_files():
+    command = ["git", "ls-files", "--others", "--ignored", "--exclude-standard"]
+    result = subprocess.run(command, capture_output=True, text=True, check=True)
+    return [line for line in result.stdout.strip().split("\n") if line]
```
stevenbal, line 66: Have you verified that the script works? I'm trying to run it locally but it takes really long to complete, possibly because it's loading all files that are ignored by git into a list.

I think you could call `get_gitignore` once (outside of `get_files_in_dir`) to read the glob patterns into a list and then use something like `pathlib.Path.match` to check if a path matches

stevenbal (changes requested):
Looks good overall :+1:

bin/check_spdx.py
```diff
@@ -62,11 +62,18 @@ def main(src):
...
+
+
+def get_files_in_dir(path: Path, root: Path, gitignore_files: List[str]) -> List[Path]:
```
stevenbal, line 71: The return type is incorrect I think, it should be a Generator
afbeelding

stevenbal (approved):
Typing error, but good otherwise

PR #2035, opened 2025-05-20
:zap: [#2025] add component_type check for unnecessary db query's

### LINKED ISSUE #2025: `EnabledMiddleware` performs database query, even for non API requests

1.21.0 / latest

Noticed this on k8s because of liveness probes (requests on `/`), the `EnabledMiddleware` performs a database query on `InternalService`, even though this is not an API request. This causes unnecessary overhead, so this middleware should only perform the query if the request has a `component_type`

Image

_No response_

_No response_

TimdeBeer1:
Closes #2025 

**Changes**

Added check in EnableMiddleware to to return None before InternalService db query if there is no component_type.

Added tests.

**Checklist**

- Commit hygiene

### DIFF
--- a/src/openzaak/tests/test_enabled_middleware.py
+++ b/src/openzaak/tests/test_enabled_middleware.py
@@ -1,0 +1,48 @@
+# SPDX-License-Identifier: EUPL-1.2
+# Copyright (C) 2025 Dimpact
+
+from django.http import HttpResponseNotFound
+from django.test import RequestFactory, TestCase
+
+from vng_api_common.constants import ComponentTypes
+
+from openzaak.config.models import InternalService
+from openzaak.utils.middleware import EnabledMiddleware
+
+
+class EnabledMiddlewareDatabaseQueryTest(TestCase):
+    def setUp(self):
+        self.middleware = EnabledMiddleware(lambda r: None)
+        self.factory = RequestFactory()
+
+    def test_no_db_query_if_no_component_type(self):
+        request = self.factory.get("/")
+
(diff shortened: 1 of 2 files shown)


src/openzaak/tests/test_enabled_middleware.py
```diff
@@ -0,0 +1,35 @@
...
+# Copyright (C) 2025 Dimpact
+from unittest.mock import patch
+
+from django.http import HttpResponseNotFound
+from django.test import RequestFactory, SimpleTestCase
+
+from openzaak.utils.middleware import EnabledMiddleware
+
+
+class EnabledMiddlewareDatabaseQueryTest(SimpleTestCase):
```
stevenbal, line 11: Generally I'd prefer to mock as little as possible, so instead of using mocks here, let's use the database
```suggestion
class EnabledMiddlewareDatabaseQueryTest(TestCase):
```

You can then do something like:

```python
    def test_no_db_query_if_no_component_type(self):
        request = self.factory.get("/")

        with self.assertNumQueries(0):
            result = self.middleware.process_view(request, None, None, None)

        self.assertIsNone(result)
```

And for the 404 test you'd have to ensure the `InternalService` you're testing is disabled. You could also add a third test to verify that `InternalService.enabled=True` works as expected, that way all cases are covered.

stevenbal (changes requested):

stevenbal (changes requested):

stevenbal (approved):

PR #2044, opened 2025-05-27
:sparkles: [#1852] add connection pooling

### LINKED ISSUE #1852: Open Zaak Django 5.1 upgrade + provide image with native connection pooling support
### Thema / Theme

Other

Taiga DH 607 option 4

The request is that we attempt to upgrade Open Zaak to Django 5.1 in order to check with DH if the native connection pooling (requires psycopg[pool]) speeds up the installations where Postgresql isn't on the same machine/cluster:

https://docs.djangoproject.com/en/5.1/ref/databases/#connection-pool

After we have a build DH can perform a loadtest using this image, which should give certainty on if this significantly improves performance or not.

Up to Team Bron if this should lead to a regular Open Zaak release or not, as we usually don't make use of non-LTS versions of Django

_No response_

_No response_

TimdeBeer1:
Closes #1852

**Changes**

Add connection pooling

**Checklist**

- Commit hygiene

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,26 +6,25 @@ on:
     branches:
       - main
     tags:
-      - '*'
+      - "*"
   pull_request:
   workflow_dispatch:
 
 env:
   IMAGE_NAME: openzaak/open-zaak
   DJANGO_SETTINGS_MODULE: openzaak.conf.ci
-  DB_PASSWORD: ''
+  DB_PASSWORD: ""
   DB_USER: postgres
   ALLOWED_HOSTS: openzaak.nl
-  COMPOSE_DOCKER_CLI_BUILD: '1'
-  DOCKER_BUILDKIT: '1'
+  COMPOSE_DOCKER_CLI_BUILD: "1"
+  DOCKER_BUILDKIT: "1"
 
 jobs:
   # determine changed files to decide if certain jobs can be skipped or not
   changed-files:
-    runs-on: ubuntu-latest  # windows-latest | macos-latest
+    runs-on: ubuntu-latest # windows-latest | macos-latest
     name: Determine changed files
     steps:
-
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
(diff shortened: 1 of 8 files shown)


.github/workflows/ci.yml
```diff
@@ -161,14 +167,19 @@ jobs:
             --verbosity 2
           coverage combine
           mv .coverage .coverage-regular
+        env:
+          DJANGO_SETTINGS_MODULE: openzaak.conf.ci
+          SECRET_KEY: dummy
+          DB_USER: postgres
+          DB_PASSWORD: ''
+          DB_POOL_ENABLED: ${{ matrix.use_pooling }}
```
Floris272, line 175: does this work with only adding `DB_POOL_ENABLED` ?

.github/workflows/ci.yml
```diff
@@ -307,22 +318,20 @@ jobs:
...
-          name: coverage-cmis-tests
           path: coverage-data
+
       - name: Combine reports
         run: |
           pip install $(grep codecov= requirements/ci.txt)
           ls -la coverage-data
-          coverage combine coverage-data/.coverage-regular coverage-data/.coverage-cmis
+          coverage_files=$(find coverage-data -name '.coverage-regular' -o -name '.coverage-cmis')
+          coverage combine $coverage_files
```
Floris272, line 332: maybe this can be done with `coverage-data/*` ?

Floris272 (changes requested):

.github/workflows/ci.yml
```diff
@@ -161,14 +167,19 @@ jobs:
...
       - name: Upload Coverage Report
-        # TODO find a better way to upload this only once
-        if: ${{ matrix.postgres == '17' && matrix.postgis == '3.5' }}
         uses: actions/upload-artifact@v4
         with:
-          name: coverage-regular-tests
-          path: ${{ github.workspace }}/.coverage-regular
+          name: coverage-regular-tests-pg${{ matrix.postgres }}-postgis${{ matrix.postgis }}${{ matrix.use_pooling && '-pooling' || '' }}
+          path:  ${{ github.workspace }}/.coverage-regular
           include-hidden-files: true
```
Floris272, line 184: I think the if statement should be kept op only upload the coverage of pg 17 with gis 3.5

To keep the naming consistent i think coverage-regular-tests & coverage-db-pooling-tests is better. The path should also probably change based on `matrix.use_pooling`.

.github/workflows/ci.yml
TimdeBeer1: not exactly, fixed it with .coverage-* for the file names

PR #2096, opened 2025-07-04
Feature/oaf 157 datadump script

Floris272:
Closes https://github.com/maykinmedia/open-api-framework/issues/157

**Changes**

[Describe the changes here]

**Checklist**

- Commit hygiene


bin/dump_data.sh
```diff
@@ -0,0 +1,41 @@
...
+
+>&2 echo "exporting: ${APPS[*]}"
+
+# create -t flags for each app
+INCLUDES=()
+for app in "${APPS[@]}"; do
+    INCLUDES+=("-t" "${app}_*")
+done
+
+pg_dump "${INCLUDES[@]}" -f "$DUMP_FILE"
```
stevenbal, line 39: Not using `--data-only` here does mean that you can't import this data into another instance that already has its tables set up (migrations have run), because it will also export all the table info, indexes etc.

@joeribekker I'm not 100% if the use case of this script is to dump data and load it somewhere else, but if that is the case, I think it should be dumping data only.

bin/dump_data.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+
+# Dump component data
+# Run this script from the root of the repository
+# dump_data.sh for all apps or dump_data.sh zaken documenten ... to specify specific apps
```
stevenbal, line 5: Could you add a docs page that explains this command and refers to this script in github? I think it would also be good to mention how this dumped data can be loaded

.github/workflows/bin-check.yml
```diff
@@ -0,0 +1,57 @@
...
+          DB_PASSWORD: ""
+          DB_USER: postgres
+          DB_NAME: openzaak
+          DB_HOST: localhost
+          SECRET_KEY: secret
+          DJANGO_SETTINGS_MODULE: openzaak.conf.ci
+          ALLOWED_HOSTS: localhost,127.0.0.1
+
+      - name: validate dump
+        run: |
```
stevenbal, line 52: You could also try loading it into another database with `psql`, to verify that it works

stevenbal (changes requested):

.github/workflows/bin-check.yml
Floris272: added step

bin/dump_data.sh
Floris272: Changed it to do the following:

By default it will create a complete schema dump and a separate data dump for all apps, The datadump will disable triggers because zgw_consumers.Service objects won't exist.

with `--data-only`, `--schema-only` only one sql dump will be created

using `--combined` will append the data dump to the schema dump

docs/manual/scripts.rst
```diff
@@ -0,0 +1,35 @@
+.. _scripts:
+
+Scripts
+=======
+
+Dump data
+---------
+
+Met het script `dump_data.sh` kan de data van alle componenten (zaken, documenten...) worden geëxporteerd naar een sql bestand.
```
stevenbal, line 9: Otherwise it's rendered as *italic* in rst
```suggestion
Met het script ``dump_data.sh`` kan de data van alle componenten (zaken, documenten...) worden geëxporteerd naar een sql bestand.
```

stevenbal, line 9:
```suggestion
Met het script `dump_data.sh` kan de data van alle componenten (zaken, documenten, etc.) worden geëxporteerd naar een sql bestand.
```

docs/manual/scripts.rst
```diff
@@ -0,0 +1,35 @@
...
+
+Standaard wordt het volledige schema en data in twee sql bestanden gegenereerd. dit kan worden aangepast via de flags ``--data-only``, ``--schema-only`` & ``--combined``
+waardoor een bestand wordt gegenereerd. De data dump bevat standaard alle catalogi, zaak, besluit & document data.
+Om alleen specifieke data te exporteren kunnen de gewenste component namen worden meegegeven:
+
+.. code-block:: shell
+
+    /dump_data.sh zaken documenten
```
stevenbal, line 17:
```suggestion
    ./dump_data.sh zaken documenten
```

bin/dump_data.sh
```diff
@@ -0,0 +1,91 @@
...
+# pg_dump also does not add related tables automatically, so `dump_data.sh zaken` does not add related zaaktype data to the dump.
+
+
+set -e
+
+DEFAULT_APPS=(besluiten catalogi documenten zaken)
+
+export PGHOST=${DB_HOST:-db}
+export PGPORT=${DB_PORT:-5432}
+export PGUSER=${DB_USER:-postgres}
```
stevenbal, line 20:
```suggestion
export PGUSER=${DB_USER:-openzaak}
```

bin/dump_data.sh
```diff
@@ -0,0 +1,91 @@
...
+#
+# By default a schema and data dump are created separately. This can be changed with the flags --data-only, --schema-only
+# or --combined which appends the data dump to the schema dump.
+# The schema dump could not use -t to filter tables because this excludes extensions like postgis in the dump.
+# pg_dump also does not add related tables automatically, so `dump_data.sh zaken` does not add related zaaktype data to the dump.
+
+
+set -e
+
+DEFAULT_APPS=(besluiten catalogi documenten zaken)
```
stevenbal, line 16: One thing that's important to note is that EnkelvoudigInformatieObjecten point to specific files, so we should at least mention in the docs that the same files must exist in the new Open Zaak instance storage to properly export documenten.

bin/dump_data.sh
```diff
@@ -0,0 +1,91 @@
...
+export PGPASSWORD=${DB_PASSWORD:-""}
+
+SCRIPT=$(readlink -f "$0")
+SCRIPTPATH=$(dirname "$SCRIPT")
+
+${SCRIPTPATH}/wait_for_db.sh
+
+DUMP_FILE=${DUMP_FILE:-"dump_$(date +'%Y-%m-%d_%H-%M-%S').sql"}
+
+SCHEMA=true
```
stevenbal, line 31: When you load the schema, note that the django_migrations table will still be empty, so if you actually deploy an instance things will crash, because the tables already exist. I think we should at least document this (that you should run `src/manage.py migrate --fake`) or maybe we should include django_migrations in the data dump as well to avoid this

.github/workflows/bin-check.yml
```diff
@@ -0,0 +1,62 @@
...
+          grep "COPY public.besluiten_besluit" dump.sql
+          grep "COPY public.catalogi_resultaattype" dump.sql
+          grep "COPY public.documenten_gebruiksrechten" dump.sql
+          grep "COPY public.zaken_rol" dump.sql
+          ! grep "COPY public.auth_group" dump.sql
+
+      - name: load into other db
+        run: |
+          createdb -h localhost -U postgres test
+          psql -v ON_ERROR_STOP=1 -h localhost -U postgres -d test -f dump.sql
```
stevenbal, line 62: to fully ensure that it works, it might be good to also run `src/manage.py migrate` on this new database

stevenbal (changes requested):

PR #2130, opened 2025-07-23
Feature/2060 zaak registreren

### LINKED ISSUE #2060: Add `zaak_registreren` endpoint to create a Zaak with related resources
This endpoint should allow the creation of a Zaak with related resources to 

- (required) the Zaak itself
- (required) Rol(len) to be related to the zaak (i.e. for the `initiator` and `behandelaar`), let's allow an arbitrary number of rollen
- (optional) ZaakInformatieObjecten, an arbitrary number of relations to EnkelvoudigInformatieObjecten (that already exist in the Documenten API
- (optional) ZaakObjecten, also an arbitrary number
- (required) Status

Specification:

```
POST /api/v1/zaak_registreren

{
    "zaak": {...},
    "rollen": [{...}, {...}],
    "zaakinformatieobjecten": [{...}, {...}],
    "zaakobjecten": [{...}, {...}],
    "status": {...}
}
```

Floris272:
Closes #2060 

**Changes**
- adds zaak_registreren

[Describe the changes here]

**Checklist**

- Commit hygiene

### DIFF
--- a/src/openzaak/components/zaken/api/serializers/zaakobjecten.py
+++ b/src/openzaak/components/zaken/api/serializers/zaakobjecten.py
@@ -91,42 +91,46 @@ class ObjectTypeOverigeDefinitieSerializer(serializers.Serializer):
     )
 
 
+# TODO fix in cgac
+mapping = {
+    ZaakobjectTypes.adres: ObjectAdresSerializer(),
+    ZaakobjectTypes.besluit: None,
+    ZaakobjectTypes.buurt: ObjectBuurtSerializer(),
+    ZaakobjectTypes.enkelvoudig_document: None,
+    ZaakobjectTypes.gemeente: ObjectGemeenteSerializer(),
+    ZaakobjectTypes.gemeentelijke_openbare_ruimte: ObjectGemeentelijkeOpenbareRuimteSerializer(),
+    ZaakobjectTypes.huishouden: ObjectHuishoudenSerializer(),
+    ZaakobjectTypes.inrichtingselement: ObjectInrichtingselementSerializer(),
+    ZaakobjectTypes.kadastrale_onroerende_zaak: ObjectKadastraleOnroerendeZaakSerializer(),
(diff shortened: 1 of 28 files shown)


src/openzaak/components/zaken/api/serializers/zaakobjecten.py
```diff
@@ -91,42 +91,46 @@ class ObjectTypeOverigeDefinitieSerializer(serializers.Serializer):
     )
 
 
+# TODO fix in cgac
+mapping = {
```
Floris272, line 95: for Rol and Zaakobject i had to add the discriminator explicitly, otherwise it does not exist [here](https://github.com/maykinmedia/commonground-api-common/blob/main/vng_api_common/polymorphism.py#L171) in attrs

The zaakobject discriminator has another issue because it uses group_field, which is used to alter the fields in mapping so that everything is nested, When ZaakObjectSubSerializer is loaded it does it again so it becomes object_informatie__object_informatie__...

currently fixed it by copying the mapping to both separately, another solution would maybe be to not add the group_field in the  ZaakObjectSubSerializer or change something in cgac.

src/openzaak/components/zaken/tests/test_audittrails.py
```diff
@@ -65,6 +74,137 @@ def test_create_zaak_audittrail(self):
...
+            ],
+            "status": {
+                "statustype": statustype_url,
+                "datumStatusGezet": "2011-01-01T00:00:00",
+            },
+        }
+
+        response = self.client.post(url, data)
+
+        print(response.data)
```
stevenbal, line 171:
```suggestion
```

src/openzaak/components/zaken/api/serializers/zaakobjecten.py
stevenbal: Could you add a todo for this and make an issue to fix this properly?

stevenbal (changes requested):

stevenbal (approved):

PR #2146, opened 2025-08-15
:sparkles: [#2068] Add zaak_afsluiten endpoint to close a Zaak

### LINKED ISSUE #2068: Add `zaak_afsluiten` endpoint to close a Zaak
This endpoint should allow the closing of a Zaak 

- (required) Resultaat
- (required) the new Status of the Zaak (eindstatus)
- (required) the updated Zaak information

Specification:

```
POST /api/v1/afsluiten_zaak/<zaak_uuid>

{
    "resultaat": {...},
    "zaak": {...},
    "status": {...}
}
```

In the endpoint description it should be stated that the data under `zaak` is treated like it would be during a PATCH and the data under `resultaat` and `status` should be treated like they would be during a POST.

There should be validation:
- [ ] that the supplied status is indeed an eindstatus.
- [ ] it should not be possible to call `zaak_afsluiten` for the same Zaak again, unless the Zaak is reopened (could probably be achieved with `ClosedZaakMixin`)

Questions:
- what to do with `SCOPE_ZAKEN_GEFORCEERD_BIJWERKEN`, can a zaak be closed twice?

OlhaZahoruiko:
Closes #2068

**Changes**

[Describe the changes here]

**Checklist**

- Commit hygiene


src/openzaak/components/zaken/tests/test_zaken_afsluiten.py
```diff
@@ -0,0 +1,282 @@
...
+                "gezetdoor": f"http://testserver{reverse(self.rol)}",
+            },
+            "resultaat": {
+                "zaak": get_full_url(self.zaak),
+                "resultaattype": get_full_url(self.resultaattype),
+                "toelichting": "Behandeld",
+            },
+        }
+        response = self.client.post(self.url, payload)
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
```
stevenbal, line 271: Could you add assertions for the error message code here as well?

src/openzaak/components/zaken/api/serializers/zaken.py
```diff
@@ -1720,6 +1727,52 @@ def update(self, instance, validated_data):
...
+        resultaat_serializer = ResultaatSerializer(
+            data={**self.initial_data.get("resultaat", {}), **zaak_data},
+            context=self.context,
+        )
+        resultaat_serializer.is_valid(raise_exception=True)
+        resultaat = resultaat_serializer.save()
+
+        if not zaak.einddatum:
+            zaak.einddatum = now().date()
+            zaak.save(update_fields=["einddatum"])
```
stevenbal, line 1757: `einddatum` will be set by the `StatusSerializer` (based on `datum_status_gezet`), so I think this can be removed
```suggestion
```

stevenbal (changes requested):

stevenbal (approved):

PR #2150, opened 2025-08-19
Feature/2066 zaak verlengen

### LINKED ISSUE #2066: Add `zaak_verlengen` endpoint to set a recovery period for a Zaak
This endpoint should allow the creation of a Zaak with related resources to 

- (required) Status
- (required) Zaak

Specification:

```
POST /api/v1/zaak_verlengen/<zaak_uuid>

{
    "status": {...},
    "zaak": {...},
}
```

In the endpoint description it should be stated that the data under `zaak` is treated like it would be during a PATCH and the data under `status` is treated like it would be during a POST.

Floris272:
Closes #2066 

**Changes**

- adds zaak_verlengen endpoint
- refactors zaak update convenience endpoints into base class and subclasses

**Checklist**

- Commit hygiene

### DIFF
--- a/src/openzaak/components/zaken/api/viewsets.py
+++ b/src/openzaak/components/zaken/api/viewsets.py
@@ -2164,9 +2164,12 @@ class ZaakUpdateActionViewSet(
         return serializer.save()
 
 
-@extend_schema(
-    summary="Short een zaak op",
-    description=mark_experimental("Schort een zaak op en maak een status aan."),
+@extend_schema_view(
+    post=extend_schema(
+        "zaakopschorten",
+        summary="Schort een zaak op",
+        description=mark_experimental("Schort een zaak op en maak een status aan."),
+    )
 )
 class ZaakOpschortenViewset(ZaakUpdateActionViewSet):
     serializer_class = ZaakOpschortenSerializer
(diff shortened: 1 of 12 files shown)


src/openzaak/components/zaken/openapi.yaml
```diff
@@ -20625,6 +20700,397 @@ components:
...
+      - verlenging
+      - vertrouwelijkheidaanduiding
+      - zaakgeometrie
+      - zaakinformatieobjecten
+      - zaakobjecten
+      - zaaktype
+    ZaakVerlengingSubRequest:
+      type: object
+      properties:
+        opschorting:
```
stevenbal, line 21086: Not sure why, but this includes `opschorting` instead of `verlenging`

afbeelding

src/openzaak/components/zaken/api/viewsets.py
```diff
@@ -2070,14 +2071,9 @@ def perform_create(self, serializer):
         )
 
 
-@extend_schema(
-    summary="Short een zaak op",
-    description=mark_experimental("Schort een zaak op en maak een status aan."),
-)
-class ZaakOpschortenViewset(
+class ZaakUpdateActionViewSet(
```
stevenbal, line 2074: I seem to be getting this error when loading the redoc page:

```
Warning: operationId "None_post" has collisions [('/zaak_bijwerken/{uuid}', 'post'), ('/zaak_opschorten/{uuid}', 'post'), ('/zaak_verlengen/{uuid}', 'post')]. resolving with numeral suffixes.
```

Could you make sure they have an operationId?

src/openzaak/components/zaken/api/viewsets.py
```diff
@@ -2160,7 +2161,18 @@ def _create_audit_logs(self, response, serializer, zaak_version_before_edit):
         )
 
     def perform_post(self, serializer):
-        serializer.save()
+        return serializer.save()
+
+
+@extend_schema(
+    summary="Short een zaak op",
```
stevenbal, line 2168:
```suggestion
    summary="Schort een zaak op",
```

src/openzaak/components/zaken/api/viewsets.py
stevenbal, line 2074: Is this viewset also something that could be used for `zaak_afsluiten`?

src/openzaak/components/zaken/tests/test_zaak_verlengen.py
```diff
@@ -0,0 +1,399 @@
...
+            | {"statustype": f"http://testserver{self.statustype_url_2}"},
+        }
+
+        response = self.client.post(self.url, content)
+
+        self.assertEqual(
+            response.status_code, status.HTTP_400_BAD_REQUEST, response.data
+        )
+        error = get_validation_errors(response, "status.nonFieldErrors")
+        self.assertEqual(error["code"], "eindstatus-not-allowed")
```
stevenbal, line 382: Got this question from Olha (related to zaak_afsluiten), what happens if you try this with the `geforceerd-bijwerken` scope?

src/openzaak/utils/hooks.py
```diff
@@ -22,6 +22,10 @@ def get_lib_doc_excludes():
     """
     Exclude ConvertNoneMixin docstring from api spec generation
     """
-    from .serializers import ConvertNoneMixin, SubSerializerMixin
+    from .serializers import ConvertNoneMixin, SubSerializerMixin, ReadOnlyMixin
```
stevenbal, line 25: Could you fix the import order here?

stevenbal (changes requested):
Looks good overall :+1:

src/openzaak/components/zaken/api/viewsets.py
Floris272: fixed by changing to  `extend_schema_view`

stevenbal (approved):

PR #2187, opened 2025-09-19
:arrow_up: Bump commonground-api-common to 2.10.0

danielmursa-dev:
:arrow_up: Bump `commonground-api-common` to 2.10.0

- Update related to this issue [maykinmedia/open-api-framework#175]: added an exception logging event that includes code, status, and the invalidParams message.

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -68,7 +68,7 @@ click-repl==0.2.0
     # via celery
 cmislib-maykin==0.7.4
     # via drc-cmis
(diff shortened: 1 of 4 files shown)


stevenbal (changes requested):
@danielmursa-dev could you rebase on main and could you still add the issue refs to the commit messages? this makes it easier to remember what issue this was for when making the changelog

stevenbal (approved):

PR #2197, opened 2025-10-02
:sparkles: [#1621] add besluittype to zaaktype admin

### LINKED ISSUE #1621: Add the ability to, via the admin, add a zaaktype to an already published besluittype
### Thema / Theme

Admin

Taiga DH 566

It would be useful to allow the ability to link additional zaaktypes to a published besluittype via the admin. Currently the only way how to do so is via importing a zaaktype.

Linking zaaktypes to an already published informatieobjecttype is possible, but that is because this is a M2M-through table.

Not an explicit enhancement, just an example on how the admin could be made more functional when managing zaaktypes

More workable admin for zaaktypes

_No response_

Floris272:
Closes #1621

**Changes**

- aadd besluittype inline to zaaktype admin

[Describe the changes here]

**Checklist**

- Commit hygiene

### DIFF
--- a/src/openzaak/components/catalogi/admin/zaaktypen.py
+++ b/src/openzaak/components/catalogi/admin/zaaktypen.py
@@ -128,6 +128,15 @@ class ZaakTypeInformatieObjectTypeInline(EditInlineAdminMixin, admin.TabularInli
     fields = ZaakTypeInformatieObjectTypeAdmin.list_display
 
 
+class BesluitTypeInline(EditInlineAdminMixin, admin.TabularInline):
+    model = ZaakType.besluittypen.through
+    fk_name = "zaaktype"
+    fields = ("besluittype",)
+    raw_id_fields = ("besluittype",)
+    verbose_name = _("Besluittype")
+    verbose_name_plural = _("Besluittypen")
+
+
 @admin.register(ZaakType)
 class ZaakTypeAdmin(
     ReadOnlyPublishedMixin,
@@ -269,6 +278,7 @@ class ZaakTypeAdmin(
     raw_id_fields = ("catalogus", "deelzaaktypen")
     readonly_fields = ("versiedatum",)
     inlines = (
+        BesluitTypeInline,
         ZaakTypenRelatieInline,
         StatusTypeInline,
         RolTypeInline,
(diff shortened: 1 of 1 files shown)


src/openzaak/components/catalogi/admin/forms.py
```diff
@@ -49,6 +51,16 @@
...
 
 
+class ReadOnlyWidget(admin.widgets.ManyToManyRawIdWidget):
+    template_name = "admin/widgets/raw_id_read_only.html"
+
+    def format_value(self, value):
+        if not value:
+            return ""
+        models = self.rel.model.objects.filter(pk__in=value)
+        return ", ".join(str(model) for model in models) if value else ""
```
stevenbal: These are instances right?
```suggestion
        objects = self.rel.model.objects.filter(pk__in=value)
        return ", ".join(str(obj) for obj in objects) if value else ""
```

src/openzaak/components/catalogi/admin/forms.py
```diff
@@ -62,6 +74,23 @@ class ZaakTypeForm(forms.ModelForm):
...
+                field=ZaakType._meta.get_field("besluittypen"),
+                to=BesluitType,
+                through=ZaakType,
+            ),
+            admin_site=admin.site,
+        ),
+        help_text=_(
+            "De BESLUITTYPE(n) waaronder BESLUITEN kunnen voorkomen bij ZAAKen van dit ZAAKTYPE."
+        ),
+        label="Besluittypen",
```
stevenbal:
```suggestion
        label=_("Besluittypen"),
```

src/openzaak/components/catalogi/admin/forms.py
```diff
@@ -49,6 +51,16 @@
 )
 
 
+class ReadOnlyWidget(admin.widgets.ManyToManyRawIdWidget):
```
stevenbal: Now I'm wondering how this works for other raw_id fields, doesn't django already have something like this? :thinking:

stevenbal: so there is no existing widget in django already that we could reuse here?

PR #2238, opened 2025-11-11
:sparkles: [#2237] Add a callback that handles notifications

### LINKED ISSUE #2237: Add a callback that handles notifications (similar to https://github.com/maykinmedia/commonground-api-common/blob/main/vng_api_common/notifications/api/views.py#L16)
**To be implemented (Open Zaak)**
- [ ] Add a callback that handles notifications (similar to https://github.com/maykinmedia/commonground-api-common/blob/main/vng_api_common/notifications/api/views.py#L16)
  - [ ] `resource=object` and `actie=create` should result in ZaakObject(en) being created that links the `zaken` and `resourceUrl`
    - [ ] ZaakObject.objectType should be `overig` and ZaakObject.objectTypeOverige should be equal to objecttypeOmschrijving (?)
    - [ ] ZaakObject.relatieomschrijving (?)
  - [ ] `resource=object` and `actie=update` should result in the ZaakObject(en) being updated or deleted (if necessary). This is possible if the notificatie contains the list of linked `zaken`
  - [ ] `resource=object` and `actie=destroy` should result in the ZaakObject(en) being deleted

**Questions**

- [x] Is the context sufficiently specced? -> Probably
- [x] Do we need to more "payload"? -> Probably not
- [ ] What to do when the messages is not applicable -> ~Find out along the way~ niets doen (if statement),
- objecten kanaal

OlhaZahoruiko:
Closes #2237

**Changes**

[Describe the changes here]

**Checklist**

- Commit hygiene

### DIFF
--- a/src/openzaak/notifications/handler_objecten.py
+++ b/src/openzaak/notifications/handler_objecten.py
@@ -3,10 +3,12 @@
 
 from django.db import transaction
 
+import structlog
 from django_loose_fk.loaders import BaseLoader
 from vng_api_common.constants import CommonResourceAction
 
 KANAAL_OBJECTEN = "objecten"
+logger = structlog.stdlib.get_logger(__name__)
 
 
 @transaction.atomic
@@ -59,10 +61,35 @@ def handle(message: dict) -> None:
         ]
         ZaakObject.objects.bulk_create(to_create)
 
+        if to_create:
+            created_zaak_ids = [obj.zaak_id for obj in to_create]
+            logger.info(
+                "Created %d ZaakObject relation(s) for object '%s' to Zaak IDs: %s",
+                len(to_create),
+                resource_url,
+                created_zaak_ids,
+            )
+
     elif actie == CommonResourceAction.update:
-        ZaakObject.objects.filter(object=resource_url).exclude(
-            zaak_id__in=resolved_zaken_ids
-        ).delete()
+        to_delete_zaak_ids = (
+            ZaakObject.objects.filter(object=resource_url)
+            .exclude(zaak_id__in=resolved_zaken_ids)
+            .values_list("zaak_id", flat=True)
+        )
+
+        deleted_count, _ = (
+            ZaakObject.objects.filter(object=resource_url)
+            .exclude(zaak_id__in=resolved_zaken_ids)
+            .delete()
+        )
+
+        if deleted_count > 0:
+            logger.info(
+                "Deleted %d stale ZaakObject relation(s) for object '%s' from Zaak IDs: %s",
+                deleted_count,
+                resource_url,
+                list(to_delete_zaak_ids),
+            )
 
         existing_ids = set(
             ZaakObject.objects.filter(object=resource_url).values_list(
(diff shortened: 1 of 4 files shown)


src/openzaak/notifications/default.py
```diff
@@ -0,0 +1,43 @@
...
+from openzaak.components.zaken.models import ZaakObject
+
+
+def handle(message: dict) -> None:
+    resource = message.get("resource")
+    actie = message.get("actie")
+    resource_url = message.get("resourceUrl")
+    kenmerken = message.get("kenmerken", {})
+    objecttype_omschrijving = kenmerken.get("objecttypeOmschrijving")
+    zaken_ids = message.get("zaken", [])
```
stevenbal: I *think* the `zaken` field will contain URLs, so you can't directly do filtering by `zaak_id` (because those are the internal ids of the zaken.

I think what you could do is instantiate a `BaseLoader` and check for each URL if it is local (if it isn't, you can ignore it) and if it is, load the object with https://github.com/maykinmedia/django-loose-fk/blob/main/django_loose_fk/loaders.py#L69

Then you can use those instances in the code below

src/openzaak/notifications/default.py
```diff
@@ -0,0 +1,43 @@
...
+        to_keep = set(zaken_ids)
+        for obj in existing:
+            if obj.zaak_id not in to_keep:
+                obj.delete()
+        for zaak_id in zaken_ids:
+            ZaakObject.objects.get_or_create(
+                zaak_id=zaak_id,
+                object=resource_url,
+                defaults={"object_type": "overig"},
+            )
```
stevenbal: Could you check if this can be implemented without loops and instead with a `.exclude().delete()` and a `bulk_create`?

src/openzaak/notifications/default.py
```diff
@@ -0,0 +1,43 @@
...
+    actie = message.get("actie")
+    resource_url = message.get("resourceUrl")
+    kenmerken = message.get("kenmerken", {})
+    objecttype_omschrijving = kenmerken.get("objecttypeOmschrijving")
+    zaken_ids = message.get("zaken", [])
+
+    if resource != "object":
+        return
+
+    if actie == "create":
```
stevenbal: Could you use constants for this (and the other two elifs)? https://github.com/maykinmedia/commonground-api-common/blob/f9afc525e1ddceaef0e7252d06a7e6f618ba94de/vng_api_common/constants.py#L254

src/openzaak/notifications/views.py
```diff
@@ -0,0 +1,57 @@
+# SPDX-License-Identifier: EUPL-1.2
```
stevenbal: I think this file can be removed if the view from commonground-api-common is reused

stevenbal (changes requested):

src/openzaak/notifications/handler_objecten.py
```diff
@@ -0,0 +1,82 @@
...
+        return
+
+    loader = BaseLoader()
+    resolved_zaken_ids = []
+
+    for url in zaken_urls:
+        # Resolve local zaak URLs
+        for url in zaken_urls:
+            parsed = urlparse(url)
+            if parsed.netloc == "openzaak.example.com":
```
stevenbal: This is now hardcoded to compare to `openzaak.example.com`. You can use `BaseLoader.is_local_url` instead to check if the URL is local: https://github.com/maykinmedia/django-loose-fk/blob/c9acff87f0c9dae8cc2cd707bb043a85e74eda6b/django_loose_fk/loaders.py#L33

src/openzaak/notifications/handler_objecten.py
```diff
@@ -0,0 +1,82 @@
...
+                "zaak_id", flat=True
+            )
+        )
+        to_create = [
+            ZaakObject(
+                zaak_id=zaak_id,
+                object=resource_url,
+                object_type="overig",
+                object_type_overige=objecttype_omschrijving,
+                relatieomschrijving="",
```
stevenbal: Could you add a `# TODO` comment above this one? I'm not sure yet if we need to put something here

stevenbal (changes requested):

PR #2270, opened 2025-12-12
Create zaakobjecten backreferences

### LINKED ISSUE #2251: Create ZaakObjecten backreferences
maykinmedia/open-archiefbeheer#864

In order for Open Archief Beheer (OAB) to show objects to be destroyed/archived
linked to a Zaak, all these object should have a ZaakObject.

  - [x] OZ define a Cloud Event: `nl.overheid.zaken.zaak-gelinkt`
    data 
    ```
    {
    zaak: string (format URI), 
    linkTo: string (format URI),
    label: string (max length 80),
    linkObjectType: Optional[string],
    ```
    linkType is an indicator for OAB to decide which plugin to use for display/destruction
    linkTo
    See:  
    - open-zaak/open-notificaties#333
    - open-zaak/open-zaak#2195

  - [x] OZ listens to `nl.gemeente.zaken.link_zaak` events, and for each `zaak` it
    recognizes, adds
    [ZaakObject](https://openzaak.test.maykin.opengem.nl/zaken/api/v1/schema/#tag/zaakobjecten)
    with 
      - linkTo -> object
      - linkType -> objectType iff in enum else objectTypeOverige
      - label -> relatieomschrijving  or does OAB need to control this?

CharString:
Closes #2251

- [x] Needs docs after approval
- [ ] Needs SetupConfiguration? 

And needs way better tests. But I came to the conclusion that this project is not the correct place to test all the interdependencies.

After looking at what notification-api-common provided, and how open-inwoner does notification handling, I tried hard to shoe horn this in the existing notification zaak viewset, but it became a fancy, but entangled brittle mess switching on the content-type.

Then I found this reference implementation and it all became a lot simpler.

### DIFF
--- a/src/openzaak/components/autorisaties/middleware.py
+++ b/src/openzaak/components/autorisaties/middleware.py
@@ -76,12 +76,8 @@ class JWTAuth(_JWTAuth):
         return self._catalogus_cache[init_component]
 
     def has_auth(
-        self, scopes: List[str], init_component: str | None = None, **fields
+        self, scopes: List[str], component: str | None = None, **fields
     ) -> bool:
-        # TODO: super's 3rd param is named "component" not "init_component"
-        # What's the difference?
-        # Should init_component be keyword only or renamed?
-
         if scopes is None:
             return False
 
@@ -92,11 +88,11 @@ class JWTAuth(_JWTAuth):
         if self.has_alle_autorisaties:
             return True
 
-        if not init_component:
+        if not component:
             return False
 
-        autorisaties = self.get_autorisaties(init_component)
-        catalogus_autorisaties = self.get_catalogus_autorisaties(init_component)
+        autorisaties = self.get_autorisaties(component)
+        catalogus_autorisaties = self.get_catalogus_autorisaties(component)
         has_catalogus_autorisaties = catalogus_autorisaties.exists()
         scopes_provided = set()
 
--- a/src/openzaak/components/catalogi/api/validators.py
+++ b/src/openzaak/components/catalogi/api/validators.py
@@ -333,7 +333,7 @@ def is_force_write(serializer) -> bool:
 
     return request.jwt_auth.has_auth(
         scopes=SCOPE_CATALOGI_FORCED_WRITE,
-        init_component=serializer.Meta.model._meta.app_label,
+        component=serializer.Meta.model._meta.app_label,
     )
 
 
--- a/src/openzaak/components/catalogi/api/viewsets/mixins.py
+++ b/src/openzaak/components/catalogi/api/viewsets/mixins.py
@@ -38,7 +38,7 @@ class ConceptDestroyMixin:
     def perform_destroy(self, instance):
         forced_delete = self.request.jwt_auth.has_auth(
             scopes=SCOPE_CATALOGI_FORCED_DELETE,
-            init_component=self.queryset.model._meta.app_label,
+            component=self.queryset.model._meta.app_label,
         )
 
         if not forced_delete:
@@ -104,7 +104,7 @@ class M2MConceptDestroyMixin:
     def perform_destroy(self, instance):
         forced_delete = self.request.jwt_auth.has_auth(
             scopes=SCOPE_CATALOGI_FORCED_DELETE,
-            init_component=self.queryset.model._meta.app_label,
+            component=self.queryset.model._meta.app_label,
         )
 
         if not forced_delete:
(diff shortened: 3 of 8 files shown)


src/openzaak/urls.py
```diff
@@ -79,6 +80,7 @@
     path("ref/", include("notifications_api_common.urls")),
     # auth backends
     path("oidc/", include("mozilla_django_oidc.urls")),
+    path("events/", CloudEventWebhook.as_view(), name="cloudevent-webhook"),
```
stevenbal: Since this is just a single endpoint and all of the other API endpoints are without a trailing slash, let's keep it consistent with that
```suggestion
    path("events", CloudEventWebhook.as_view(), name="cloudevent-webhook"),
```

src/openzaak/notifications/viewsets.py
```diff
@@ -36,3 +43,43 @@ def _message(self, data, instance=None):
                 )
 
                 transaction.on_commit(lambda msg=message: send_notification.delay(msg))
+
+
+type CloudEvenHandler = Callable[[CloudEvent], None | NoReturn]
+
+
+class CloudEventWebhook(APIView):
```
stevenbal: This view currently doesn't require any auth and we also don't have any scopes (`vng_api_common.scopes.Scope`) defined for cloud events. The existing permission class `openzaak.utils.permissions.AuthRequired` doesn't seem to be appropriate for this because it requires a `main_resource`, which is not really applicable here.

Could you make sure a scope is added for publishing cloud events and that this scope is required for this view?

src/openzaak/notifications/viewsets.py
```diff
@@ -36,3 +43,43 @@ def _message(self, data, instance=None):
...
+    def register_handler[T: CloudEvenHandler](cls, f: T, /) -> T:
+        cls.handlers.add(f)
+        return f
+
+    def post(self, request: Request):
+        try:
+            event: CloudEvent = from_http(request.headers, request.body)
+        except GenericException as e:
+            raise ValidationError(
+                {"cloudevent": [str(e)]}, code="malformed-cloudevent"
```
stevenbal: here it's also maybe better to log the error and not return the full traceback as part of the API response?

stevenbal (changes requested):
@CharString could you briefly mention this `/events` endpoint and the supported event types (only `zaak-gelinkt` currently) in `docs/api/experimental.rst` in the `Cloud events` section?

src/openzaak/components/zaken/tests/test_cloud_events.py
```diff
@@ -1177,3 +1182,292 @@ def test_zaak_verlengen_post_sends_gemuteerd_cloud_event(self):
 
             self.assertEqual(response.status_code, status.HTTP_200_OK, response.data)
             self.assert_cloud_event_sent(ZAAK_GEMUTEERD, self.zaak, mock_send)
+
+
+class IncomingZaakCloudEventTests(JWTAuthMixin, APITestCase):
+    heeft_alle_autorisaties = True
```
stevenbal: To make sure it's tested that this specific scope gives permission to publish events
```suggestion
    heeft_alle_autorisaties = False
```

src/openzaak/notifications/viewsets.py
stevenbal: Ah you're right, sorry, I also didn't realize this is just the exception raised by `cloudevents` while parsing the event. Let's revert back `str(e)`, as this does give a clearer message:

with `str(e)`:

```
{
      "name": "cloudevent",
      "code": "malformed-cloudevent",
      "reason": "Found invalid specversion 1.a0"
}
```

with `__class__.__name__`:

```
{
      "name": "cloudevent",
      "code": "malformed-cloudevent",
      "reason": "InvalidRequiredFields"
}
```

src/openzaak/components/zaken/api/cloud_events.py
```diff
@@ -5,25 +5,85 @@
...
+        return
+
+    event_data = event.get_data()
+    if not event_data:
+        logger.warning("incoming_cloud_event_error", code="missing-data")
+        return
+
+    if not (zaak := _resolve_zaak_uri(event_data.get("zaak", ""))):
+        # info; zaak may live in other register
+        logger.info("incoming_cloud_event_error", code="unknown-zaak")
```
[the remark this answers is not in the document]
CharString: :thinking: Shall I add an info on success, so the successful handling of events will be completely clear? Or is the `zaakobject_created` message enough...

stevenbal (changes requested):
Two minor things remaining

PR #2297, opened 2026-01-15
:sparkles: [#2260] Add gerelateerdeZaken attribute to /zaken endpoint

### LINKED ISSUE #2260: Rename `relevanteAndereZaken` to `gerelateerdeZaken`
### Thema / Theme

Zaken API

The attribute `Zaak.relevanteAndereZaken` should be renamed to `Zaak.gerelateerdeZaken`. The name `relevanteAndereZaken` stems from a concept version of the specification and is confusing, because this name is used nowhere else in practice.

In addition to this, the relationship between Zaken via `gerelateerdeZaken` should be mirrored (only if both Zaken exist within the same zaaksysteem), which means that:
- If Zaak A is being related to Zaak B via `gerelateerdeZaken`, the reverse relationship should be visible from Zaak B as well
- If Zaak A has a relationship with Zaak B and this relationship is removed, the reverse relationship should be removed as well
- If possible, it would also be good to have this relationship visible in the audittrail of Zaak A and Zaak B (but this might be a larger issue)

Additionally, the fields `Zaak.relevanteAndereZaken.aardRelatie`, `Zaak.relevanteAndereZaken.overigeRelatie` and `Zaak.relevanteAndereZaken.toelichting` should be made optional and marked as deprecated

*During refinement*
* Out of scope is: Relations to cases outside Open Zaak. So, the reflexivity is not enforced but up to the client doing this for now -> might be changed later with notifications).
* audit trail out of scope.

_No response_

_No response_

stevenbal:
Closes #2260

I decided to use a different underlying database field (as opposed to using the same field as specified in the issue), because:
  - what happens to the additional metadata in relevanteAndereZaken if only gerelateerdeZaken is used (you could force it to overig, but it's a bit ugly)
  - the API behavior is a bit iffy if you want to preserve backwards compatibility (relevanteAndereZaken should take precedence over gerelateerdeZaken to avoid losing metadata on writes)
  - reflexivity is harder to implement, instead of saying symmetrical=True, we have to add signals + a data migration to fix this for `relevanteAndereZaken`? Also `relevanteAndereZaken` cannot be reflexive because of `aardRelatie`

TODO
- [ ] need to do datamigration in major release (add to a checklist somewhere?)

**Changes**

* Add gerelateerdeZaken attribute to /zaken endpoint
* Mark Zaak.relevanteAndereZaken as deprecated in OAS

**Checklist**

- Commit hygiene


Floris272 (approved):
Only thing i could find is that it can link to iself but it will not create the reverse so its fine.

PR #2342, opened 2026-03-09
:sparkles: [#2182] Fix recalculation of archiving when Besluit ingangsdatum or vervaldatum changes

### LINKED ISSUE #2182: Re-setting the archiving period for results of `Vervaldatum besluit` and other afleidingswijzen

1.23.0

When closing a _zaak_ using a result `Vervaldatum besluit`, the _Startdatum bewaartermijn_ is calculated correctly. However, when updating the _vervaldatum_ of the _besluit_ later on, the archiving period is **not** re-calculated.

In general, changes to a closed Zaak that could affect the `brondatum`, should cause the `brondatum` to be re-calculated, regardless of `afleidingswijze`.

**To do:**
- [ ] `afgehandeld` -> if `Zaak.einddatum` is changed, the brondatum should be recalculated based on the new `einddatum`
- [ ] `ander_datumkenmerk` -> should be updated on the Zaak manually?
- [ ] `eigenschap` -> Open Zaak needs to check if this afleidingswijze is used when the Eigenschap is modified
- [ ] `hoofdzaak` -> if the `einddatum` of the hoofdzaak changes, this change should also affect the deelzaken
- [ ] `ingangsdatum_besluit` -> if the ingangsdatum of the Besluit changes, this should affect the brondatum as well
- [ ] `termijn` -> if termijn on the resultaattype changes, this change should **not** affect Zaken that use the resultaattype that are already archived (e.g. if the laws around the duration of a permit change, this change should only affect new Zaken). If the brondatum needs to be changed `startdatum_bewaartermijn` and `archiefactiedatum` should be overridden manually?
- [ ] `vervaldatum_besluit` -> once this attribute is set on the Besluit, the brondatum for the zaak should be calculated
- [ ] `zaakobject` -> if it's a local object, Open Zaak could notice this when the local object is updated, but for external objects Open Zaak should receive a notificatie/cloud event about this? 
- [ ] When a Resultaat is removed from a Zaak, the archiving period should be reset and when a new Resultaat is added, the archiving period should be re-calculated
- [ ] Describe the above behavior in the docs (`docs/manual/archiving.rst`)

1. Open a _zaak_
2. Close it with a _besluit_ that has an _einddatum_
3. Update the _besluit_ to have a different _einddatum_

After updating the _einddatum_, the _Startdatum bewaartermijn_ get re-calculated

OlhaZahoruiko:
Closes #2182

**Changes**

[Describe the changes here]

**Checklist**

- Commit hygiene

### DIFF
--- a/src/openzaak/components/besluiten/models.py
+++ b/src/openzaak/components/besluiten/models.py
@@ -224,33 +224,48 @@ class Besluit(ETagMixin, AuditTrailMixin, APIMixin, models.Model):
             old_ingangsdatum = old_values["ingangsdatum"]
 
         super().save(*args, **kwargs)
+
         vervaldatum_changed = self.vervaldatum != old_vervaldatum
         ingangsdatum_changed = self.ingangsdatum != old_ingangsdatum
 
-        if vervaldatum_changed or ingangsdatum_changed:
-            from vng_api_common.constants import (
-                BrondatumArchiefprocedureAfleidingswijze,
+        if not (vervaldatum_changed or ingangsdatum_changed):
+            return
+
+        from vng_api_common.constants import BrondatumArchiefprocedureAfleidingswijze
+
+        from openzaak.components.zaken.archiving import try_calculate_archiving
+
+        zaak = self.zaak
+        if not zaak or not zaak.einddatum:
+            return
+
+        resultaat = getattr(zaak, "resultaat", None)
+        if not resultaat:
+            return
+
+        resultaattype = resultaat.resultaattype
+        afleidingswijze = resultaattype.brondatum_archiefprocedure.get(
+            "afleidingswijze"
+        )
+
+        if afleidingswijze in {
+            BrondatumArchiefprocedureAfleidingswijze.ingangsdatum_besluit,
+            BrondatumArchiefprocedureAfleidingswijze.vervaldatum_besluit,
+        }:
+            source_date = (
+                self.vervaldatum
+                if afleidingswijze
+                == BrondatumArchiefprocedureAfleidingswijze.vervaldatum_besluit
+                else self.ingangsdatum
             )
 
-            from openzaak.components.zaken.archiving import try_calculate_archiving
-
-            zaak = self.zaak
-            if not zaak or not zaak.einddatum:
-                return
-
-            resultaat = getattr(zaak, "resultaat", None)
-            if not resultaat:
-                return
-
-            resultaattype = resultaat.resultaattype
-            afleidingswijze = resultaattype.brondatum_archiefprocedure.get(
-                "afleidingswijze"
-            )
-
-            if afleidingswijze in {
-                BrondatumArchiefprocedureAfleidingswijze.ingangsdatum_besluit,
-                BrondatumArchiefprocedureAfleidingswijze.vervaldatum_besluit,
-            }:
+            if source_date is None:
+                zaak.archiefactiedatum = None
+                zaak.startdatum_bewaartermijn = None
+                zaak.save(
+                    update_fields=["archiefactiedatum", "startdatum_bewaartermijn"]
+                )
+            else:
                 try_calculate_archiving(zaak, force=True)
 
     def unique_representation(self):
--- a/src/openzaak/components/besluiten/models.py
+++ b/src/openzaak/components/besluiten/models.py
@@ -211,20 +211,46 @@ class Besluit(ETagMixin, AuditTrailMixin, APIMixin, models.Model):
         is_update = self.pk is not None
 
         old_vervaldatum = None
+        old_ingangsdatum = None
+
         if is_update:
-            old_vervaldatum = (
+            old_values = (
                 Besluit.objects.filter(pk=self.pk)
-                .values_list("vervaldatum", flat=True)
+                .values("vervaldatum", "ingangsdatum")
                 .first()
             )
 
-        super().save(*args, **kwargs)
+            old_vervaldatum = old_values["vervaldatum"]
+            old_ingangsdatum = old_values["ingangsdatum"]
 
-        if self.vervaldatum and self.vervaldatum != old_vervaldatum:
+        super().save(*args, **kwargs)
+        vervaldatum_changed = self.vervaldatum != old_vervaldatum
+        ingangsdatum_changed = self.ingangsdatum != old_ingangsdatum
+
+        if vervaldatum_changed or ingangsdatum_changed:
+            from vng_api_common.constants import (
+                BrondatumArchiefprocedureAfleidingswijze,
+            )
+
             from openzaak.components.zaken.archiving import try_calculate_archiving
 
             zaak = self.zaak
-            if zaak:
+            if not zaak or not zaak.einddatum:
+                return
+
+            resultaat = getattr(zaak, "resultaat", None)
+            if not resultaat:
+                return
+
+            resultaattype = resultaat.resultaattype
+            afleidingswijze = resultaattype.brondatum_archiefprocedure.get(
+                "afleidingswijze"
+            )
+
+            if afleidingswijze in {
+                BrondatumArchiefprocedureAfleidingswijze.ingangsdatum_besluit,
+                BrondatumArchiefprocedureAfleidingswijze.vervaldatum_besluit,
+            }:
                 try_calculate_archiving(zaak, force=True)
 
     def unique_representation(self):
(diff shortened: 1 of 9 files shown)


stevenbal (changes requested):
@OlhaZahoruiko the commit message mentions `vervaldatum`, but that was already implemented previously I think? Could you change the commit (and maybe split it up) per afleidingswijze for which recalculation was implemented? Currently it is not really clear from the commit message for which afleidingswijzen it was implemented and also not that it resetting the parameters was implemented when the Resultaat is deleted

Also could you add a test for the afleidingswijze `zaakobject` part (only if the related object is a local reference, we won't implement it for external references) mentioned in #2182 and implement that?

src/openzaak/components/zaken/tests/test_zaak_archive.py
```diff
@@ -1532,3 +1536,114 @@ def test_eigenschap_recalculates_archiefactiedatum_on_patch_update(self):
...
+        }
+
+        response = self.client.patch(besluit_url, data, **ZAAK_WRITE_KWARGS)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+        zaak.refresh_from_db()
+
+        self.assertEqual(zaak.archiefactiedatum, date(2026, 1, 1))
+
+    def test_zaak_einddatum_recalculates_archiefactiedatum_on_update(self):
```
stevenbal:
```suggestion
    def test_afgehandeld_zaak_einddatum_recalculates_archiefactiedatum_on_update(self):
```

src/openzaak/components/zaken/tests/test_zaak_archive.py
```diff
@@ -1532,3 +1536,114 @@ def test_eigenschap_recalculates_archiefactiedatum_on_patch_update(self):
...
+        self.assertIsNone(zaak.archiefactiedatum)
+
+        zaak.einddatum = date(2025, 1, 1)
+        zaak.save()
+
+        zaak.refresh_from_db()
+        self.assertEqual(zaak.startdatum_bewaartermijn, date(2025, 1, 1))
+        self.assertEqual(zaak.archiefactiedatum, date(2035, 1, 1))
+
+    def test_delete_resultaat_recalculates_archiefactiedatum(self):
```
stevenbal:
```suggestion
    def test_delete_resultaat_sets_archiefactiedatum_to_none(self):
```

stevenbal (changes requested):

src/openzaak/components/zaken/models/objecten.py
```diff
@@ -357,6 +357,43 @@ class WozWaarde(models.Model):
         help_text="De datum waarnaar de waarde van het WOZ-object wordt bepaald.",
     )
 
+    def save(self, *args, **kwargs):
```
stevenbal, line 384: Could you make sure the recalculation is also applied to the `Overige` model? Objects of this model could also contain dates, but since we don't know the shape, you can skip the check to see if a specific field was changed for that model. 

Let's also add a test for this

src/openzaak/components/zaken/models/zaken.py
```diff
@@ -1397,6 +1468,29 @@ def unique_representation(self):
         object = self.object.rstrip("/") if self.object.endswith("/") else self.object
         return f"({self.zaak.unique_representation()}) - {object.rsplit('/')[-1]}"
 
+    def save(self, *args, **kwargs):
```
stevenbal: I think that this override might not be necessary, because the relevant models related to ZaakObject that could contain dates are `WozWaarde` (for which you already added this logic in the `save`) and `Overige`.

At the time `ZaakObject` itself is saved, those related instances for `WozWaarde` are not updated yet, so we can remove this logic here I think


```suggestion
```

src/openzaak/components/zaken/tests/test_zaak_archive.py
```diff
@@ -1532,3 +1536,210 @@ def test_eigenschap_recalculates_archiefactiedatum_on_patch_update(self):
...
+
+        zaak.refresh_from_db()
+        self.assertIsNotNone(zaak.archiefactiedatum)
+
+        resultaat.delete()
+
+        zaak.refresh_from_db()
+
+        self.assertIsNone(zaak.archiefactiedatum)
+        self.assertIsNone(zaak.startdatum_bewaartermijn)
```
stevenbal: For this tests and the tests below it, could you add explicit assertions on the values of `archiefactiedatum` and `startdatum_bewaartermijn`?

src/openzaak/components/zaken/tests/test_zaak_archive.py
```diff
@@ -1532,3 +1536,210 @@ def test_eigenschap_recalculates_archiefactiedatum_on_patch_update(self):
...
+            },
+        )
+
+        zaak.refresh_from_db()
+        self.assertIsNotNone(zaak.archiefactiedatum)
+
+        resultaat_url = reverse("resultaat-detail", kwargs={"uuid": resultaat.uuid})
+        response = self.client.delete(resultaat_url)
+
+        self.assertEqual(response.status_code, 204)
```
stevenbal, line 1799: I'm sort of surprised to see that this returns a 204 here, I tested it locally via the API and the logic to reset the archiving parameters does work, but I run into this issue:

```
TypeError at /zaken/api/v1/resultaten/20dc7f12-51f0-4e39-9b34-e204de5439a3
Model instances without primary key value are unhashable
```

The issue seems to stem from this part: https://github.com/maykinmedia/commonground-api-common/blob/main/vng_api_common/caching/etags.py#L65

after deletion `instance` no longer has a pk, so hashing of it fails. On `main` this seems to work however, so I think some of the changes in the resultaat related code in this PR might have caused this?

src/openzaak/components/besluiten/models.py
```diff
@@ -211,20 +211,46 @@ def save(self, *args, **kwargs):
...
-                .values_list("vervaldatum", flat=True)
+                .values("vervaldatum", "ingangsdatum")
                 .first()
             )
 
+            old_vervaldatum = old_values["vervaldatum"]
+            old_ingangsdatum = old_values["ingangsdatum"]
+
         super().save(*args, **kwargs)
+        vervaldatum_changed = self.vervaldatum != old_vervaldatum
```
stevenbal, line 228: Because `Besluit.vervaldatum` is optional, it should also be possible to reset the archiving parameters to `null` if `Besluit.vervaldatum` is set to `null` again, this doesn't seem to work currently (not sure why though, the code seems to be able to handle it).

Could you add a test for this scenario and make sure it works?

src/openzaak/components/zaken/models/zaken.py
```diff
@@ -524,8 +524,63 @@ def save(self, *args, **kwargs):
...
+            resultaattype = resultaat.resultaattype
+
+            afleidingswijze = resultaattype.brondatum_archiefprocedure.get(
+                "afleidingswijze"
+            )
+
+            if afleidingswijze == BrondatumArchiefprocedureAfleidingswijze.hoofdzaak:
+                if subzaak.einddatum != self.einddatum:
+                    subzaak.einddatum = self.einddatum
+                    subzaak.save(update_fields=["einddatum"])
```
stevenbal: I think the `einddatum` for the subzaken is determined by the statuses set for those zaken individually, so I don't think we should force them to be the same as the einddatum of the hoofdzaak


```suggestion
```

src/openzaak/components/zaken/models/zaken.py
```diff
@@ -940,11 +995,27 @@ def save(self, *args, **kwargs):
...
+            from openzaak.components.zaken.archiving import try_calculate_archiving
+
+            try_calculate_archiving(zaak, force=True)
+
+            if not Resultaat.objects.filter(zaak=zaak).exists():
+                zaak.archiefactiedatum = None
+                zaak.startdatum_bewaartermijn = None
+                zaak.save(
+                    update_fields=["archiefactiedatum", "startdatum_bewaartermijn"]
+                )
```
stevenbal: Resultaat has a one to one relationship with Zaak, so after deleting a Resultaat, we can assume that there are no resultaten left for the zaak, so we can simplify this a bit:

```suggestion
                zaak.archiefactiedatum = None
                zaak.startdatum_bewaartermijn = None
                zaak.save(
                    update_fields=["archiefactiedatum", "startdatum_bewaartermijn"]
                )
```

stevenbal (changes requested):

stevenbal (approved):

PR #2386, opened 2026-05-12
Increase filename size

### LINKED ISSUE #2385: File name size too small

1.28.1 (related to GPP-Woo, issue from ICATT)

Trying to upload file with a name bigger then 100 chars via 'bestandsdeel-detail' url results in the following error:

```
django.core.exceptions.SuspiciousFileOperation: Storage can not find an available filename for '...' Please make sure that the corresponding file field allows sufficient "max_length".
```

1. create a EnkelvoudigInformatieObject via the api.
2. upload a file with a name longer then 100 chars.

Be able to upload files with a larger file name then 100 chars.

bart-maykin:
Closes #2385

**Changes**

Increased filename size.


stevenbal (approved):
Thanks for the fix, it seems correct, though for some reason I haven't managed to reproduce this locally via the API. I tried doing the same requests as in the test you added, but for some reason it seems to work fine locally in docker-compose

Could you show/share the requests you're doing that lead to this error?

alextreme:
Taiga issue ICATT 25 indicates that this is a problem with non-alphanumeric characters in the filename, and isn't related to the filename length https://taiga.maykinmedia.nl/project/icatt/issue/25

stevenbal:
@bart-maykin since the original taiga issue was closed, can I close this PR as well?

PR #2396, opened 2026-06-03
Feature/2372 uwv zaakidentificatie

### LINKED ISSUE #2372: Zaakidentificatie / zaaknummering - het mogelijk maken dat zaaknummer creatie op een andere wijze plaatsvindt
### Thema / Theme

Zaken API

UWV 22

Currently the zaakidentificatie is generated as follows:

 `ZAAK-2026-0000000014`

Client requests a different approach for generating the zaakidentificatie:

> Zaaknummer 
ZAAKNUMMER 
A00000006 t/m ZZ99999990 (nummers worden in oplopende volgorde volgorde uitgedeeld).

> Het zaaknummer bestaat uit 1 of 2 hoofdletters gevolgd door 8 cijfers.

> Een zaaknummer is valide indien het voldoet aan de volgende 11-proef (elfproef):

> Bepaal voor de eerste positie als volgt de waarde: 
het verschil tussen het teken en ‘A ‘ (dus ‘A’ = 0 en ‘Z’ = 25)
Bepaal voor de volgende 8 posities als volgt de waarde: 
het verschil tussen het teken en ‘0’ (dus ‘0’ = 0 en ‘9’ = 9); vermenigvuldig de waarde van elk teken met het getal van de positie in de string: eerste positie heeft getal 1, tweede positie heeft getal 2, …, negende positie heeft getal 9
tel dit op, deel de som door 11 (elf) en bepaal de rest

> Als de rest 10 is, dan is het een geldig zaaknummer.

> Indien de hoogste waarde met 9 posities bereikt is (Z99999990), dan wordt de reeks uitgebreid door een letter vóór de 9 posities te plaatsen en op dezelfde wijze de elfproef uit te voeren. Hierbij vermenigvuldigt Zomer de waarde van de eerste letter met 1, de waarde van de tweede letter met 2, het cijfer op de derde positie met 3 enzovoort. 
GENERATIE ZAAKNUMMER 
> Om een nieuwe nummer te genereren wordt een sequence uit de DB opgehaald en opgehoogd, de letter er voor gezet en getest of het geheel elfproef is. Zoniet, herhaal totdat zaaknummer elfproef is 

Estimation of Team:

The team estimated this on 10 days since we need to make the number-generator pluggable and support the existing number-generator scheme and the new one, which will also take some time to create.

_No response_

_No response_

Floris272:
Closes #2372 

**Changes**
- removes ZaakIdentificatieManager
- Adds Identification classes for existing year identification & UWV identification

**Checklist**

- Commit hygiene

- Architectural design record

### DIFF
--- a/docs/installation/case_identification.rst
+++ b/docs/installation/case_identification.rst
@@ -1,0 +1,16 @@
+===================
+Case Identification
+===================
+
+Open Zaak has multiple ``zaak identificatie`` generator options which are used when creating a zaak or when reserving a zaaknummer.
+A generator can be selected using ``ZAAK_IDENTIFICATIE_GENERATOR`` see :ref:`installation_env_config`.
+
+
+Standard options
+================
+* ``use-start-datum-year`` (default), uses the zaak start_datum for the identification. (e.g. ``ZAAK-2026-0000000001``)
+* ``use-creation-year`` (default), uses the current year when the zaak is created (or when zaaknummmmer is reserved). (e.g. ``ZAAK-2026-0000000001``)
+
+Specialized options
+===================
+* ``use-uwv-identification`` is a custom zaak identification for the UWV which uses an 11-proef. (e.g. ``A00000006``)
--- a/src/openzaak/components/zaken/api/serializers/zaken.py
+++ b/src/openzaak/components/zaken/api/serializers/zaken.py
@@ -26,6 +26,7 @@ import structlog
 from django_loose_fk.virtual_models import ProxyMixin
 from drf_spectacular.utils import extend_schema_serializer
 from drf_writable_nested import NestedCreateMixin, NestedUpdateMixin, UniqueFieldsMixin
+from models.identification_classes import get_base_identification_class
 from rest_framework import serializers
 from rest_framework.exceptions import ErrorDetail
 from rest_framework_gis.fields import GeometryField
@@ -229,7 +230,7 @@ class GenerateZaakIdentificatieSerializer(serializers.ModelSerializer):
         fields = ("bronorganisatie", "startdatum")
 
     def create(self, validated_data):
-        func = import_string(
+        identification_class = import_string(
             settings.ZAAK_IDENTIFICATIE_GENERATOR_OPTIONS.get(
                 settings.ZAAK_IDENTIFICATIE_GENERATOR,
                 settings.ZAAK_IDENTIFICATIE_GENERATOR_OPTIONS.get(
@@ -237,7 +238,7 @@ class GenerateZaakIdentificatieSerializer(serializers.ModelSerializer):
                 ),
             )
         )
-        return func(validated_data)
+        return identification_class(**validated_data).generate()
 
     def update(self, instance, data):  # pragma:nocover
         raise NotImplementedError("Updating is not supported in this serializer")
@@ -275,17 +276,15 @@ class ReserveZaakIdentificatieSerializer(serializers.ModelSerializer):
         bronorganisatie = validated_data["bronorganisatie"]
         today = date.today()
 
-        if aantal == 1:
-            return self.Meta.model.objects.generate(
-                bronorganisatie,
-                today,
-            )
+        identification_class = get_base_identification_class()
 
-        return self.Meta.model.objects.generate_bulk(
-            bronorganisatie,
-            today,
+        created_identifications = identification_class(
+            bronorganisatie, date=today
+        ).generate_bulk(
             aantal,
         )
+
+        return created_identifications[0] if aantal == 1 else created_identifications
 
     def update(self, instance, data):  # pragma:nocover
         raise NotImplementedError("Updating is not supported in this serializer")
--- a/src/openzaak/components/zaken/api/serializers/zaken.py
+++ b/src/openzaak/components/zaken/api/serializers/zaken.py
@@ -26,7 +26,6 @@ import structlog
 from django_loose_fk.virtual_models import ProxyMixin
 from drf_spectacular.utils import extend_schema_serializer
 from drf_writable_nested import NestedCreateMixin, NestedUpdateMixin, UniqueFieldsMixin
-from models.identification_classes import get_base_identification_class
 from rest_framework import serializers
 from rest_framework.exceptions import ErrorDetail
 from rest_framework_gis.fields import GeometryField
@@ -105,6 +104,7 @@ from ...models import (
     ZaakRelatie,
     ZaakVerzoek,
 )
+from ...models.identification_classes import get_base_identification_class
 from ..validators import (
     DateNotInFutureValidator,
     DeelzaakReopenValidator,
--- a/src/openzaak/components/zaken/models/identification_classes.py
+++ b/src/openzaak/components/zaken/models/identification_classes.py
@@ -1,3 +1,6 @@
+# SPDX-License-Identifier: EUPL-1.2
+# Copyright (C) 2026 Open Zaak maintainers
+
 from abc import ABC, abstractmethod
 from datetime import date
 
--- a/src/openzaak/components/zaken/models/identification_classes.py
+++ b/src/openzaak/components/zaken/models/identification_classes.py
@@ -18,11 +18,11 @@ from .identification import ZaakIdentificatie
 LOCK_ID_IDENTIFICATION_GENERATION = "generate-zaak-identification"
 
 
-class BaseIdentificatie(ABC):
-    def __init__(self, organisation: str, **kwargs):
-        self.organisation = organisation
+class BaseZaakIdentificatie(ABC):
+    model = ZaakIdentificatie
 
-        self.model = ZaakIdentificatie
+    def __init__(self, bronorganisatie: str, **kwargs):
+        self.bronorganisatie = bronorganisatie
 
     @abstractmethod
     def current(self) -> str:
@@ -53,7 +53,7 @@ class BaseIdentificatie(ABC):
         with pg_advisory_lock(LOCK_ID_IDENTIFICATION_GENERATION):
             return self.model.objects.create(
                 identificatie=next(self._sequence(self.current())),
-                bronorganisatie=self.organisation,
+                bronorganisatie=self.bronorganisatie,
             )
 
     def generate_bulk(self, amount: int):
@@ -70,14 +70,14 @@ class BaseIdentificatie(ABC):
         """
         with pg_advisory_lock(LOCK_ID_IDENTIFICATION_GENERATION):
             return self.model.objects.bulk_create(
-                self.model(identificatie=id, bronorganisatie=self.organisation)
+                self.model(identificatie=id, bronorganisatie=self.bronorganisatie)
                 for id in islice(self._sequence(self.current()), amount)
             )
 
 
-class YearIdentification(BaseIdentificatie):
-    def __init__(self, organisation: str, date: date, **kwargs):
-        super().__init__(organisation, **kwargs)
+class YearIdentification(BaseZaakIdentificatie):
+    def __init__(self, bronorganisatie: str, date: date, **kwargs):
+        super().__init__(bronorganisatie, **kwargs)
 
         self.prefix = f"ZAAK-{date.year}"
 
@@ -111,7 +111,7 @@ class StartDatumYearIdentification(YearIdentification):
         super().__init__(bronorganisatie, startdatum, **kwargs)
 
 
-class UWVIdentification(BaseIdentificatie):
+class UWVIdentification(BaseZaakIdentificatie):
     """
     Custom zaak identification for UWV
 
@@ -211,7 +211,7 @@ class UWVIdentification(BaseIdentificatie):
 
 def get_base_identification_class():
     match settings.ZAAK_IDENTIFICATIE_GENERATOR:
-        case "use-uvw-identification":
+        case "use-uwv-identification":
             identification_class = UWVIdentification
         case _:
             identification_class = YearIdentification
--- a/src/openzaak/components/zaken/models/identification_classes.py
+++ b/src/openzaak/components/zaken/models/identification_classes.py
@@ -1,17 +1,19 @@
 from abc import ABC, abstractmethod
 from datetime import date
 
+from django.conf import settings
 from django.db.models import Max
 from django.db.models.functions import Length
 
-from openzaak.components.zaken.models import ZaakIdentificatie
 from openzaak.utils.db import pg_advisory_lock
+
+from .identification import ZaakIdentificatie
 
 LOCK_ID_IDENTIFICATION_GENERATION = "generate-zaak-identification"
 
 
 class BaseIdentificatie(ABC):
-    def __init__(self, organisation: str):
+    def __init__(self, organisation: str, **kwargs):
         self.organisation = organisation
 
         self.model = ZaakIdentificatie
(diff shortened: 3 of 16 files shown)


src/openzaak/components/zaken/tests/test_identification.py
```diff
@@ -0,0 +1,190 @@
...
+
+        with self.assertRaises(ValueError):
+            self.uwv._increase("ZZ99999999")
+
+    def test_next(self):
+        self.assertEqual(self.uwv._next("A00000000"), "A00000006")
+        self.assertEqual(self.uwv._next("A00000006"), "A00000023")
+        self.assertEqual(self.uwv._next("A99999999"), "B00000001")
+        self.assertEqual(self.uwv._next("Z99999999"), "AA00000001")
+        self.assertEqual(self.uwv._next("Z99999988"), "Z99999990")
```
CharString: This is testing the implementation, not the functionality. All we need to know is that `_validate` is correct and that given some `_next(some_current)` validates and is strictly greater than `some_current` in the partial order that is defined on the identifications.

```suggestion
    def test_validate(self):
        self.assertTrue(self.uwv._validate("A00000006"))
        self.assertTrue(self.uwv._validate("A00000023"))

        self.assertTrue(self.uwv._validate("B00000001"))
        self.assertTrue(self.uwv._validate("Z99999990"))
        self.assertTrue(self.uwv._validate("ZZ99999985"))

        self.assertFalse(self.uwv._validate("A00000000"))
        self.assertFalse(self.uwv._validate("ZZ99999990"))
        
    @given(current=st.from_regex(r'^[A-Z]{1,2}[0-9]{8}$', fullmatch=True))
    def test_next(self, current):
        next = self.uwv._next(current)
        assert self.uwv._validate(next)
        assert next > current
```

replace `assert` with whatever `self.assert...` is needed here.

I don't know how to PBT validate without repeating the elfproef code in the test.

PS: While writing an Iterator implementation and testing if it was equivalent of the current one, I noticed `next > current` is not the only assertion. There should be no valid identifiers between next and current..

And this surfaced a bug in this current implementation:
```python
       assert (next_prefix == current_prefix) or next_pos == current_pos + 1
E       AssertionError: assert ('BZ' == 'AZ'
E         
E         - AZ
E         + BZ or 77 == (51 + 1))
E       Falsifying example: test_orig_uwv_generator(
E           current='AZ99999990',
E       )
```

Is this example actually correct?

```python
    def test_increase(self):
        ...
        self.assertEqual(self.uwv._increase("AZ99999999"), "BZ00000000")
```

I completely read passed this at first review.

src/openzaak/components/zaken/models/identification_classes.py
```diff
@@ -0,0 +1,206 @@
...
+                else:
+                    chars = chars[0] + increase_char(chars[1])
+
+        return f"{chars}{digits}"
+
+    def _next(self, current_identificatie: str) -> str:
+        next_identificatie = self._increase(current_identificatie)
+        while not self._validate(next_identificatie):
+            next_identificatie = self._increase(next_identificatie)
+        return next_identificatie
```
CharString: We can calculate 8th digit from the 7 preceding ones. Naively iterating till it validates works, but is expensive. We want to minimise the time we keep the lock.

CharString (changes requested):
While we're refactoring this anyway, I'd aim for an API that doesn't conflate identifier generation with database actions. Have it implement `Iterator` for ease of use, and separate from DB IO, so we can run lots of test examples, without being dragged down by IO:

```python
from itertools import chain, count, islice, starmap
from operator import mul
from typing import Iterable, Protocol, runtime_checkable

import pytest
from hypothesis import given, strategies as st

from openzaak.components.zaken.models.identification_classes import UWVIdentification

UWV = UWVIdentification("111222333")


class IdentificationGenerator[T](Iterable[T], Protocol):
    def __init__(self, initial: T, /, **kwargs): ...


class UWVGenerator:
    def __init__(self, init: str = "A" + 8 * "0", /, **_):
        # split current
        self._prefix = init.strip("0123456789")
        # sequence number without elf proef checksum
        self._seq_no = init[len(self._prefix) : len(self._prefix) + 7]

    @staticmethod
    def _as_int(char: str) -> int:
        return int(char) if char.isnumeric() else (ord(char) - 65)

    @staticmethod
    def _next_prefix(prefix: str) -> str:
        # convert to int (base 26 bijective)
        n = 0
        for char in prefix:
            # 64: A->1
            n = n * 26 + (ord(char) - 64)

        # next
        n += 1

        # convert back to str
        res = []
        while n > 0:
            n -= 1  # bijective! (A=1, not A=0)
            res.append(chr((n % 26) + 65))  # 65 is 'A'
            n //= 26

        return "".join(reversed(res))

    @staticmethod
    def _next(prefix: str, seq: str) -> tuple[str, str, int]:
        # static method, to show there is no state used from self
        if seq == "9999999":
            next_seq = f"{0:07n}"
            next_prefix = UWVGenerator._next_prefix(prefix)
        else:
            # if `init` wasn't valid *and* its checksum digit (init[-1]) was too
            # low, then adding 1 here will go passed the first valid, and we
            # *could* use prefix and seq to calculate the correct checksum
            #
            # But the common case *should* be that init *is* valid so I forego
            # doing a validity check and checksum comparison here
            next_seq = f"{int(seq) + 1:07d}"
            next_prefix = prefix

        if len(next_prefix) > 2:
            raise StopIteration("Max identification reached")

        values = map(UWVGenerator._as_int, chain(next_prefix, next_seq))
        weights = count(1)
        weighted_sum = sum(starmap(mul, zip(values, weights)))
        w = next(weights)  # weight of the checksum digit identifier[-1]

        # valid identifier is by definition:
        #   weighted_sum + checksum × w ≡ 10 (mod 11)
        # So algebra ♥ :
        #                  checksum × w = 10 − weighted_sum (mod 11)
        #                      checksum = (10 − weighted_sum) * w⁻¹ (mod 11)

        # This pow breaks when w is not co-prime to 11. 11 is prime so ValueError
        # is raised iff w is multiple of 11. But max len of an id is 2 + 8 and
        # we checked for len(next_prefix), so w < 11 always holds.
        checksum = ((10 - weighted_sum) * pow(w, -1, 11)) % 11

        if checksum == 10:
            # we need a single digit checksum
            return UWVGenerator._next(next_prefix, next_seq)
        return next_prefix, next_seq, checksum

    # By implementing the Iterator protocol, zip, enumerate, comprehensions all work so
    # bulk_generation is easy and flexible
    def __next__(self):
        self._prefix, self._seq_no, checksum = self._next(self._prefix, self._seq_no)
        return f"{self._prefix}{self._seq_no}{checksum}"

    def __iter__(self):
        return self


def _is_valid_uwv_identifier(value: str):
    # validation is not needed on the generator, it can be moved to the test
    return UWV._validate(value)


def uwv_identifier() -> st.SearchStrategy[str]:
    # this composite is better at searching the space for edge cases than
    # st.from_regex(r"^[A-Z]{1,2}[0-9]{8}$", fullmatch=True))
    return st.tuples(
        st.text(alphabet="ABCDEFGHIJKLMNOPQRSTUVWXYZ", min_size=1, max_size=2),
        st.integers(min_value=0, max_value=99_999_999).map(lambda n: f"{n:08d}"),
    ).map("".join)


def test_implements_protocol():
    assert isinstance(UWVGenerator, runtime_checkable(IdentificationGenerator))


# A, ..., Z, AA, AB, ..., AZ, BA, ... ZZ
_LETTER_SEQUENCE = [chr(n + 65) for n in range(26)] + [
    chr(fst + 65) + chr(snd + 65) for fst in range(26) for snd in range(26)
]
assert _LETTER_SEQUENCE[0] == "A"
assert _LETTER_SEQUENCE[_LETTER_SEQUENCE.index("AA") + 1] == "AB"
assert _LETTER_SEQUENCE[_LETTER_SEQUENCE.index("AZ") + 1] == "BA"


@given(current=uwv_identifier())
def test_uwv_generator(current):
    next_id = next(UWVGenerator(current))
    assert _is_valid_uwv_identifier(next_id)
    assert len(next_id) > len(current) or next_id > current

    # continuum hypothesis 🙃
    # assert there is no other prefix in between current and next
    current_prefix = current.strip("0123456789")
    next_prefix = next_id.strip("0123456789")
    current_pos = _LETTER_SEQUENCE.index(current_prefix)
    next_pos = _LETTER_SEQUENCE.index(next_prefix)
    assert (next_prefix == current_prefix) or next_pos == current_pos + 1


@given(current=uwv_identifier())
def test_orig_uwv_generator(current):
    next_id = UWV._next(current)
    assert _is_valid_uwv_identifier(next_id)
    assert len(next_id) > len(current) or next_id > current

    # continuum hypothesis 🙃
    # assert there is no other prefix in between current and next
    current_prefix = current.strip("0123456789")
    next_prefix = next_id.strip("0123456789")
    current_pos = _LETTER_SEQUENCE.index(current_prefix)
    next_pos = _LETTER_SEQUENCE.index(next_prefix)
    assert (next_prefix == current_prefix) or next_pos == current_pos + 1


def original_impl(current, length) -> list[str]:
    orig_sequence = []
    next_id = current
    for _ in range(length):
        orig_sequence.append((next_id := UWV._next(next_id)))
    return orig_sequence


def new_impl(current, length) -> list[str]:
    # turning into list here for the comparison,
    # but don't materialize the generator when doing a bulk create:
    # Model.objects.bulk_create(
    #     Model(identificate=id, bronorganisatie=constant_bronorganisatie)
    #     for id in islice(UWVGenerator(current_form_db), amount)
    # )
    return list(islice(UWVGenerator(current), length))


@pytest.mark.benchmark(group="UWV")
@pytest.mark.parametrize(
    "impl",
    [original_impl, new_impl],
)
@pytest.mark.parametrize(
    "initial_id",
    [
        "A00000000",
        "A00000008",
        "Z99999908",
        "AZ99999908",
    ],
)
@pytest.mark.parametrize("length", [1, 5, 100])
def test_benchmark(benchmark, impl, length, initial_id: str):
    # I wondered if all the iterators in the new implementation were slowing
    # things down. But this implementation is always faster ~4-6X on longer
    # batches
    benchmark.group = f"UWV, {length}"
    seq = benchmark(impl, initial_id, length)

    if impl is original_impl:
        return

    if initial_id != "AZ99999908":  # original has a bug
        if initial_id == "A00000000":  # new impl may skip first if initial is invalid
            assert seq[:-1] == original_impl(initial_id, length)[1:]
        else:
            assert seq == original_impl(initial_id, length)

```

src/openzaak/conf/includes/base.py
```diff
@@ -447,8 +447,9 @@
 
 
 ZAAK_IDENTIFICATIE_GENERATOR_OPTIONS = {
-    "use-creation-year": "openzaak.components.zaken.api.utils.generate_zaak_identificatie_with_creation_year",
-    "use-start-datum-year": "openzaak.components.zaken.api.utils.generate_zaak_identificatie_with_start_datum_year",
+    "use-creation-year": "openzaak.components.zaken.models.identification_classes.CreationYearIdentification",
+    "use-start-datum-year": "openzaak.components.zaken.models.identification_classes.StartDatumYearIdentification",
+    "use-uvw-identification": "openzaak.components.zaken.models.identification_classes.UWVIdentification",
```
stevenbal: I'm a bit hestitant to name this `uwv-identification`, but then again, I don't know any other name for this, maybe something mentioning that it uses the elf-proef (might be too generic though)? 

I also think it would be good to have a separate docs page to list these options and add some examples. We could refer to that page from the envvar helptext


```suggestion
    "use-uwv-identification": "openzaak.components.zaken.models.identification_classes.UWVIdentification",
```

src/openzaak/components/zaken/api/serializers/zaken.py
```diff
@@ -229,15 +230,15 @@ class Meta:
...
+        identification_class = import_string(
             settings.ZAAK_IDENTIFICATIE_GENERATOR_OPTIONS.get(
                 settings.ZAAK_IDENTIFICATIE_GENERATOR,
                 settings.ZAAK_IDENTIFICATIE_GENERATOR_OPTIONS.get(
                     "use-start-datum-year"
                 ),
             )
         )
-        return func(validated_data)
+        return identification_class(**validated_data).generate()
```
stevenbal, line 241: When doing a POST to /zaken via the API, I seem to be getting an error:

```
TypeError: BaseIdentificatie.__init__() missing 1 required positional argument: 'organisation'
```

Could you make sure there are tests for POST on zaken and the reserveer_zaaknummer endpoint with UWV identification?

src/openzaak/components/zaken/models/identification_classes.py
```diff
@@ -0,0 +1,219 @@
...
+from .identification import ZaakIdentificatie
+
+LOCK_ID_IDENTIFICATION_GENERATION = "generate-zaak-identification"
+
+
+class BaseIdentificatie(ABC):
+    def __init__(self, organisation: str, **kwargs):
+        self.organisation = organisation
+
+        self.model = ZaakIdentificatie
```
stevenbal: Could maybe be set as a class variable?

stevenbal (changes requested):

src/openzaak/conf/includes/base.py
CharString: @stevenbal Maybe it shouldn't even be offered to other users than UWV. The whole idea of the elfproef checksum is to prevent mix ups.  They could argue that if multiple orgs use this numbering scheme, it undermines that feature slightly.

src/openzaak/components/zaken/models/identification_classes.py
```diff
@@ -0,0 +1,219 @@
...
+        """
+
+        values = map(self._as_int, chain(prefix, seq))
+        weights = count(1)
+        weighted_sum = sum(starmap(mul, zip(values, weights)))
+        w = next(weights)  # weight of the checksum digit identifier[-1]
+
+        # This pow breaks when w is not co-prime to 11. 11 is prime so ValueError
+        # is raised if w is a multiple of 11. But max len of an id is 2 + 8, and
+        # we checked for len(next_prefix), so w < 11 always holds.
```
CharString: We didn't check it in this context... 
```suggestion
        # we checked for len(next_prefix) in _sequence, so w < 11 always holds.
```

CharString (changes requested):

src/openzaak/components/zaken/api/serializers/zaken.py
Floris272: added api tests & fixed error

src/openzaak/conf/includes/base.py
Floris272: added a docs page, The implementation too specific for a name like elf-proef since you could have many variations.

CharString (approved):

docs/installation/case_identification.rst
```diff
@@ -0,0 +1,16 @@
...
+
+
+Standard options
+================
+* ``use-start-datum-year`` (default), uses the zaak start_datum for the identification. (ZAAK-2026-0000000001)
+* ``use-creation-year`` (default), uses the current year when the zaak is created (or when zaaknummmmer is reserved). (ZAAK-2026-0000000001)
+
+Specialized options
+===================
+* ``use-uwv-identification`` is a custom zaak identification for the UWV which uses an 11-proef. (A00000006)
```
stevenbal:
```suggestion
* ``use-start-datum-year`` (default), uses the zaak start_datum for the identification. (e.g. ``ZAAK-2026-0000000001``)
* ``use-creation-year`` (default), uses the current year when the zaak is created (or when zaaknummmmer is reserved). (e.g. ``ZAAK-2026-0000000001``)

Specialized options
===================
* ``use-uwv-identification`` is a custom zaak identification for the UWV which uses an 11-proef. (e.g. ``A00000006``)
```

stevenbal (approved):

PR #2400, opened 2026-06-09
2369 ensure outgoing requests works the same as open formulieren

### LINKED ISSUE #2369: Ensure outgoing requests works the same as Open Formulieren
### Thema / Theme

Admin

1.5 years ago in #1784 it was decided to not enable LOG_REQUESTS by default, however by doing so it's not clear from the admin if log outgoing requests is working as expected or not when there are no requests logged.

As Joeri expected I'd like to revisit this decision for Open Zaak and go for the same approach as Open Formulieren, where LOG_REQUESTS is enabled by default but also where the settings for activating logoutgoingrequests resets after an hour for safety reasons: https://github.com/open-formulieren/open-forms/blob/eb5724c7e158d15bed458e3a3fbb06bd7eb5d13e/src/openforms/conf/base.py#L1311 

A bit of consistency when using our shared libraries across our components avoids surprises and friction when solving issues

Happier support

_No response_

CharString:
Closes #2369

💥 This changes configuration defaults! 

 - `LOG_REQUESTS`: enable logging of the outgoing requests. This must be enabled along with LOG_OUTGOING_REQUESTS_DB_SAVE to save outgoing request logs in the database. Defaults to: `True`.
 - `LOG_OUTGOING_REQUESTS_DB_SAVE_BODY`: Whether or not outgoing request bodies should be saved to the database. Defaults to: `False`.

✨ and adds a new option

 - `LOG_OUTGOING_REQUESTS_RESET_DB_SAVE_AFTER`: After the config has been changed via the admin, reset back to the default LOG_OUTGOING_REQUESTS_DB_SAVE after x minutes. Defaults to: 60.

Note:

A bug in the directive maykin-common 0.19.0 that makes it generate paragraphs
that start with `*` instead of bullet lists. But after that's fixed there, the
new docs look identical.


CharString:
After generation with a fixed maykin-common these are the visual diffs:
doc_diff

And screenshot from `Celery` group to bottom are the same:

```console
❯ odiff ~/Downloads/Sc*doc*.png celerydiff.png
Images are identical
```

stevenbal (approved):

PR #2422, opened 2026-06-24
2406 zaken zoek has query params in api spec which do not actually exist

### LINKED ISSUE #2406: zaken/_zoek has query params in api spec which do not actually exist

1.28

zaken/_zoek has query params in api spec which do not actually exist

Image

post zaken/_zoek with query param `identificatie` and add a filter to the body that includes multple zaken.

The zaken from the body filter will be returned and the param filter did not do anything.

the api spec should not show query params.

Took a quick look and found that `@extend_schema` can have `filters=False` which removes it from the api spec but sadly `AutoSchema.get_request_serializer` does use the filterset to build the body so `filters=False` will remove the fields from the body aswell.

CharString:
Closes #2406

**Changes**

- **:memo: [#2406] fix: remove _zoek query_parameters from schema**
  Ensure the VNG parameters are present
  

- **:bug: [#2406] fix: expand on /zaken/_zoek**
  Renamed the ZAAK_WRITE_KWARGS to POST_KWARGS, because we're not writing.
  

- **:label: chore: improve annotations on JWTAuthMixin**
  While making sense of the Auth in the _zoek tests, I bumped into these.
  
- **:memo: [#2406] fix: documenten/_zoek schema**
  Found out that EIO/_zoek was also broken.

  I find it odd that `uuid__in` is required. 
  It neuters the other search terms: in order to search for those needles,
  you have to bring your own uuid[] haystack.

NB :warning: CI workflows haven't run... Of course, I don't write bugs, so it's okay to merge anyway :upside_down_face:


src/openzaak/utils/mixins.py
```diff
@@ -49,9 +49,12 @@ def include_allowed(self):
         return self.action in ["list", "_zoek", "retrieve"]
 
     def get_requested_inclusions(self, request):
-        # Pull expand parameter from request body in case of _zoek operation
+        # Pull expand parameter from request body and/or query_param in case of _zoek operation
         if request.method == "POST":
-            return ",".join(request.data.get(self.expand_param, []))
+            return ",".join(
+                request.data.get(self.expand_param, [])
```
stevenbal, line 55: Could you document in the API spec that the values for expand in the request body and query param are joined together?

src/openzaak/components/documenten/openapi.yaml
```diff
@@ -2726,108 +2589,15 @@ paths:
                     type: string
                     format: uuid
                   description: Lijst van unieke resource identifiers (UUID4)
-                auteur:
```
stevenbal, line 2729: Did these additional search parameters in the request body actually work? The search parameters as defined in https://vng-realisatie.github.io/gemma-zaken/standaard/documenten/redoc-1.7.0#tag/enkelvoudiginformatieobjecten/operation/enkelvoudiginformatieobject__zoek seem to make the endpoint not that useful imo, since there's only three parameters there, which already exist as query parameters on the list endpoint anyway 😅

stevenbal (changes requested):

src/openzaak/components/documenten/openapi.yaml
CharString: They don't. I'll add the test I wrote as an `expectedFailure`.

IMHO the required `uuid__in` already makes it pretty challenged in the usefulness department. 🙃  But I guess that's by design; I think the aim is that you can't do blanket searches on documents, but can only narrow down the documents that you know exist from a zaak->zaakinformatieobject

stevenbal (approved):

PR #2423, opened 2026-06-26
Feature/2416 add api namespaces

### LINKED ISSUE #2416: Add namespaces to all apis
because of the duplicate reverse names in #2407 & #2408 the apis need a `app_name` /namespace. This means all reverse calls  some pkg functions that call reverse need to be updated

Floris272:
Closes #2416 

**Changes**

[Describe the changes here]

**Checklist**

- Commit hygiene

- Architectural design record


src/openzaak/utils/urls.py
```diff
@@ -0,0 +1,62 @@
...
+
+from django.conf import settings
+from django.db import models
+from django.urls import reverse as _reverse, reverse_lazy as _reverse_lazy
+
+
+def _magic_args(args, kwargs):
+    """
+    Do some trivial introspection to translate objects/models into common
+    used urls.
```
stevenbal, line 19: Could you add a comment here to mention that these were adapted from `commonground-api-common` to support the namespaces? And let's also mention why this can't be easily ported back to `commonground-api-common`

src/openzaak/utils/urls.py
```diff
@@ -0,0 +1,62 @@
+# SPDX-License-Identifier: EUPL-1.2
+# Copyright (C) 2026 Dimpact
+import inspect
+
+from django.conf import settings
+from django.db import models
+from django.urls import reverse as _reverse, reverse_lazy as _reverse_lazy
+
+
+def _magic_args(args, kwargs):
```
stevenbal: Could you apply some basic typehints to the functions here (at least the return types)?

src/openzaak/utils/serializer_fields.py
```diff
@@ -116,3 +138,37 @@ def _get_model_and_field(self) -> tuple:
         source = source.split("__")[0]
         model_field = model_class._meta.get_field(source)
         return model_class, model_field
+
+
+MOVED_MODELS = ["besluit", "besluitinformatieobject"]
```
stevenbal: Could you add a docstring to this to explain what these are for?

src/openzaak/utils/serializer_fields.py
```diff
@@ -116,3 +138,37 @@ def _get_model_and_field(self) -> tuple:
...
+    that the deprecated apis still return their original urls.
+    """
+
+    def get_url(
+        self, obj: Model, view_name: str, request: Request, format: str | None
+    ) -> str | None:
+        if obj._meta.model_name in MOVED_MODELS and getattr(
+            request, "resolver_match", None
+        ):
+            if request.resolver_match.namespace != "admin":
```
stevenbal, line 163: Did you run into requests resolving to `/admin` here? Sort of suprised to see that for a serializerfield

src/openzaak/utils/mixins.py
```diff
@@ -38,7 +39,21 @@ def audittrail(self):
...
+
+        reverse_kwargs = {"uuid": self.uuid}  # type: ignore[attr-defined]
+        reverse_kwargs.update(**kwargs)
+
+        url = reverse(
+            f"{namespace}:{resource_name}-detail",
+            kwargs=reverse_kwargs,
+            request=request,
+        )
+        return url
```
stevenbal: Could you add a comment to explain why is this override is needed?

src/openzaak/utils/hooks.py
```diff
@@ -29,3 +29,17 @@ def get_lib_doc_excludes():
...
         ReadOnlyMixin,
     ]
+
+
+DEPRECATED_URLCONFS = ["openzaak.components.besluiten.api.urls"]
+
+
+def postprocess_deprecate_apis(result, generator, request, public):
+    """
+    Deprecates all methods of urnconf.
```
stevenbal:
```suggestion
    Deprecates all methods of a predefined list of urlconfigs.
```

src/openzaak/components/zaken/api/serializers/zaakobjecten.py
```diff
@@ -189,6 +193,7 @@ class Meta:
                     ),
                     LooseFkIsImmutableValidator(),
                 ],
+                "view_name": "catalogi:zaakobjecttype-detail",
```
stevenbal, line 196: How was the view name inferred automatically before?

src/openzaak/components/documenten/tests/test_audittrails.py
```diff
@@ -288,7 +288,9 @@ def test_read_audittrail(self):
         eio = EnkelvoudigInformatieObject.objects.get()
         audittrails = AuditTrail.objects.get()
         audittrails_url = reverse(
-            audittrails, kwargs={"enkelvoudiginformatieobject_uuid": eio.uuid}
+            audittrails,
+            kwargs={"enkelvoudiginformatieobject_uuid": eio.uuid},
+            namespace="documenten",
```
stevenbal, line 293: How did this previously work without explicitly specifying that it was for documenten?

src/openzaak/components/besluiten/openapi.yaml
```diff
@@ -2479,8 +2494,6 @@ components:
           type: string
           format: uri
           description: URL-referentie naar het BESLUIT.
-          minLength: 1
-          maxLength: 1000
```
stevenbal, line 2483: Why were these removed?

src/openzaak/components/besluiten/openapi.yaml
```diff
@@ -2542,7 +2550,12 @@ components:
             API) waarin (een deel van) het besluit beschreven is.
           format: uri
           maxLength: 1000
+        besluit:
```
stevenbal: Where do these changes come from?

src/openzaak/components/besluiten/openapi.yaml
stevenbal, line 1: Could you make sure the API description contains a disclaimer that all of the endpoints in this API are deprecated and have been moved to the Zaken API? Let's mention that this API will be removed from Open Zaak in version 3.0

src/openzaak/utils/serializer_fields.py
```diff
@@ -89,7 +92,26 @@ def get_context_cache_key(field):
...
 
-class FKOrServiceUrlField(FKOrURLField):
+class ViewNameInjectionMixin:
+    """
+    Used for FKOrURLField classes to be able to set the view_name in the
+    serializer fields which is needed for the component namespaces
+    """
+
+    def __init__(self, *args, view_name: str | None = None, **kwargs):
+        # TODO FKOrServiceUrlField seems to work when changing this to a safer property but breaks OnlyRemoteOrFKOrURLField
```
stevenbal: Could you be a bit more specific as to how it breaks?

src/openzaak/tests/test_api_deprecation.py
stevenbal, line 1: I guess these could be added to `components/besluiten`?

src/openzaak/tests/test_api_deprecation.py
```diff
@@ -0,0 +1,123 @@
...
+            )
+            self.assertIn("besluiten/api/v1/besluiten", response.data["besluit"])
+            self.assertIn(
+                "documenten/api/v1/enkelvoudiginformatieobjecten",
+                response.data["informatieobject"],
+            )
+
+    def test_zaken_api_response(self):
+        with self.subTest("besluit"):
+            url = reverse(self.besluit, namespace="zaken")
```
stevenbal, line 53: What's the default behavior here actually if you do `reverse(self.besluit)`?

src/openzaak/components/zaken/openapi.yaml
```diff
@@ -103,52 +103,50 @@ info:
     name: EUPL 1.2
     url: https://opensource.org/licenses/EUPL-1.2
 paths:
```
stevenbal, line 105: The nested endpoints under `/zaken` seem to be absent (at least in redoc):

Before changes:
Image

After:

Image

src/openzaak/components/besluiten/api/serializers.py
```diff
@@ -169,7 +176,7 @@ class BesluitSubSerializer(SubSerializerMixin, BesluitSerializer):
     pass
 
 
-class BesluitInformatieObjectSerializer(serializers.HyperlinkedModelSerializer):
+class BesluitInformatieObjectSerializer(NamespacedHyperlinkedModelSerializer):
```
stevenbal: The description for this endpoint mentions aardRelatie and registratiedatum:

Image

But neither seem to be part of the response bodies:

Image

Could you check if this is the same as the Besluiten API standard and if so, remove the comment from the endpoint description?

stevenbal: This serializer seems to work fine with the `/besluiten/api/v1/besluiten/1234` urls, but if I try to create an object with the new `/zaken/api/v1/besluiten/1234` url I get this error (both via the new and the old endpoint):

```
{
  "type": "http://localhost:8000/ref/fouten/ValidationError/",
  "code": "invalid",
  "title": "Ongeldige invoer.",
  "status": 400,
  "detail": "Ongeldige invoer.",
  "instance": "urn:uuid:54d7f945-aadd-4efb-8f0a-19b69a2b5638",
  "invalidParams": [
    {
      "name": "besluit",
      "code": "incorrect_match",
      "reason": "Ongeldige hyperlink - Ongeldige URL"
    }
  ]
}
```

This also means that the new `/zaken/api/v1/besluitinformatieobjecten` endpoint is most likely untested?

stevenbal (changes requested):
I guess this PR also implements part of https://github.com/open-zaak/open-zaak/issues/2419?

I do think that several things require more attention though
* we need to determine what to do with the Besluiten API scopes (also in the autorisaties form): I think we should at least move them to the Zaken API?
* notificaties could also be affected, for backwards compatibility we could keep using `besluiten`, but it would probably be more correct to use `zaken`? (also ongoing discussion in the upstream issue)
* when will we move the actual models to `zaken`, and will the database table also be renamed then?

src/openzaak/utils/serializer_fields.py
Floris272: Yes because of the audittrailAdminMixin which uses the serializer to create the json object.

src/openzaak/components/zaken/api/serializers/zaakobjecten.py
Floris272: it uses <modelname>-detail by default

src/openzaak/components/documenten/tests/test_audittrails.py
Floris272: because of the enkelvoudiginformatieobject_uuid kwarg it knew which one to use

src/openzaak/tests/test_api_deprecation.py
Floris272: it uses the model _meta app_label by default so besluiten for besluit

src/openzaak/components/besluiten/api/serializers.py
Floris272: fixed with to_internal_value override in DeprecatedNamespaceLengthHyperlinkedRelatedField

Floris272: was also in besluiten  api, removed

src/openzaak/components/zaken/openapi.yaml
Floris272: seems to have been fixed now

src/openzaak/tests/test_api_deprecation.py
Floris272: I would like to keep it separately since it is for multiple components and catalogi api deprecation tests could also be added here.

src/openzaak/components/besluiten/openapi.yaml
Floris272: added

src/openzaak/utils/urls.py
```diff
@@ -0,0 +1,69 @@
...
+import inspect
+from typing import Any
+
+from django.conf import settings
+from django.db import models
+from django.urls import reverse as _reverse, reverse_lazy as _reverse_lazy
+
+
+def _magic_args(
+    args: tuple[Any, ...], kwargs: dict[str, Any]
```
stevenbal: Let's replace `Any` with `object` (applies to all uses in this file), `Any` is essentially just a way to ignore typechecking altogether, `object` implies it could be any basic object

src/openzaak/utils/serializer_fields.py
stevenbal: could you add that as a comment?

src/openzaak/components/besluiten/api/schema.py
```diff
@@ -8,7 +8,11 @@
 
 from .kanalen import KANAAL_BESLUITEN
 
-description = f"""Een API om een besluitregistratiecomponent (BRC) te benaderen.
+description = f"""
+
+**Deze API is deprecated en zal worden verwijderd in 3.0. Besluiten zijn verplaatst naar de zaken API.**
```
stevenbal:
```suggestion
**Deze API is deprecated en zal worden verwijderd in Open Zaak versie 3.0. Besluiten zijn verplaatst naar de Zaken API.**
```

Could you add the same deprecation warning in the dashboard?
Image

src/openzaak/components/zaken/openapi.yaml
stevenbal: They don't seem to show up for me yet
image

src/openzaak/components/besluiten/api/audits.py
```diff
@@ -2,4 +2,4 @@
 # Copyright (C) 2019 - 2020 Dimpact
 from vng_api_common.audittrails.audits import Audit
 
-AUDIT_BRC = Audit("BRC", "besluit")
+AUDIT_BRC = Audit("BRC", "besluit")  # TODO just change this to ZRC?
```
stevenbal: Is this something for another PR?

src/openzaak/components/besluiten/api/serializers.py
```diff
@@ -231,8 +252,13 @@ def create(self, validated_data):
 class BesluitInformatieObjectSubSerializer(
     SubSerializerMixin, BesluitInformatieObjectSerializer
 ):
-    class Meta(BesluitInformatieObjectSerializer.Meta):
-        read_only_fields = ("besluit",)
+    besluit = DeprecatedNamespaceLengthHyperlinkedRelatedField(
+        view_name="zaken:besluit-detail",
+        lookup_field="uuid",
+        help_text=get_help_text("besluiten.BesluitInformatieObject", "besluit"),
+        validators=[IsImmutableValidator()],
```
stevenbal, line 254: Does this field not need a queryset (like the declaration in `BesluitInformatieObjectSerializer`)?

src/openzaak/components/zaken/tests/admin/besluiten/test_besluit_audittrail.py
```diff
@@ -48,7 +48,7 @@ def _create_besluit(self):
 
     def test_create_besluit(self):
         besluit = self._create_besluit()
-        besluit_url = get_operation_url("besluit_read", uuid=besluit.uuid)
+        besluit_url = reverse(besluit, namespace="zaken")
```
stevenbal, line 50: Shouldn't these tests use the `besluiten` namespace instead, since they are in the `besluiten` component? Also applies for the other audittrail tests

src/openzaak/components/besluiten/tests/test_besluit_with_external_zaak.py
```diff
@@ -154,7 +155,7 @@ def test_update_external_zaak(self):
         # update zaak in the besluit
         zaak_new = f"{self.base}zaken/{uuid.uuid4()}"
         zaakbesluit_new_data = get_zaakbesluit_response(zaak_new)
-        besluit_url = f"http://testserver{reverse(besluit)}"
+        besluit_url = f"http://testserver{reverse(besluit, namespace='zaken')}"
```
stevenbal: Should probably also use `besluiten` namespace?

src/openzaak/components/zaken/tests/besluiten/test_audittrails.py
```diff
@@ -0,0 +1,330 @@
...
+        self.assertEqual(besluit_create_audittrail.actie, "create")
+        self.assertEqual(besluit_create_audittrail.resultaat, 201)
+        self.assertEqual(besluit_create_audittrail.oud, None)
+        self.assertEqual(besluit_create_audittrail.nieuw, besluit_response)
+
+    def test_update_besluit_audittrails(self):
+        besluit_data = self._create_besluit()
+
+        modified_data = deepcopy(besluit_data)
+        url = modified_data.pop("url")
```
stevenbal, line 66: I checked this locally, it seems that the default behavior still points to `/besluiten/api/v1/` instead of `/zaken/api/v1/`

```
(Pdb) url
'http://testserver/besluiten/api/v1/besluiten/40a45fab-154e-4b42-bca5-048f98677515'
(Pdb) besluit_response
{'url': 'http://testserver/besluiten/api/v1/besluiten/40a45fab-154e-4b42-bca5-048f98677515', 'identificatie': '123123', 'verantwoordelijke_organisatie': '000000000', 'besluittype': 'http://testserver/catalogi/api/v1/besluittypen/3381fbe2-5b83-4703-bd82-fb9685839789', 'zaak': '', 'datum': '2019-04-25', 'toelichting': 'aangepast', 'bestuursorgaan': '', 'ingangsdatum': '2019-04-26', 'vervaldatum': '2019-04-28', 'vervalreden': '', 'vervalreden_weergave': '', 'publicatiedatum': None, 'verzenddatum': None, 'uiterlijke_reactiedatum': None}
```

src/openzaak/components/zaken/tests/besluiten/test_audittrails.py
```diff
@@ -0,0 +1,330 @@
+# SPDX-License-Identifier: EUPL-1.2
+# Copyright (C) 2019 - 2022 Dimpact
```
stevenbal, line 2: Could you add  a comment to `src/openzaak/components/besluiten/tests/__init__.py` to mention that the tests have been duplicated to `src/openzaak/components/zaken/tests/besluiten` because the besluit resource has been moved to Zaken API?

src/openzaak/utils/serializer_fields.py
```diff
@@ -116,3 +146,86 @@ def _get_model_and_field(self) -> tuple:
...
+
+    def to_internal_value(self, data):
+        """
+        Override for rest_framework HyperlinkedRelatedField.to_internal_value for incorrect matches because of deprecated apis.
+
+        all mappings in _ALLOWED_INCORRECT_MATCHES are allowed
+        """
+
+        request = self.context.get("request")
+        try:
```
stevenbal: Do we need to duplicate the entire function, or could we call super(), and catch the `self.fail("incorrect_match")` maybe?

src/openzaak/utils/mixins.py
```diff
@@ -36,9 +39,28 @@ def audittrail(self):
...
+
+        reverse_kwargs = {"uuid": self.uuid}  # type: ignore[attr-defined]
+        reverse_kwargs.update(**kwargs)
+
+        url = reverse(
+            f"{namespace}:{resource_name}-detail",
+            kwargs=reverse_kwargs,
+            request=request,
+        )
+        return url if with_namespace else url.replace(f"/{namespace}", "", 1)
```
stevenbal: But removing the `namespace` means that it's no longer the correct URL, so I'm not sure when you would want to use this?

src/openzaak/tests/test_api_deprecation.py
```diff
@@ -0,0 +1,367 @@
...
+                        "bestuursorgaan": "",
+                        "publicatiedatum": None,
+                        "vervalredenWeergave": "",
+                        "uiterlijkeReactiedatum": None,
+                        "verantwoordelijkeOrganisatie": "517439943",
+                    },
+                },
+            )
+
+        # TODO urls are stored in auditrail itself and wont changes when changing namespace
```
stevenbal: Not sure if there is an easy or clean way around this, with domains it's slightly easier because we could use the ServiceURLField, but even then the path is still fixed.

stevenbal (changes requested):

src/openzaak/components/besluiten/api/serializers.py
Floris272: It is not needed (and not allowed) because its read only

src/openzaak/utils/mixins.py
Floris272: Added it because of the audittrail filter in AuditTrailMixin. using the correct url it will only get the audittrails created in that namespace. 

```python
class AuditTrailMixin:
    @property
    def audittrail(self):
        qs = AuditTrail.objects.filter(
            hoofd_object__contains=self.get_absolute_api_url(
                version=1, with_namespace=False
            )
        ).order_by("-aanmaakdatum")
...
```

src/openzaak/components/zaken/tests/admin/besluiten/test_besluit_audittrail.py
Floris272: the admin audittrails have already been moved over to zaken since they use the default viewset namespace (which is zaken)

src/openzaak/components/besluiten/api/schema.py
Floris272: added

src/openzaak/components/besluiten/tests/test_besluit_with_external_zaak.py
Floris272: I changed the BesluitSerializer.create_zaakbesluit reverse call to zaken already. The besluit_url in this test is for the url used in the external OZ instance. I could look into fetching the current namespace but  external stuff will be removed soon.

src/openzaak/utils/serializer_fields.py
Floris272: got it working in fail() by using request.resolver_match.namespace again.

src/openzaak/utils/mixins.py
Floris272: Also had to add namespace param for the call in verwerkbesluitSerializer.

src/openzaak/components/zaken/openapi.yaml
Floris272: it was probably because the commonground pkg was using a branch tag. it now uses the new 2.13.0 release

stevenbal: Ah my bad, fixed now indeed :+1:

src/openzaak/components/zaken/tests/besluiten/test_audittrails.py
stevenbal: This one still seems to be the same, I think the namespace "zaken" should be explicitly passed to the `reverse(Besluit)` call? Could you also add an assertion on the url of the besluit to verify that the new path with `/zaken` is used?

src/openzaak/components/zaken/tests/besluiten/test_auth.py
```diff
@@ -0,0 +1,650 @@
...
+            applicatie=self.applicatie,
+            component=self.component,
+            scopes=self.scopes,
+            max_vertrouwelijkheidaanduiding=self.max_vertrouwelijkheidaanduiding,
+        )
+
+        besluit1 = BesluitFactory.create(besluittype=self.besluittype)
+        besluit2 = BesluitFactory.create()
+
+        url = reverse(BesluitInformatieObject)
```
stevenbal: I think the same issue about using the old namespace still applies for most tests under zaken:

```
(Pdb) url
'/besluiten/api/v1/besluitinformatieobjecten'
```

Or are there separate tests to verify that the new path `/zaken/api/v1/...` works?

src/openzaak/utils/mixins.py
stevenbal: If the `with_namespace` param is only needed for audittrails, can't this issue be fixed in the `AuditTrailMixin` instead? I don't want `get_absolute_api_url` to return something that is not the full URL (or relative path if request is not specified)

src/openzaak/components/besluiten/tests/test_besluit_with_external_zaak.py
stevenbal: Ah sorry, didn't see that this was for external relations

src/openzaak/components/zaken/tests/admin/besluiten/test_besluit_audittrail.py
stevenbal: But I would assume that the old tests under `besluiten/tests/` keep using the old namespaces, to verify that that still works?

stevenbal (changes requested):

src/openzaak/utils/mixins.py
```diff
@@ -36,9 +37,32 @@ def audittrail(self):
 
 
 class APIMixin(_APIMixin):
```
stevenbal: Since there are no more super() calls happening here, I think we can remove the subclass

```suggestion
class APIMixin:
```

stevenbal (approved):

stevenbal:
@Floris272 forgot about this earlier, but we should probably mark the added besluiten resources in zaken API as experimental (because it's not yet part of the ZGW standard, planned for 1.8 https://github.com/VNG-Realisatie/gemma-zaken/issues/2622)

PR #2424, opened 2026-06-28
Fixes #2313 -- use Service mTLS certificates when validating remote URLs

### LINKED ISSUE #2313: Creating ZaakObject with url requiring mTLS fails

1.26.0

openzaak fails to create a ZaakObject when the url provided requires mTLS

the cause is that you cannot configure the certificates required for mTLS under services like you can in (for example) open-formulieren

you can only configure authorization headers in the web interface for the url

1. specify a url in services for your BAG provider
2. create a ZaakObject for an address with a source url pointing to this service
3. openzaak then does a request to this url and expects a http status code 200
4. if the url requires mTLS (with for instance a private certificate chain+key) the request will fail
5. openzaak will not create the ZaakObject

you should be able to configure the certificate or maybe you should be able to disable the verification requiring the http status code 200

cinpol:
Closes #2313

**Changes**

Creating a `ZaakObject` whose `object` URL points to a service requiring mutual TLS (mTLS) failed during URL validation. The `LINK_FETCHER (openzaak.nlx.fetcher)` already resolves the matching `zgw_consumers Service` to apply NLX rewriting, but it then called `requests.get()` without the client/server certificates that are already configurable on that `Service` — so the certificates were never used.

This change passes the Service's `server_certificate` and `client_certificate` to requests as the `verify` and `cert` arguments, mirroring `zgw_consumers.client.ServiceConfigAdapter.get_client_session_kwargs` so behaviour is consistent with the rest of the stack. Explicitly supplied `verify`/`cert` kwargs still take precedence. No model or admin changes are required, since the certificate fields already exist on `Service`.

Regression tests are added in `openzaak/nlx/tests/test_fetcher.py` covering: server certificate → `verify`, client certificate (with and without private key) → `cert`, the no-service and no-certificate no-op cases, explicit-kwargs precedence, and preservation of NLX rewriting.

**Checklist**

- Commit hygiene

- Architectural design record

  - [x] If a design decision was made that changes functionality, create an architectural design record for it [here](https://github.com/open-zaak/open-zaak/wiki/Architectural-Decision-Records-(ADR)) `(N/A — bugfix; uses existing certificate configuration, no functional design decision)`

**Note:** I recently submitted a vulnerability report and had some spare time, so I looked into open issues I could put together a PR for. Apologies in advance if anything here is off — happy to adjust based on your feedback.

### DIFF
--- a/src/openzaak/nlx/__init__.py
+++ b/src/openzaak/nlx/__init__.py
@@ -7,11 +7,29 @@ from zgw_consumers.models import Service
 def fetcher(url: str, *args, **kwargs):
     """
     Fetch the URL using requests.
+
     If the NLX address is configured, rewrite absolute url to NLX url.
+
+    If the matching :class:`zgw_consumers.models.Service` has TLS certificates
+    configured, use them so that endpoints requiring a server and/or client
+    certificate (mutual TLS) can be reached during validation. See #2313.
     """
     service = Service.get_service(url)
-    if service and service.nlx:
-        # rewrite url
-        url = url.replace(service.api_root, service.nlx, 1)
+    if service:
+        if service.nlx:
+            # rewrite url
+            url = url.replace(service.api_root, service.nlx, 1)
+
+        # apply the (mutual) TLS configuration of the service, mirroring
+        # zgw_consumers.client.ServiceConfigAdapter.get_client_session_kwargs
+        if "verify" not in kwargs and (server_cert := service.server_certificate):
+            kwargs["verify"] = server_cert.public_certificate.path
+        if "cert" not in kwargs and (client_cert := service.client_certificate):
+            client_cert_path = client_cert.public_certificate.path
+            kwargs["cert"] = (
+                (client_cert_path, privkey.path)
+                if (privkey := client_cert.private_key)
+                else client_cert_path
+            )
 
     return requests.get(url, *args, **kwargs)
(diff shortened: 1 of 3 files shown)


src/openzaak/nlx/__init__.py
```diff
@@ -7,11 +7,29 @@
...
+        # zgw_consumers.client.ServiceConfigAdapter.get_client_session_kwargs
+        if "verify" not in kwargs and (server_cert := service.server_certificate):
+            kwargs["verify"] = server_cert.public_certificate.path
+        if "cert" not in kwargs and (client_cert := service.client_certificate):
+            client_cert_path = client_cert.public_certificate.path
+            kwargs["cert"] = (
+                (client_cert_path, privkey.path)
+                if (privkey := client_cert.private_key)
+                else client_cert_path
+            )
```
stevenbal, line 33: I think instead of injecting these manually, I'd rather let `zgw_consumers.client.build_client` take care of 
this, since that also calls `ServiceConfigAdapter.get_client_session_kwargs`

```suggestion
        client = build_client(service)
        return client.get(url, *args, **kwargs)
```

stevenbal (changes requested):
Thanks for your contribution!

Small suggestion regarding the fix. Additionally I think it would be good to have a regression test for the specific error mentioned in #2313 as well (in `src/openzaak/components/zaken/tests/test_zaakobject.py`) to assert that certificates configured are picked up

src/openzaak/nlx/__init__.py
OlhaZahoruiko: https://github.com/open-zaak/open-zaak/pull/2564

stevenbal:
closing this in favor of https://github.com/open-zaak/open-zaak/pull/2564

PR #2450, opened 2026-07-10
Feature/oo 757 upgrade notifications api common

Floris272:
**Changes**

- migrates failednotification to n-a-c failednotification
- removes failednotification model & code
- fixes tests

**Checklist**

- Commit hygiene

- Architectural design record


src/openzaak/notifications/migrations/0005_migrate_failednotifications.py
```diff
@@ -0,0 +1,40 @@
...
+
+    notifications = []
+    notification_responses = []
+    for ofn in OldFailedNotification.objects.iterator(chunk_size=SIZE):
+        # cloudevents where never caught in the logging filter
+        nfn = NewFailedNotification(message=ofn.message,  type="notification")
+
+        # status_code & response_status are both an integer field that allow null
+        # msg is a textfield and could be larger than exception max length
+        nr = NotificationResponse(failed_notification=nfn, exception=ofn.msg[:500], response_status=ofn.status_code)
```
Floris272, line 22: could also use the ofn.trace instead of msg

src/openzaak/notifications/viewsets.py
```diff
@@ -24,6 +26,21 @@
 logger = structlog.stdlib.get_logger(__name__)
 
 
+def create_failed_notification(message: dict) -> int | None:
```
stevenbal: Same question as https://github.com/maykinmedia/open-klant/pull/608#discussion_r3615093844, perhaps this could/should be part of the library instead?

src/openzaak/notifications/migrations/0005_migrate_failednotifications.py
```diff
@@ -0,0 +1,40 @@
...
+
+    NewFailedNotification.objects.bulk_create(notifications)
+    NotificationResponse.objects.bulk_create(notification_responses)
+
+    OldFailedNotification.objects.all().delete()
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('notifications_log', '0004_alter_failednotification_status_code'),
```
stevenbal, line 35: This probably also has a dependency on one of the recent migrations in `notifications_api_common`, like `notifications_api_common/migrations/0011_failednotification_notificationresponse.py`?

src/openzaak/notifications/migrations/0005_migrate_failednotifications.py
stevenbal: What's the difference between the two?

src/openzaak/notifications/migrations/0005_migrate_failednotifications.py
```diff
@@ -0,0 +1,40 @@
...
+def migrate_failed_notifications(apps, schema_editor):
+    OldFailedNotification = apps.get_model('notifications_log', 'FailedNotification')
+    NewFailedNotification = apps.get_model('notifications_api_common', 'FailedNotification')
+    NotificationResponse = apps.get_model('notifications_api_common', 'NotificationResponse')
+
+    notifications = []
+    notification_responses = []
+    for ofn in OldFailedNotification.objects.iterator(chunk_size=SIZE):
+        # cloudevents where never caught in the logging filter
+        nfn = NewFailedNotification(message=ofn.message,  type="notification")
```
stevenbal, line 18: Could you use the constant from the library here for the `notification` type?

src/openzaak/components/zaken/api/viewsets.py
```diff
@@ -2462,7 +2465,10 @@ def send_rol_notification(rol, action):
                 model=config["model"],
                 action=action,
             )
-            transaction.on_commit(lambda msg=message: send_notification.delay(msg))
+
+            pk = create_failed_notification(message)
+
+            transaction.on_commit(lambda msg=message: send_notification.delay(msg, pk))
```
stevenbal: I think some logic here is missing for other notifications on this convenience viewset (and perhaps others as well). I did a POST locally on `/zaak_registreren` with a status and rollen (resulting in a failing notif), but only a failednotification was stored for the Rol notification

Image

It does seem to have all the responses for the other notifications linked to it however:

Image

A test for this might be missing as well?

src/openzaak/conf/includes/base.py
```diff
@@ -608,6 +588,23 @@
     ),
 )
 
+LOG_NOTIFICATIONS_IN_DB = config(
+    "LOG_NOTIFICATIONS_IN_DB",
+    default=False,
```
stevenbal: I'm sort of leaning towards enabling this by default, since it only stores data on failure, I think the impact should be quite minimal and then we won't have situations where a municipality runs into issues and forgot to turn it on. Could you do the same for Open Object? We can leave the default in the library as False

stevenbal (changes requested):

src/openzaak/notifications/migrations/0005_migrate_failednotifications.py
Floris272: Yes, but the migration would break if the value ever changes or if NotificationType is removed. I couldn't load it in like the models.

src/openzaak/notifications/viewsets.py
Floris272: moved

src/openzaak/notifications/migrations/0005_migrate_failednotifications.py
Floris272: msg has the log msg for example `'Could not deliver message to https://notificaties-api.vng.cloud/api/v1/'`

trace has the stack trace
```
Traceback (most recent call last):
  File "/home/floris272/Documents/repos/open-zaak/.venv/lib/python3.12/site-packages/notifications_api_common/tasks.py", line 34, in send_notification
    response.raise_for_status()
  File "/home/floris272/Documents/repos/open-zaak/.venv/lib/python3.12/site-packages/requests/models.py", line 1028, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 403 Client Error: None for url: https://notificaties-api.vng.cloud/api/v1/notificaties
```
I could get the last line based on `\n`

src/openzaak/components/zaken/api/viewsets.py
Floris272: should be working now.

src/openzaak/conf/includes/base.py
Floris272: enabled here and in OK for now

stevenbal: Could you also do it in Open Object?

src/openzaak/notifications/migrations/0005_migrate_failednotifications.py
stevenbal: ah that's right, good point, let's keep it as is

stevenbal (approved):

PR #2475, opened 2026-07-24
:fire: [#2460] Remove external zaaktype reference support

### LINKED ISSUE #2460: Remove support for external catalogi references
Currently Open Zaak allows both foreign key relations and external relations (via services + a path) to zaaktypen, besluittypen and informatieobjecttypen. Because this functionality (to refer to external catalogi) is not used in practice and adds quite a bit of complexity to the codebase, we are going to remove this in Open Zaak 2.0.

Migrate to ForeignKey:
- [ ] Zaak.zaaktype
- [ ] Status.statustype
- [ ] Resultaat.resultaattype
- [ ] Rol.roltype
- [ ] ZaakObject.zaakobjecttype
- [ ] ZaakEigenschap.eigenschap
- [ ] EnkelvoudigInformatieObject.informatieobjecttype
- [ ] Besluit.besluittype
- [ ] Autorisatie.zaaktype / Autorisatie.besluittype / Autorisatie.informatieobjecttype

OlhaZahoruiko:
Closes #2460

**Changes**

- Remove support for external `Zaak.zaaktype` references.
- Remove support for external `Status.statustype` references.
- Replace `FkOrServiceUrlField` with regular `ForeignKey` fields.
- Remove external URL validation, serializers, filters, admin support, and related tests.

**Checklist**

- Commit hygiene

- Architectural design record


src/openzaak/components/zaken/models/zaken.py
```diff
@@ -801,38 +779,13 @@ class Status(ETagMixin, APIMixin, models.Model):
...
-            "URL-referentie naar extern STATUSTYPE (in een andere Catalogi API)."
-        ),
-    )
-    _statustype = models.ForeignKey(
+    statustype = models.ForeignKey(
         "catalogi.StatusType",
         on_delete=models.PROTECT,
         help_text="URL-referentie naar het STATUSTYPE (in de Catalogi API).",
         null=True,
         blank=True,
```
stevenbal: We will add (in a different PR) an upgrade check that will block upgrading to 2.0 if external types are still used, so we can and should make the new foreignkeys for these types required:

```suggestion
null=False,
blank=False,
```

src/openzaak/components/zaken/models/zaken.py
```diff
@@ -157,38 +157,13 @@ class Zaak(ETagMixin, AuditTrailMixin, APIMixin, ZaakIdentificatie):
...
-            "URL-referentie naar extern ZAAKTYPE (in een andere Catalogi API)."
-        ),
-    )
-    _zaaktype = models.ForeignKey(
+    zaaktype = models.ForeignKey(
         "catalogi.ZaakType",
         on_delete=models.PROTECT,
         help_text="URL-referentie naar het ZAAKTYPE (in de Catalogi API).",
         null=True,
         blank=True,
```
stevenbal: Same comment as https://github.com/open-zaak/open-zaak/pull/2475/changes#r3712782047, let's make this field required:


```suggestion
null=False,
blank=False,
```

src/openzaak/components/zaken/tests/test_auth.py
stevenbal, line 2401:
```suggestion
```

src/openzaak/utils/query.py
```diff
@@ -77,10 +77,13 @@ def get_filters(
...
-            f"_{self.loose_fk_field}" if local else f"_{self.loose_fk_field}_url"
-        )
-
+        supports_external_field = getattr(self, "supports_external_field", True)
+        if supports_external_field:
+            loose_fk_field = (
+                f"_{self.loose_fk_field}" if local else f"_{self.loose_fk_field}_url"
+            )
+        else:
+            loose_fk_field = self.loose_fk_field
```
stevenbal, line 89: This is a workaround to make sure tests pass until all of the external types are removed right? If so, could you add a TODO comment as a reminder to remove this once that is the case?

src/openzaak/components/zaken/tests/test_zaakobject_zaakobjecttype.py
```diff
@@ -192,122 +186,6 @@ def test_read_with_zaakobjecttype_external(self):
             },
         )
 
-    @requests_mock.Mocker()
-    def test_create_with_zaakobjecttype_external(self, m):
```
stevenbal, line 196: Technically not related to the two external types in this PR, but since we need to remove this when #2460 is finished anyway, I'm fine with removing these already in this PR

src/openzaak/components/zaken/tests/test_zaak_zoek.py
```diff
@@ -229,171 +228,3 @@ def test_zoek_supports_vng_query_params(self):
...
-
-        data = response.json()["results"]
-        data = sorted(data, key=lambda zaak: zaak["identificatie"])
-
-        self.assertEqual(len(data), 2)
-        self.assertEqual(data[0]["url"], f"http://testserver.com{reverse(zaak1)}")
-        self.assertEqual(data[1]["url"], f"http://testserver.com{reverse(zaak2)}")
-
-    @override_settings(ALLOWED_HOSTS=["testserver", "testserver.com"])
-    def test_zoek_zaaktype_not_in_local(self):
```
stevenbal, line 295: I think we should keep this test (otherwise we have no other tests for `zaaktype__not_in`), you could modify it to create three local zaaktypes and apply the filter on just two of them

src/openzaak/components/zaken/tests/test_zaak_verlengen.py
```diff
@@ -417,8 +417,8 @@ def test_with_empty_status(self):
             response.status_code, status.HTTP_400_BAD_REQUEST, response.data
         )
 
-        statustype = get_validation_errors(response, "status.statustype")
-        self.assertEqual(statustype["code"], "required")
+        verlenging = get_validation_errors(response, "zaak.verlenging")
+        self.assertEqual(verlenging["code"], "required")
```
stevenbal: This seems like an odd change? Statustype should still be required. When the statustype foreignkey is made required at the database level, the original error will return again. I think this test change should be reverted

src/openzaak/components/zaken/tests/test_zaak_opschorten.py
```diff
@@ -395,8 +395,8 @@ def test_with_empty_status(self):
             response.status_code, status.HTTP_400_BAD_REQUEST, response.data
         )
 
-        statustype = get_validation_errors(response, "status.statustype")
-        self.assertEqual(statustype["code"], "required")
+        opschorting = get_validation_errors(response, "zaak.opschorting")
+        self.assertEqual(opschorting["code"], "required")
```
stevenbal: Same comment as https://github.com/open-zaak/open-zaak/pull/2475/changes#r3712935450, should be reverted.

src/openzaak/components/zaken/tests/test_zaak_bijwerken.py
```diff
@@ -507,12 +507,11 @@ def test_with_empty_status(self):
...
 
         self.assertEqual(
-            response.status_code, status.HTTP_400_BAD_REQUEST, response.data
+            response.status_code,
+            status.HTTP_400_BAD_REQUEST,
+            response.data,
         )
 
-        statustype = get_validation_errors(response, "status.statustype")
-        self.assertEqual(statustype["code"], "required")
```
stevenbal, line 514: Same comment as https://github.com/open-zaak/open-zaak/pull/2475/changes#r3712935450, should be reverted.

src/openzaak/components/zaken/tests/test_zaak_afsluiten.py
```diff
@@ -441,9 +441,6 @@ def test_zaak_afsluiten_empty_status(self):
             response.status_code, status.HTTP_400_BAD_REQUEST, response.data
         )
 
-        statustype = get_validation_errors(response, "status.statustype")
-        self.assertEqual(statustype["code"], "required")
```
stevenbal, line 445: Same comment as https://github.com/open-zaak/open-zaak/pull/2475/changes#r3712935450, should be reverted.

src/openzaak/components/zaken/tests/test_validation_zaak.py
```diff
@@ -40,56 +40,6 @@ def setUpTestData(cls):
 
         ServiceFactory.create(api_root="https://example.com/", api_type=APITypes.ztc)
 
-    @override_settings(ALLOWED_HOSTS=["testserver"])
-    def test_validate_zaaktype_bad_url(self):
```
stevenbal, line 44: Could you check happens when the zaaktype URL does point to the local openzaak instance, but there is no valid zaaktype at that URL (e.g. `http://testserver/catalogi/api/v1/zaaktypen/<uuid>` but no zaaktype with that UUID exists)?

src/openzaak/components/zaken/tests/test_auth.py
```diff
@@ -2283,132 +2219,48 @@ def test_zaak_list_with_filtering(self):
...
-        url1 = reverse(zaak1)
-        url2 = reverse(zaak2)
-
-        response1 = self.client.get(url1, **ZAAK_READ_KWARGS)
-        response2 = self.client.get(url2, **ZAAK_READ_KWARGS)
-
-        self.assertEqual(response1.status_code, status.HTTP_200_OK)
-        self.assertEqual(response2.status_code, status.HTTP_403_FORBIDDEN)
-
-    def test_statussen_list(self):
```
stevenbal, line 2509: The removed tests in this file that used to test both internal and external types should probably be kept and modified to only test internal types, otherwise we no longer have auth tests for these endpoints I think

src/openzaak/components/zaken/models/zaken.py
```diff
@@ -508,9 +483,12 @@ def __init__(self, *args, **kwargs):
...
         super().__init__(*args, **kwargs)
-        self._original_values = {
-            f.attname: getattr(self, f.attname) for f in self._meta.concrete_fields
-        }
+
+        self._original_values = {}
+
+        for f in self._meta.concrete_fields:
+            if f.attname in self.__dict__:
+                self._original_values[f.attname] = self.__dict__[f.attname]
```
stevenbal, line 496: Why was this change needed?

src/openzaak/components/zaken/api/validators.py
```diff
@@ -343,18 +343,15 @@ def __call__(self, attrs: dict):
             return
 
         self.check_status_exists(zaak.deelzaken)
-        self.check_internal_status_types(zaak.deelzaken.filter(_zaaktype__isnull=False))
-        self.check_external_status_types(
-            zaak.deelzaken.filter(_zaaktype_relative_url__isnull=False)
-        )
+        self.check_internal_status_types(zaak.deelzaken)
```
stevenbal: Let's rename these, since we now only support internal types
```suggestion
        self.check_status_types(zaak.deelzaken)
```

src/openzaak/components/zaken/api/serializers/zaken.py
```diff
@@ -1020,14 +1017,7 @@ def update_deelzaken(self, qs, brondatum: date | None):
...
         """
-        self._update_deelzaken_with_internal_catalogi(
-            qs.filter(_zaaktype__isnull=False),
-            brondatum,
-        )
-        self._update_deelzaken_with_external_catalogi(
-            qs.filter(_zaaktype_relative_url__isnull=False),
-            brondatum,
-        )
+        self._update_deelzaken_with_internal_catalogi(qs, brondatum)
```
stevenbal: Let's rename these, since we now only support internal types
```suggestion
        self._update_deelzaken(qs, brondatum)
```

src/openzaak/components/zaken/api/serializers/zaken.py
```diff
@@ -827,9 +826,7 @@ class Meta:
                 "lookup_field": "uuid",
                 "max_length": 1000,
                 "min_length": 1,
-                "validators": [
-                    LooseFkResourceValidator("StatusType", settings.ZTC_API_STANDARD),
-                ],
+                "validators": [],
```
stevenbal:
```suggestion
```

src/openzaak/components/zaken/admin/zaken.py
```diff
@@ -77,12 +77,10 @@ class Meta:
...
 
-        if not cleaned_data.get("_statustype") and not cleaned_data.get(
-            "_statustype_base_url"
-        ):
+        if not cleaned_data.get("statustype"):
             raise forms.ValidationError(
                 "Je moet een statustype opgeven: "
-                "selecteer een statustype uit de catalogus of vul een externe URL in."
+                "selecteer een statustype uit de catalogus."
             )
```
stevenbal: Once statustype is made required at the model level, this custom error can be removed

src/openzaak/components/zaken/admin/zaken.py
```diff
@@ -779,12 +777,9 @@ class Meta:
...
 
-        if not cleaned_data.get("_zaaktype") and not cleaned_data.get(
-            "_zaaktype_base_url"
-        ):
+        if not cleaned_data.get("zaaktype"):
             raise forms.ValidationError(
-                "Je moet een zaaktype opgeven: "
-                "selecteer een zaaktype of vul een externe URL in."
+                "Je moet een zaaktype opgeven: selecteer een zaaktype uit de catalogus."
             )
```
stevenbal: Once zaaktype is made required at the model level, this custom error can be removed

src/openzaak/components/zaken/migrations/0051_remove_status_zaken_status__statustype_or__statustype_base_url_filled_and_more.py
```diff
@@ -0,0 +1,46 @@
+# SPDX-License-Identifier: EUPL-1.2
+# Copyright (C) 2026 Dimpact
+# Generated by Django 5.2.15 on 2026-07-24 14:48
+
+import django.db.models.deletion
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
```
stevenbal: Could you combine this migration with 0050? I'd like to group them since they are basically the same change. Easiest way would be to migrate back to 0049_alter_zaak_laatst_gemuteerd locally, remove the files and rerun makemigrations:

src/openzaak/components/zaken/tests/test_hoofdzaak_afsluiting.py
```diff
@@ -674,39 +472,16 @@ def test_queries_with_one_deelzaak_with_external_catalogi(self):
 
     def test_queries_with_many_deelzaken_with_internal_catalogi(self):
```
stevenbal: `test_queries_with_many_deelzaken` currently tests the same (except with 10 more deelzaken). Let's keep `test_queries_with_many_deelzaken` and remove `test_queries_with_many_deelzaken_with_internal_catalogi`

src/openzaak/components/zaken/tests/test_hoofdzaak_afsluiting.py
```diff
@@ -619,49 +449,17 @@ def test_queries_with_no_deelzaken(self):
             self.assertEqual(response.status_code, status.HTTP_201_CREATED)
 
     def test_queries_with_one_deelzaak_with_internal_catalogi(self):
```
stevenbal:
```suggestion
    def test_queries_with_one_deelzaak(self):
```

src/openzaak/components/zaken/tests/test_hoofdzaak_afsluiting.py
```diff
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: EUPL-1.2
```
stevenbal, line 1: Could you remove `internal` from the names of the tests in this file? Mentioning it is no longer relevant, since all types will be internal from now on

stevenbal (changes requested):
Good call to split it up into multiple PRs, I didn't realize how much code was related to this feature. Glad we're finally removing it 🎉

src/openzaak/utils/query.py
OlhaZahoruiko: Yes, that's correct. I'll add a TODO comment to make it clear that this compatibility code can be removed once external type support is fully removed

src/openzaak/components/zaken/models/zaken.py
OlhaZahoruiko: This change is needed because using getattr(self, f.attname) in __init__ forces deferred fields to be loaded. For VirtualZaak instances this results in a refresh_from_db() call and raises VirtualZaak.DoesNotExist. By only storing fields that are already present in __dict__, we avoid triggering deferred field loading while still tracking the original values

src/openzaak/components/zaken/tests/test_validation_zaak.py
OlhaZahoruiko: I checked this case. If an authorization contains a local zaaktype URL whose UUID doesn't exist, get_resources_for_paths() resolves only the existing resources and then raises a RuntimeError because the number of resolved objects doesn't match the number of input paths. This currently results in a 500 Internal Server Error for the request

src/openzaak/components/zaken/tests/test_zaak_zoek.py
Floris272: lets rename it to test_zoek_zaaktype_not_in

src/openzaak/components/zaken/tests/test_hoofdzaak_afsluiting.py
Floris272, line 36: all `ext` stuff could be removed from setupclass & setup

src/openzaak/components/zaken/models/zaken.py
```diff
@@ -157,37 +157,12 @@ class Zaak(ETagMixin, AuditTrailMixin, APIMixin, ZaakIdentificatie):
...
         help_text="URL-referentie naar het ZAAKTYPE (in de Catalogi API).",
-        null=True,
-        blank=True,
-    )
-    zaaktype = FkOrServiceUrlField(
-        fk_field="_zaaktype",
-        url_field="_zaaktype_url",
-        help_text="URL-referentie naar het ZAAKTYPE (in de Catalogi API).",
+        null=False,
+        blank=False,
```
Floris272: could be removed because False is the default for both.

src/openzaak/components/zaken/api/serializers/zaken.py
```diff
@@ -532,7 +532,6 @@ class Meta:
                 "max_length": 1000,
                 "min_length": 1,
                 "validators": [
-                    LooseFkResourceValidator("ZaakType", settings.ZTC_API_STANDARD),
                     LooseFkIsImmutableValidator(),
```
Floris272: Can be replaced with IsImmutableValidator

src/openzaak/components/zaken/models/zaken.py
Floris272: It fails because of the besluiten tests with external zaak, Virtual models are created in the ProxyMixin.

on main `besluit_zaak_id` is None which does not cause django to do a fetch, so it should be fixed when besluit.zaak is changed to local only and proxymixin is removed. 

could fix besluit.zaak in this pr or just add a TODO here.

src/openzaak/components/zaken/tests/test_validation_zaak.py
Floris272: valid_testserver_url can be removed

src/openzaak/components/zaken/tests/test_validation_zaak.py
Floris272: I think we should keep the bad_url & maybe invalid_resource tests (without the request mocker) to make sure it gives an relevant error.

src/openzaak/utils/query.py
stevenbal: I think you forgot to add it?

OlhaZahoruiko: no, it's right here :grimacing:

Image

src/openzaak/components/zaken/models/zaken.py
OlhaZahoruiko: added TODO for now

Floris272 (approved):

src/openzaak/utils/query.py
stevenbal: check, maybe github had some hiccups again :weary:

stevenbal (approved):

src/openzaak/components/zaken/openapi.yaml
```diff
@@ -14627,7 +14627,7 @@ paths:
                     type: string
                   description: Meerdere waarden kunnen gescheiden worden door komma's.
                 zaaktype:
-                  type: string
+                  type: integer
                   description: URL-referentie naar het ZAAKTYPE (in de Catalogi API).
```
stevenbal, line 14631: This also needs to stay a string (ideally with format uri)

src/openzaak/components/zaken/openapi.yaml
```diff
@@ -11280,7 +11280,7 @@ paths:
       - in: query
         name: zaaktype
         schema:
-          type: string
+          type: integer
```
stevenbal, line 11283: Should remain a string

src/openzaak/components/zaken/openapi.yaml
```diff
@@ -4949,7 +4949,7 @@ paths:
       - in: query
         name: statustype
         schema:
-          type: string
+          type: integer
```
stevenbal, line 4952: Should remain a string

stevenbal (changes requested):

src/openzaak/components/zaken/openapi.yaml
OlhaZahoruiko: https://github.com/open-zaak/open-zaak/pull/2505/changes#diff-a44139bd824a9c4ffafcea6f807e8311ae9624e5bc5fdcfa0c1d8529363c36a6L4950

src/openzaak/components/zaken/openapi.yaml
OlhaZahoruiko: https://github.com/open-zaak/open-zaak/pull/2505/changes#diff-a44139bd824a9c4ffafcea6f807e8311ae9624e5bc5fdcfa0c1d8529363c36a6L11281

PR #2482, opened 2026-07-28
Feature/2451 besluiten notifications

### LINKED ISSUE #2451: How to deal with notifications, audit trails and authorisations when moving Besluiten
It was discussed that for the deprecation period:

# Notifications

We'll sent 2 notifications:
a) Via the Besluiten channel, with Besluiten context and with URLs that point to the Besluiten API
b) Via the Zaken channel, if the Besluit is related to a zaak, but with Zaak context (so it does not need to be equal to the Besluiten channel notification) and with URLs that point to the Zaken API

# Audit trails

We'll create 2 audit entries for practicality
a) One in the Besluit audit table
b) One in the Zaak audit table

# Authorizations

TODO

Related to: Status: #2441, #2445

Floris272:
Partially closes #2451 

**Changes/Notes**
- refactors MultipleNotificationMixin to MultipleObjectsNotificationMixin
- adds MultipleChannelNotificationMixin & MultipleObjectsMultipleChannelNotificationMixin

- viewsets in deprecated apis, (besluit & besluitIO in this pr) will now have MultipleChannelNotificaitonMixin so that notification can be send on original channel (besluit) and new one (zaak). 
- besluitverwerken already needed to send notifications for multiple objects and not for mulitple channels as well.

- Note that for besluit the notif on zaak kanaal can only be send if the besluit has a zaak since that is not required yet.
- for the zaak notif, the original notif mixin tries to get the main_object (the zaak) from the data but BIO does not have the zaak in its data. In MultipleChannelNotificationMixin the zaak is injected into the data to not have to override the notif mixin functions to much.In the BIOviewset the zaak is fetched from viewset data serializer which has acces to the besluit model instance.
- Besluitverwerken already has the besluit since it will send a notif for that as well so the zaak can be fetched from there.

**Checklist**

- Commit hygiene

- Architectural design record


src/openzaak/notifications/viewsets.py
```diff
@@ -20,44 +25,267 @@
...
+
+    notification_fields: dict[str, NotificationFieldConfig]
+
+    def _iter_field_notifications(
+        self,
+        data: dict,
+        notification_fields: dict[
+            str, NotificationFieldConfig | MultipleChannelNotificationFieldConfig
+        ],
+    ):
```
stevenbal: could you add the return type for this as well?

src/openzaak/notifications/viewsets.py
```diff
@@ -20,44 +25,267 @@
...
+            if data is None:
+                return ""
+            data = data.get(field)
+        return data if isinstance(data, str) else ""
+
+    def _get_nested_main_object_url(
+        self,
+        key: str,  # format a.b.c
+        nested_main_object_resource: Type[models.Model] | dict | None,
+    ):
```
stevenbal: Could you add typehints for the arguments/return types here as well?

src/openzaak/notifications/viewsets.py
```diff
@@ -20,44 +25,267 @@
...
+                continue
+
+            for field in replace_urls_for:
+                notification_data[field] = self._replace_namespace(
+                    notification_data[field], kanaal.label
+                )
+
+            yield kanaal, notification_data
+
+    def _message(self, data, instance=None):
```
stevenbal:
```suggestion
    def _message(self, data, instance=None) -> None:
```

src/openzaak/notifications/viewsets.py
```diff
@@ -20,44 +25,267 @@
...
+
+    def _iter_kanalen(
+        self,
+        data: dict,
+        model: models.Model,
+        kanalen: list[Kanaal],
+        replace_urls_for: list[str] | None = None,
+        main_resource_keys: dict[str, str] | None = None,
+        nested_main_object_resource: Type[models.Model] | dict | None = None,
+    ):
```
stevenbal: could you add the return type for this as well?

src/openzaak/notifications/viewsets.py
```diff
@@ -20,44 +25,267 @@
...
+        for field in key.split("."):
+            if data is None:
+                return ""
+            data = data.get(field)
+        return data if isinstance(data, str) else ""
+
+    def _get_nested_main_object_url(
+        self,
+        key: str,  # format a.b.c
+        nested_main_object_resource: Type[models.Model] | dict | None,
```
stevenbal: In what situation would this be `None`? Also `nested_main_object_resource` will be an instance of a model, right? In that case I think the typehint should be `models.Model`, otherwise it means you're just passing the `Zaak` class for instance. Same applies to other uses of `Type[models.Model]` in this file most likely

src/openzaak/components/besluiten/tests/test_notifications_send.py
```diff
@@ -60,26 +67,103 @@ def test_send_notif_create_besluit(self, mock_notif):
...
+                            "besluittype.catalogus": f"http://testserver{reverse(besluittype.catalogus)}",
+                        },
+                    },
+                    None,
+                ),
+            ]
         )
 
     @tag("convenience-endpoints")
     def test_send_notif_verwerk_besluit(self, mock_notif):
```
stevenbal: Could you also add this test without specifying a zaak?

src/openzaak/components/besluiten/tests/test_notifications_send.py
```diff
@@ -212,6 +367,65 @@ def test_send_notif_delete_resultaat(self, mock_notif):
             None,
         )
 
+    def test_send_notif_delete_besluitinformatieobject_with_zaak(self, mock_notif):
```
stevenbal, line 1025: Could you also add the two tests for BesluitInformatieObject create (with and without zaak)?

src/openzaak/components/besluiten/tests/test_notifications_send.py
stevenbal: Could you also add the tests for the other operations (PUT/PATCH/DELETE) for besluit (with and without zaak)?

src/openzaak/components/besluiten/tests/test_notifications_send.py
```diff
@@ -60,26 +67,103 @@ def test_send_notif_create_besluit(self, mock_notif):
...
+                        "kanaal": "besluiten",
+                        "hoofdObject": data["url"],
+                        "resource": "besluit",
+                        "resourceUrl": data["url"],
+                        "actie": "create",
+                        "aanmaakdatum": "2018-09-07T00:00:00Z",
+                        "kenmerken": {
+                            "verantwoordelijkeOrganisatie": "517439943",
+                            "besluittype": f"http://testserver{besluittype_url}",
+                            "besluittype.catalogus": f"http://testserver{reverse(besluittype.catalogus)}",
```
stevenbal: Would it be possible to explicitly specify the `catalogi` namespace for these already? Or will that have to wait for #2407?

src/openzaak/components/zaken/tests/besluiten/test_notifications_send.py
```diff
@@ -125,16 +210,35 @@ def test_send_notif_verwerk_besluit(self, mock_notif):
...
+
+        self.assertCountEqual(
+            [bio["url"] for bio in data["besluitinformatieobjecten"]],
+            [
+                f"http://testserver{reverse(bio_1, namespace='zaken')}",
+                f"http://testserver{reverse(bio_2, namespace='zaken')}",
+            ],
+        )
+
+        self.assertEqual(mock_notif.call_count, 6)
```
stevenbal: This is kind of a lot, perhaps we should also introduce some feature flag to disable the "old" notifications, that way municipalities can disable them if they're fully migrated to the new situation

src/openzaak/components/besluiten/tests/test_notifications_send.py
```diff
@@ -11,13 +11,15 @@
 from rest_framework import status
```
stevenbal, line 11: Would there be no way to define these tests once and reuse the testcase classes in the zaken component? I think the only difference is the namespace of the URL you do API calls on, so maybe that could be defined as a classvar?

src/openzaak/components/zaken/tests/besluiten/test_notifications_send.py
```diff
@@ -12,6 +12,7 @@
 from rest_framework.test import APITestCase
```
stevenbal, line 12: Could you add the same missing tests mentioned in my comments in `src/openzaak/components/besluiten/tests/test_notifications_send.py` (or reuse the testcases like mentioned in https://github.com/open-zaak/open-zaak/pull/2482/changes#r3682895948)?

src/openzaak/notifications/viewsets.py
```diff
@@ -20,44 +25,267 @@
...
+            return main_resource_keys.get(kanaal.label)
+
+        return kanaal.main_resource._meta.model_name
+
+    def _get_nested_main_object_url_from_instance(self, key, instance):
+        """Returns the url of an nested FK field"""
+        obj = instance
+        for field in key.split("."):
+            obj = getattr(obj, field, None)
+        return obj.get_absolute_api_url(request=self.request) if obj else ""
```
stevenbal: Shouldn't this raise an exception instead if obj is None? Otherwise we're silently passing empty urls, right?

src/openzaak/notifications/viewsets.py
```diff
@@ -20,44 +25,267 @@
...
+
+    def _get_nested_main_object_url_from_dict(self, key, data):
+        """Returns a nested url field from a dict"""
+        for field in key.split("."):
+            if data is None:
+                return ""
+            data = data.get(field)
+        return data if isinstance(data, str) else ""
```
stevenbal: Same remark as above, isn't it problematic to return empty strings here?

src/openzaak/notifications/viewsets.py
stevenbal: Can't it be called `main_object_resource` instead? Why is it nested?

src/openzaak/notifications/viewsets.py
```diff
@@ -20,44 +25,267 @@
...
+        if the main_object_url is not part of the notification data it is fetched from an
+        instance or dict and added to the notification data
+        """
+        if "." not in key:
+            # original flow
+            url = data.get(key)
+        else:
+            url = self._get_nested_main_object_url(key, nested_main_object_resource)
+            final_key = key.split(".")[-1]
+            data[final_key] = url
```
stevenbal: It seems a bit iffy to have a method called `_main_object_url_exists` that also mutates one of the arguments you pass into it. Can't this be done differently?

src/openzaak/notifications/viewsets.py
```diff
@@ -20,44 +25,267 @@
...
+
+            yield kanaal, notification_data
+
+    def _message(self, data, instance=None):
+        for kanaal, notification_data in self._iter_kanalen(
+            data,
+            self.get_queryset().model,
+            self.notifications_kanalen,
+            getattr(self, "notifications_replace_urls_for", None),
+            getattr(self, "notifications_main_resource_keys", None),
```
stevenbal: I think you can set the defaults for both of these to `None` at the class level, that way you don't have to rely on `getattr` and have the benefits of typehints

src/openzaak/notifications/viewsets.py
```diff
@@ -20,44 +25,267 @@
...
+class NotificationFieldConfig(TypedDict):
+    notifications_kanaal: Kanaal
+    model: Type[models.Model]
+    action: str | None
+
+
+class MultipleChannelNotificationFieldConfig(TypedDict):
+    notifications_kanalen: list[Kanaal]
+    model: Type[models.Model]
+    action: NotRequired[str]
```
stevenbal: Won't notifications be missing an `action` if it's `None` or not defined? I see that there are type errors on the ZaakRegistrerenViewSet for `notification_fields` currently:

Image

stevenbal (changes requested):

src/openzaak/components/besluiten/tests/test_notifications_send.py
Floris272: Added NAMESPACE to all api tests in besluiten, besluiten api tests in zaken/tests/besluiten now subclass the cases from besluiten and set NAMESPACE to zaken

src/openzaak/components/besluiten/tests/test_notifications_send.py
Floris272: added

src/openzaak/notifications/viewsets.py
Floris272: should be fixed now

src/openzaak/components/zaken/tests/besluiten/test_notifications_send.py
Floris272: added SEND_NOTIFICATIONS_ON_DEPRECATED_CHANNELS

src/openzaak/notifications/viewsets.py
Floris272: refactored it by moving the mutation to _iter_kanalen also added comments

src/openzaak/notifications/viewsets.py
Floris272: it will be an instance of a model or a dict, fixed the type hint. It can be None because it is only really needed if nested_main_resource_keys are set

src/openzaak/notifications/viewsets.py
Floris272: change it to return None, if besluit does not have a zaak this will be None and _iter_kanalen will continue to the next channel

src/openzaak/notifications/viewsets.py
Floris272: Changed it to `main_resource_object`

This is for besluitInformatieObject:

Bio serializer has `notifications_main_resource_keys = {"zaken": "besluit.zaak"}`
which means that the main object  is located at <root>.besluit.zaak (root is the bio in this example)

src/openzaak/notifications/viewsets.py
stevenbal: Still seeing some typing errors here:

Type "dict[str, dict[str, Kanaal | type[Zaak]] | dict[str, Kanaal | type[Status]] | dict[str, Kanaal | type[Rol]] | dict[str, Kanaal | type[ZaakInformatieObject]] | dict[str, Kanaal | type[ZaakObject]]]" is not assignable to declared type "dict[str, NotificationFieldConfig]"
  Type "type[Zaak]" is not assignable to type "Model"
  Type "type[Status]" is not assignable to type "Model"
  Type "type[Rol]" is not assignable to type "Model"
  Type "type[ZaakInformatieObject]" is not assignable to type "Model"
  Type "type[ZaakObject]" is not assignable to type "Model"

I think `NotificationFieldConfig.model` should be `type[Model]` and `action_override` should be marked as NotRequired

stevenbal (approved):

PR #2499, opened 2026-08-04
Feature/2453 documenten api to 1.5.0

### LINKED ISSUE #2453: Upgrade Documenten API to version 1.5.0 of the standard
### Thema / Theme

Documenten API

We should upgrade our implementation of the Documenten API (currently 1.4.2) to version 1.5.0 of the standard from VNG (https://vng-realisatie.github.io/gemma-zaken/standaard/documenten/release_notes.html). I took a look at the changelog for 1.4.3, but this version seems to fix a mistake that didn't exist in Open Zaak, so I think we can just upgrade to 1.5.0 directly.

Changes to implement:
- [ ] Add property `inhoudIsVervallen` to enkelvoudiginformatieobject resource. It should result in the following API spec:

```yaml
inhoudIsVervallen:
          description: >-
            Geeft aan of de inhoud van het INFORMATIEOBJECT al dan niet
            vervallen, dus niet langer geldig is.

            * `true` De inhoud van het INFORMATIEOBJECT is vervallen.

            * `false` De inhoud van het INFORMATIEOBJECT is niet vervallen.
          title: inhoud is vervallen
          type: boolean
          nullable: true
```
- [ ] The changelog mentions that the following validation: "PUT/PATCH should only be allowed for `/enkelvoudiginformatieobjecten/<uuid>` status is not `definitief`" should be removed. I'm not sure if this validation was implemented in Open Zaak in the first place, so it should first be checked if this validation exists, if it does, we should remove it.
- [ ] Increment the `DOCUMENTEN_API_VERSION` to 1.5.0 and regenerate the open API spec with `bin/generate_schema_for_component.sh documenten`

_No response_

_No response_

abegailsieffers:
Closes #2453 

**Changes**

 - [x] Add property inhoudIsVervallen to enkelvoudiginformatieobject resource.
 - [x] Checked if the niet definitief validation exists. I if exists, remove it. 
 - [x] Increment the DOCUMENTEN_API_VERSION to 1.5.0 and regenerate the open API spec 

**Checklist**

- Commit hygiene

- Architectural design record


src/openzaak/components/documenten/models.py
```diff
@@ -213,6 +213,18 @@ class InformatieObject(models.Model):
...
+    inhoud_is_vervallen = models.BooleanField(
+        _("inhoud is vervallen"),
+        default=None,
+        blank=True,
+        null=True,
+        help_text=_(
+            "Geeft aan of de inhoud van het INFORMATIEOBJECT al dan niet"
+            "vervallen, dus niet langer geldig is."
+            "`true` De inhoud van het INFORMATIEOBJECT is vervallen."
+            "`false` De inhoud van het INFORMATIEOBJECT is niet vervallen."
```
stevenbal: Let's make sure this is formatted a bit better, to make it more readable in redoc:
```suggestion
            "Geeft aan of de inhoud van het INFORMATIEOBJECT al dan niet "
            "vervallen, dus niet langer geldig is.\n"
            "* `true` De inhoud van het INFORMATIEOBJECT is vervallen.\n"
            "* `false` De inhoud van het INFORMATIEOBJECT is niet vervallen."
```

Results in:
Image

src/openzaak/components/documenten/import_utils.py
```diff
@@ -200,6 +202,10 @@ def trefwoorden(self) -> list[str]:
 
         return trefwoorden.split(",")
 
+    @property
+    def inhoud_is_vervallen(self) -> bool | None:
+        return self._inhoud_is_vervallen in ("True", "true")
```
stevenbal: I think this can never return `None` currently, just `True` or `False`. 

Let's add an explicit check for a `None`-like value first, to make sure it's still possible to specify this attribute as `None`:


```suggestion
    def inhoud_is_vervallen(self) -> bool | None:
        value = self._inhoud_is_vervallen.lower() 
        if value in ("none", "null", ""):
            return None
        return value == "true"
```

Could you also make sure the case with the empty / null value is tested in one of the bulk import tests?

src/openzaak/components/documenten/import_utils.py
```diff
@@ -43,6 +43,7 @@ class DocumentRow:
     _informatieobjecttype: str
     _zaak_uuid: str
     _trefwoorden: str
+    _inhoud_is_vervallen: bool | None
```
stevenbal: This is the raw value that is read from the csv, so this will always be a string I think
```suggestion
    _inhoud_is_vervallen: str
```

stevenbal (commented):
Looks good overall! 👍 

Could you also modify your commit messages to use [gitmoji](https://gitmoji.dev/) (https://github.com/carloscuesta/gitmoji)? That makes it easy to identify the type of changes that a commit makes, and it also makes it easier to write atomic commits (e.g. if you feel like you have to use two emojis, it probably isn't atomic commit and should be split up). 

Minor thing for the next PR, doesn't have to be applied for this one: when naming the branch, could you name the branch like `feature/<issue_number>-<description>` (so in this case it would be something like `feature/2453-documenten-api-to-1.5.0`)? That makes them a bit easier to identify :)

src/openzaak/components/documenten/import_utils.py
abegailsieffers: I added additional assertion for `test_import_documents.py` and `test_import_document_row.py`

src/openzaak/components/documenten/tests/import/test_import_documents.py
```diff
@@ -206,6 +213,7 @@ def test_total_smaller_than_batch_size(self):
         self.assertTrue(
             all((row[-1] == ImportRowResultChoices.imported.label) for row in rows[1:])
         )
+        self.assertTrue(rows[1][-1], None)
```
stevenbal: Could you also use the `inhoud_is_vervallen_idx` here, similar to what you did in `test_simple_import`? That makes it a bit more easy to understand what is being checked here

stevenbal (changes requested):
Small remark remaining. I'm also checking with VNG to ask what the meaning of `null` values for this field are, so I'll wait with merging this until I have a response from them

src/openzaak/components/documenten/tests/import/test_import_documents.py
abegailsieffers: Sure, I'll do change it! 

I noticed there are separate tests for Azure, S3, and the import folder. I only updated the tests in the import folder. Would it be better to make the same changes to the Azure and S3 import test cases as well, to keep them consistent?

stevenbal: I think it's fine to just update the regular tests an not the azure/s3 tests. Those mostly exist to verify that the handling of files is done correctly, the rest of the import process is identical

abegailsieffers:
> Small remark remaining. I'm also checking with VNG to ask what the meaning of `null` values for this field are, so I'll wait with merging this until I have a response from them

I finished changing the remarks. 
Thanks for checking with VNG. 
I'll wait for your update before merging this.

src/openzaak/conf/includes/api.py
```diff
@@ -7,7 +7,7 @@
 AUTORISATIES_API_VERSION = "1.0.0"
 BESLUITEN_API_VERSION = "1.1.0"
 CATALOGI_API_VERSION = "1.3.1"
-DOCUMENTEN_API_VERSION = "1.4.2"
+DOCUMENTEN_API_VERSION = "1.5.0"
```
stevenbal, line 10: I forgot about this, could you also increment the API version in `docs/api/index.rst` for Documenten API?

src/openzaak/components/documenten/models.py
```diff
@@ -213,6 +213,18 @@ class InformatieObject(models.Model):
...
+    inhoud_is_vervallen = models.BooleanField(
+        _("inhoud is vervallen"),
+        default=None,
+        blank=True,
+        null=True,
+        help_text=_(
+            "Geeft aan of de inhoud van het INFORMATIEOBJECT al dan niet "
+            "vervallen, dus niet langer geldig is.\n"
+            "* `true` De inhoud van het INFORMATIEOBJECT is vervallen.\n"
+            "* `false` De inhoud van het INFORMATIEOBJECT is niet vervallen."
```
stevenbal: I got a response from VNG, this help text will be added in 1.8.0 (which is yet to be released), but I'd rather add it now to make it clear what this value means
```suggestion
            "* `true` De inhoud van het INFORMATIEOBJECT is vervallen.\n"
            "* `false` De inhoud van het INFORMATIEOBJECT is niet vervallen.\n"
            "* `null` Het is nog niet bekend of bepaald of het INFORMATIEOBJECT wel of niet vervallen is."
```

stevenbal (changes requested):

src/openzaak/components/documenten/models.py
abegailsieffers: Added!

stevenbal (approved):

PR #2505, opened 2026-08-10
:fire: [#2460] use FK instead of external URL fields for eigenschap a…

OlhaZahoruiko:
…nd zaakobjecttype

Closes #2460

**Changes**
- Replace external URL fields with `ForeignKey` for `eigenschap` and `zaakobjecttype`
- Remove related URL fields and validators
- Update serializers, admin, and viewsets accordingly
- Refactor tests and remove external URL cases

**Checklist**

Check off the items that are completed or not relevant.

- Experimental features/changes

  - [ ] Any experimental features added in this PR are backwards compatible
  - [ ] Any experimental features added in this PR are documented in the `docs/api/experimental.rst` page

- Commit hygiene

  - [ ] Commit messages refer to the relevant Github issue
  - [ ] Commit messages explain the "why" of change, not the how

- Architectural design record

  - [ ] If a design decision was made that changes functionality, create an architectural design record for it [here](https://github.com/open-zaak/open-zaak/wiki/Architectural-Decision-Records-(ADR))

PR #2522, opened 2026-08-18
:fire: [#2460] use FK instead of external URL for besluittype and inf…

### LINKED ISSUE #2460: Remove support for external catalogi references

OlhaZahoruiko:
…ormatieobjecttype

Closes #2460 

**Changes**

**Besluit.besluittype** and **EnkelvoudigInformatieObject.informatieobjecttype** now use local foreign keys instead of external URL fields    

**Checklist**

- Commit hygiene

- Architectural design record


src/openzaak/components/documenten/loaders.py
```diff
@@ -32,14 +36,16 @@ def load(self, url: str, model: ModelBase):
...
+        )
         self.add_missing_props(model, model_instance, data)
         return model_instance
 
     def resolve_io_type(self, url: str):
         try:
             return get_resource_for_path(urlparse(url).path)
-        except InformatieObjectType.DoesNotExist:
+        except ObjectDoesNotExist:
             return super().load(url, model=InformatieObjectType)
```
Floris272, line 49: why was this needed?

src/openzaak/utils/query.py
```diff
@@ -78,15 +78,7 @@ def get_filters(
...
-        # TODO (Open Zaak 2.0): Remove this once all external type support has been
-        # removed. At that point every loose FK field will be a regular FK.
-        supports_external_field = getattr(self, "supports_external_field", True)
-        if supports_external_field:
-            loose_fk_field = (
-                f"_{self.loose_fk_field}" if local else f"_{self.loose_fk_field}_url"
-            )
-        else:
-            loose_fk_field = self.loose_fk_field
+        loose_fk_field = self.loose_fk_field
```
Floris272:
```suggestion
        fk_field = self.fk_field
```

Floris272 (changes requested):
the zaak model temp fix  https://github.com/open-zaak/open-zaak/pull/2475#discussion_r3794998214 could be reverted i think.

src/openzaak/components/documenten/loaders.py
OlhaZahoruiko: This was needed because EnkelvoudigInformatieObject has ondertekening and integriteit as nested API objects, while they are stored as flattened fields on the model. The regular get_model_instance() doesn't handle that mapping, so those values would be lost when loading an EnkelvoudigInformatieObject

OlhaZahoruiko:
> the zaak model temp fix [#2475 (comment)](https://github.com/open-zaak/open-zaak/pull/2475#discussion_r3794998214) could be reverted i think.

The temp fix is still needed because Besluit.zaak still uses FkOrServiceUrlField/VirtualZaak. This branch only converts the catalogi references to local FKs, so I’ll leave the temp fix for now

src/openzaak/components/documenten/api/serializers.py
```diff
@@ -312,18 +309,6 @@ class EnkelvoudigInformatieObjectSerializer(serializers.HyperlinkedModelSerializ
             "Uitdrukking van mate van volledigheid en onbeschadigd zijn van digitaal bestand."
         ),
     )
-    informatieobjecttype = FKOrServiceUrlField(
-        lookup_field="uuid",
-        max_length=200,
```
stevenbal, line 317: We have to keep the max_length for this at 200 unfortunately, the VNG spec has this as well, you can do this via `extra_kwargs` in the Meta class:

https://vng-realisatie.github.io/gemma-zaken/standaard/documenten/redoc-1.6.0.html#tag/enkelvoudiginformatieobjecten/operation/enkelvoudiginformatieobject_create
Image

src/openzaak/components/documenten/loaders.py
stevenbal: Do you have an example where this fails if we keep `InformatieObjectType.DoesNotExist`? This method should only resolve the informatieobjecttype, I don't think it should do anything with EnkelvoudigInformatieObject fields?

stevenbal (changes requested):
Same comment as https://github.com/open-zaak/open-zaak/pull/2505#pullrequestreview-4990429235, let's rebase on `dev/2.0` (but let's start with the first PR in the stack)

src/openzaak/components/documenten/loaders.py
OlhaZahoruiko: Yes, get_eio_response() in tests/utils/mocks.py already does this by default, it sets informatieobjecttype to https://external.catalogus.nl/api/v1/informatieobjecttypen/<uuid>, which has no /catalogi prefix, so it never matches a local URL pattern. get_resource_for_path() then raises a bare ObjectDoesNotExist("URL did not resolve") (from resolve_path()'s Resolver404 handling), not InformatieObjectType.DoesNotExist , so except InformatieObjectType.DoesNotExist wouldn't catch it

src/openzaak/components/besluiten/openapi.yaml
```diff
@@ -264,7 +264,7 @@ paths:
         name: besluittype
         schema:
           type: string
-        description: URL-referentie naar het BESLUITTYPE (in de Catalogi API).
+          format: uri
```
stevenbal, line 267: Guess this will be fixed with the new commonground-api-common release?

stevenbal (approved):

src/openzaak/components/besluiten/openapi.yaml
OlhaZahoruiko: yes

Issue #2322, opened 2026-02-06
(WIP) Prepare release 2.0.0


stevenbal:
Placeholder to keep track of what needs to be removed/changed in this major release
- [ ] Rename `LOG_REQUESTS` to `LOG_OUTGOING_REQUESTS (change in OAF? or override locally? https://github.com/maykinmedia/open-api-framework/blob/f54907d65adfaa5dd6fd2bdaa634bbacb8b457d0/open_api_framework/conf/base.py#L552)
- [ ] Use `CELERY_BROKER_URL` envvar for `CELERY_BROKER_URL` setting (currently uses `CELERY_RESULT_BACKEND` as per OAF https://github.com/maykinmedia/open-api-framework/blob/a9eb1d76f22c6b3fddf4969cd30bcf7783cd1f1a/open_api_framework/conf/base.py#L1285)
- [ ] Remove deprecated views for OpenAPI specs (#2189)
- [ ] Squash migrations (?)
- [ ] See if any upgrade check constraints need to be added
- [ ] Remove `django.contrib.sites`
- [ ] Remove queryparams:
  - [ ] betrokkeneIdentificatie__vestiging__kvkNummer
  - [ ] rol__betrokkeneIdentificatie__vestiging__kvkNummer\
- [ ] Remove Rol.betrokkeneType `vestiging`
- [ ] Add a datamigration to migrate relevanteAndereZaken to gerelateerdeZaken (consider bidirectionality)
- [ ] Remove relevanteAndereZaken
- [ ] Mark zaakcontactmomenten and zaakverzoeken as deprecated?
- [ ] Rename documentnummer_reserveren -> reserveer_documentnummer and mark old as deprecated (add redirect)
- [ ] rename zaaknummer -> identificatie in reserveer endpoint?
- [ ] JWT iat in future without nbf -> start raising 403?
- [ ] TODO check more deprecations and `XXX: ...` comments
- [ ] monkeypatch_drf_camel_case? (more related to v2 of api specs)

Release checklist
- [ ] Resolve release blockers
  - [ ] ...
- [ ] Upgrade `open-api-framework` to latest version
- [ ] Check security tab and upgrade packages to fix vulnerabilities
- [ ] Check translations
- [ ] Bump API version number (if applicable)
  - [ ] Version bump
  - [ ] Regenerate API spec
  - [ ] Update READMEs with release dates + links
- [ ] Bump version number with `bin/bump-my-version.sh bump <major|minor|patch>`
[...]

PR #2527, opened 2026-08-26
:sparkles:  Feature/2417 zaak inzage retrieve endpoint

### LINKED ISSUE #2417: create zaak_inzage endpoint
- [ ] Add `/zaak_inzage`/<uuid> (only allow a GET on this detail operation)
- [ ] Related objects should be nested under zaak & zaaktype
- [ ] Related objects should have an url reference to their type
- [ ] See how existing expand serializers can be reused (ideally we can make use of those and define this view as one with a "predefined" expand configuration)

```
{
  zaaktype: {
    roltypen: []
    zaaktypeinformatieobjecttypen: []
    ...
  }
  rollen: [{roltype: <url>}]
  zaakinformatieobjecten: []
  ...
}
```

objects that should be included:
  - [ ] zaak
    - [ ] zaakeigenschappen
    - [ ] besluiten
    - [ ] resultaat
    - [ ] rollen
    - [ ] statussen
      - [ ] substatussen
    - [ ] zaakcontactmomenten
    - [ ] zaakinformatieobjecten
    - [ ] zaakobjecten
    - [ ] zaakverzoeken
    - [ ] zaaknotities
    - [ ] zaaktype
      - [ ] eigenschappen
      - [ ] resultaattypen
      - [ ] roltypen
      - [ ] statustypen
      - [ ] zaakobjecttypen
      - [ ] zaakinformatieobjecttypen
- [ ] Add a performance test for the zaak-inzage endpoints (make sure the zaak has a representative number of related resources)

abegailsieffers:
Closes #2417

**Changes**
- Create zaakinzage GET endpoint
- Add zaakinzage endpoint to openapi spec
- Create zaak inzage performance test with a zaak with 1,3,5 relationships in each collection. 

**Checklist**

- Commit hygiene

- Architectural design record


src/openzaak/components/zaken/api/serializers/zaken.py
```diff
@@ -2014,3 +2027,91 @@ def update(self, instance, validated_data):
...
+            "besluiten",
+            "statussen",
+            "zaakcontactmomenten",
+            "zaakverzoeken",
+            "zaaknotities",
+        )
+
+
+class ZaakInzageSerializer(serializers.Serializer):
+    zaak = ZaakInzageZaakSerializer(source="*", read_only=True)
```
danielmursa-dev: Could you check with @Floris272 if that's necessary? 
I don't understand why we have to group everything into a single field (`zaak`) in the DetailView, unless is really required.

Thanks

src/openzaak/components/zaken/api/viewsets.py
```diff
@@ -2565,3 +2566,50 @@ def _create_audit_logs(self, response, serializer, zaak_version_before_edit):
...
+    )
+)
+class ZaakInzageViewSet(
+    CacheQuerysetMixin,
+    viewsets.ReadOnlyModelViewSet,
+):
+    serializer_class = ZaakInzageSerializer
+    lookup_field = "uuid"
+    permission_classes = (ZaakAuthRequired,)
+    required_scopes = {"retrieve": SCOPE_ZAKEN_ALLES_LEZEN}
```
danielmursa-dev: Can you discuss also with @Floris272 what happen when you don't have the permission for some resources? 
Example you have the permissions for this `zaak`, but not for the relative `zaaktype`.

This is the same security issue with the `_expand` permissions?
https://github.com/open-zaak/security-issues/issues/10

src/openzaak/components/zaken/api/serializers/zaken.py
```diff
@@ -2014,3 +2027,91 @@ def update(self, instance, validated_data):
...
+    statustypen = StatusTypeSerializer(many=True, read_only=True)
+    zaakobjecttypen = ZaakObjectTypeSerializer(
+        source="zaakobjecttype_set", many=True, read_only=True
+    )
+    informatieobjecttypen = ZaakTypeInformatieObjectTypeSerializer(
+        source="zaaktypeinformatieobjecttype_set", many=True, read_only=True
+    )
+
+    class Meta(ZaakTypeSerializer.Meta):
+        fields = ZaakTypeSerializer.Meta.fields + ("informatieobjecttypen",)
```
danielmursa-dev: I think `ZaakTypeSerializer` has arleady `informatieobjecttypen` in Meta.fields no?

src/openzaak/components/zaken/tests/test_zaak_inzage.py
```diff
@@ -0,0 +1,768 @@
...
+
+    def test_get_complete_zaak_inzage(self):
+        zaak = self.zaak
+        zaaktype = self.zaaktype
+
+        # Prepare zaaktype related data
+        eigenschaptype = EigenschapFactory(zaaktype=zaaktype)
+        resultaattype = ResultaatTypeFactory(zaaktype=zaaktype)
+        resultaattype.brondatum_archiefprocedure_afleidingswijze = "afgehandeld"
+        resultaattype.save()
```
danielmursa-dev: Is there a specific reason why you split this?
```suggestion
        resultaattype = ResultaatTypeFactory(zaaktype=zaaktype, brondatum_archiefprocedure_afleidingswijze=Afleidingswijze.choices.afgehandeld
        )
```

danielmursa-dev (commented):
Looks good 👍

src/openzaak/components/zaken/api/serializers/zaken.py
abegailsieffers: I checked it! and you are right. removed the additional field.

src/openzaak/components/zaken/tests/test_zaak_inzage.py
abegailsieffers: I think because it was not generating properly, but I will try your suggestion. 
Thank you.

src/openzaak/components/zaken/api/serializers/zaken.py
Floris272: I wrongly added zaak in the issue, lets remove this one and rename ZaakInzageZaakSerializer to ZaakInzageSerializer

src/openzaak/components/zaken/api/viewsets.py
Floris272: Lets add required_scopes = {"retrieve": SCOPE_ZAKEN_ALLES_LEZEN & SCOPE_CATALOGI_READ & SCOPE_BESLUITEN_ALLES_LEZEN} for now, I'll discuss with Steven if the endpoint should still work if the application does not have catalogi perms (by not returning zaaktype).

@abegailsieffers Could you also add auth tests for correct & incorrect scopes? (see ZaakRegistrerenAuthTests or other convenience endpoint tests for examples)

src/openzaak/components/zaken/api/serializers/zaken.py
```diff
@@ -2014,3 +2027,91 @@ def update(self, instance, validated_data):
...
+    status = ZaakInzageStatusSerializer(
+        source="current_status",
+        read_only=True,
+        allow_null=True,
+    )
+
+    eigenschappen = ZaakEigenschapSerializer(
+        source="zaakeigenschap_set", many=True, read_only=True
+    )
+    besluiten = ZaakBesluitSerializer(
```
Floris272: because besluit has now been "moved" to the zaken api we can use those directly instead of zaakbesluit.

Floris272 (changes requested):

src/openzaak/components/zaken/api/serializers/zaken.py
abegailsieffers: Sure! I'll work on it!

src/openzaak/components/zaken/api/serializers/zaken.py
abegailsieffers: Changed! Thank you.

abegailsieffers:
@Floris272  I created a performance test for zaak inzage. Can you please take a look? Thank you.

src/openzaak/components/zaken/tests/test_zaak_inzage.py
```diff
@@ -0,0 +1,896 @@
...
+)
+from .utils import ZAAK_READ_KWARGS
+
+
+class ZaakInzageAuthTests(JWTAuthMixin, APITestCase):
+    max_vertrouwelijkheidaanduiding = VertrouwelijkheidsAanduiding.zeer_geheim
+
+    @classmethod
+    def setUpClass(cls):
+        APITestCase.setUpClass()
```
danielmursa-dev: You can user super here, I think
```suggestion
        super().setUpClass()
```

src/openzaak/components/zaken/tests/test_zaak_inzage.py
```diff
@@ -0,0 +1,896 @@
...
+
+        JWTSecret.objects.get_or_create(
+            identifier=cls.client_id, defaults={"secret": cls.secret}
+        )
+        cls.applicatie = Applicatie.objects.create(
+            client_ids=[cls.client_id],
+            label="for test",
+            heeft_alle_autorisaties=False,
+        )
```
danielmursa-dev: You can see how it was done here the auth tests: https://github.com/open-zaak/open-zaak/blob/70905cd53529a141d1e6d0caa2ceb07f98cd8189/src/openzaak/components/zaken/tests/test_auth.py#L83 
Instead, to create your own token 👍 
And I think the `JWTAuthMixin` already has this logic here

abegailsieffers: I previous based the previous implementation with the ZaakRegistrerenAuthTests,
But what you suggested is probably better so I changed the implementation and updated the PR, thanks for the catch.
@danielmursa-dev  Can you please check it out? Thanks.

performance_test/test_zaak_inzage.py
```diff
@@ -0,0 +1,74 @@
...
+    url = BASE_URL / f"zaak_inzage/{zaak_id}"
+
+    def make_request():
+        return session.get(url)
+
+    result = benchmark(make_request)
+
+    assert result.status_code == 200
+    data = result.json()
+    assert_relationships(data, relation_count)
```
Floris272, line 72: could you check if there is a performance difference if the assert_relationships is removed?

performance_test/test_zaken_list.py
```diff
@@ -17,7 +17,7 @@ def make_request():
 
     assert result.status_code == 200
     data = result.json()
-    assert data["count"] == 3500
+    assert data["count"] == 3512
```
Floris272, line 20: I get why the zaak inzage data is created separetly but i would like to keep all the counts in the other tests the same. So we could either:
- use some of the exisiting zaken & eios in the zaakinzage tests and maybe add some of the missing data into the main generate data command
- separate convenience endpoint performance tests into a new CI step.

@stevenbal what do you think?

stevenbal: The issue with option 1 is that currently the generate data command generates random UUIDs, while the changes Abegail made use fixed UUIDs to make sure those can be used in tests (otherwise you can't distinguish between the number of related resources like it's done now). If we only go for a fixed number of related resources (e.g. just 5, instead of 1, 3 and 5), we could add the zaakinzage data option to the main command and generate it for every zaak, but that would probably make the command slower (not sure by how much), though it does make the data more realistic.

@abegailsieffers could you compare how much time `generate_data` takes to run in the current state of this PR vs if we generate 5 for all of the related resources for all zaken? If it still completes in a reasonable amount of time, I'd like to go for that. Otherwise, I think going for the second option is probably better, we could run it in parallel with the current performance test job.

src/openzaak/management/commands/generate_data.py
```diff
@@ -827,3 +854,196 @@ def generate_non_superuser_credentials_many_authorized_types(self):
                 str(SCOPE_AUTORISATIES_LEZEN),
             ],
         )
+
+    def generate_zaak_inzage_data(self):
+        """
+        Generate zaak inzage data with 1,3,5 data generated for each relationship
+        - test_inzage_1 = 1 zaak + 1 deelzaak
+        - test_inzage_3 = 1 zaak + 3 deelzaak
+        - test_inzage_5 = 1 zaak + 5 deelzaak
```
stevenbal, line 863: I think it's nice that you added the different options for number of related resources, though I'm not sure how realistic it is for a Zaak to have 5 deelzaken, that seems a little much. On the other hand, I feel like 5 roltypen, statustypen, resultaattypen, etc. might be slightly underestimating the real counts.

I think to make this performance test more realistic, the counts should be closer to what they are in reality (which will probably mean specific numbers for each resource). To avoid making numbers up myself though, I want to check this with a municipality to see what the average counts in their Open Zaak instances are. For this PR I think it's okay to keep it as is, until we have more information. I'll create a separate issue for it.

src/openzaak/management/commands/generate_data.py
```diff
@@ -294,6 +304,16 @@ def add_arguments(self, parser):
             default=["zaken", "besluiten", "documenten"],
         )
 
+        parser.add_argument(
+            "--generate-zaak-inzage-data",
+            action="store_true",
+            default=False,
+            help=(
+                "Create five zaak-inzage benchmark cases containing 1, 2, 3, "
+                "5, and 7 records in every collection."
```
stevenbal, line 313:
```suggestion
                "Create five zaak-inzage benchmark cases containing 1, 3 and "
                "5 records in every collection."
```

src/openzaak/components/zaken/api/urls.py
```diff
@@ -156,6 +158,11 @@
                     zaakverlengen_view,
                     name="zaakverlengen",
                 ),
+                path(
+                    "zaak_inzage/<uuid:uuid>",
```
stevenbal, line 162: I know the other convenience endpoints currently use snake_case, but let's use kebab-case for this one. The Logius API design rules require it, and I'd rather have this endpoint be correct right away: https://github.com/Logius-standaarden/API-Design-Rules/blob/4b1fcca6c17508fc2c168ab2ff4fc19d563a5664/media/linter.yaml#L147

The other endpoints will have to be renamed (I made this part of the Open Zaak 2.0 checklist https://github.com/open-zaak/open-zaak/issues/2322)


```suggestion
                    "zaak-inzage/<uuid:uuid>",
```

src/openzaak/components/zaken/api/viewsets.py
```diff
@@ -2565,3 +2570,54 @@ def _create_audit_logs(self, response, serializer, zaak_version_before_edit):
...
+    serializer_class = ZaakInzageSerializer
+    lookup_field = "uuid"
+    permission_classes = (ZaakAuthRequired,)
+    required_scopes = {
+        "retrieve": SCOPE_ZAKEN_ALLES_LEZEN
+        & SCOPE_CATALOGI_READ
+        & SCOPE_BESLUITEN_ALLES_LEZEN
+    }
+    permission_main_object = "zaak"
+    queryset = Zaak.objects.select_related(
```
stevenbal, line 2597: There seem to be some duplicate queries happening (at least for the `zaken_zaak` table, maybe other ones as well) if I check it locally with silk:

Image

I checked them and they all refer to the same zaak ID, could you check if these duplicate queries can be avoided?

You can run this profiling locally like this: https://open-zaak.readthedocs.io/en/stable/development/performance/profiling.html#profiling-api-endpoints

src/openzaak/components/zaken/api/serializers/zaken.py
```diff
@@ -2014,3 +2028,87 @@ def update(self, instance, validated_data):
...
+    resultaattypen = ResultaatTypeSerializer(many=True, read_only=True)
+    roltypen = RolTypeSerializer(source="roltype_set", many=True, read_only=True)
+    statustypen = StatusTypeSerializer(many=True, read_only=True)
+    zaakobjecttypen = ZaakObjectTypeSerializer(
+        source="zaakobjecttype_set", many=True, read_only=True
+    )
+    informatieobjecttypen = ZaakTypeInformatieObjectTypeSerializer(
+        source="zaaktypeinformatieobjecttype_set", many=True, read_only=True
+    )
+
```
stevenbal, line 2046: Let's also include `besluittypen` here to be complete

src/openzaak/components/zaken/api/serializers/zaken.py
```diff
@@ -2014,3 +2028,87 @@ def update(self, instance, validated_data):
             "zaak": zaak,
             "status": status,
         }
+
+
+class ZaakInzageZaakTypeSerializer(ZaakTypeSerializer):
```
stevenbal, line 2033: The attributes on this serializer currently don't have descriptions in the API schema

Image

Could you add them? I don't think we can pull them from the models since they are reverse FK relations and we can't reuse the ZaakTypeSerializer descriptions either, since those mention URL references.

All of them can be like:

`<resource name> die mogelijk zijn binnen dit ZAAKTYPE` 

e.g. `EIGENSCHAPPEN die mogelijk zijn binnen dit ZAAKTYPE`, etc.

src/openzaak/components/zaken/api/serializers/zaken.py
```diff
@@ -2014,3 +2028,87 @@ def update(self, instance, validated_data):
...
+        source="zaaktypeinformatieobjecttype_set", many=True, read_only=True
+    )
+
+    class Meta(ZaakTypeSerializer.Meta):
+        fields = ZaakTypeSerializer.Meta.fields
+
+
+class ZaakInzageStatusSerializer(StatusSerializer):
+    substatussen = SubStatusSerializer(
+        source="substatus_set", many=True, read_only=True
```
stevenbal, line 2053:
```suggestion
        source="substatus_set", many=True, read_only=True, help_text=_("De SUBSTATUSSEN die horen bij deze STATUS")
```

src/openzaak/components/zaken/api/serializers/zaken.py
```diff
@@ -2014,3 +2028,87 @@ def update(self, instance, validated_data):
...
+    )
+
+    class Meta(StatusSerializer.Meta):
+        fields = StatusSerializer.Meta.fields + ("substatussen",)
+
+
+class ZaakInzageSerializer(ZaakSerializer):
+    zaaktype = ZaakInzageZaakTypeSerializer(read_only=True)
```
stevenbal, line 2061: Let's add these helptexts:

zaaktype: "Het ZAAKTYPE waartoe deze ZAAK behoort."
hoofzaak: "De ZAAK, waarom verzocht is door de initiator daarvan, die behandeld wordt in twee of meer separate ZAAKen waarvan de onderhavige ZAAK er één is."
deelzaken: "Deelzaken die horen bij deze ZAAK."
status: "De huidige STATUS van de zaak, indien er geen status bekend is, dan is de waarde `null`."
eigenschappen: "De ZAAKEIGENSCHAPPEN die horen bij deze ZAAK."
besluiten: "De BESLUITen die betrekking hebben op deze ZAAK."
resultaat: "Het RESULTAAT van deze ZAAK, indien er geen status bekend is, dan is de waarde `null`."
rollen: "De ROLlen die horen bij deze ZAAK."
statussen: "Alle STATUSsen die horen bij deze ZAAK (huidige STATUS en vorige STATUSsen)."
zaakcontactmomenten: "De ZAAKCONTACTMOMENTen die horen bij deze ZAAK."
zaakinformatieobjecten: "De ZAAKINFORMATIEOBJECTen die horen bij deze ZAAK."
zaakobjecten: "De ZAAKOBJECTen die horen bij deze ZAAK."
zaakverzoeken: "De VERZOEKen die horen bij deze ZAAK."
zaaknotities: "De NOTITIES die betrekking hebben op deze ZAAK."

src/openzaak/components/zaken/api/viewsets.py
```diff
@@ -2565,3 +2570,54 @@ def _create_audit_logs(self, response, serializer, zaak_version_before_edit):
...
             main_object=serializer.data["zaak"]["url"],
         )
+
+
+@extend_schema_view(
+    retrieve=extend_schema(
+        "zaakinzage",
+        summary="Geef inzage in een zaak",
+        description=mark_experimental(
+            "Geef de zaak, het zaaktype en alle bijbehorend resources genest terug."
```
stevenbal, line 2580:
```suggestion
            "Geef de zaak, het zaaktype en alle bijbehorende resources genest terug."
```

stevenbal (changes requested):
Looks good overall! 👍

performance_test/test_zaken_list.py
abegailsieffers: With the current PR, the runtime is:

 `real`: 0m56.782s
 `user`: 0m22.275s
 `sys`: 0m1.080s

However, with 5 relations without Documenten and Besluiten  (so nothing is generated for those), the runtime increases to:

`real`: 2m58.141s
`user`: 1m38.892s
`sys`: 0m2.915s

With this result, I would agree that the second option would be better.

src/openzaak/components/zaken/api/viewsets.py
abegailsieffers: With the current changes, I’ve managed to reduce the number of queries from 225 to 67 by using deeper-level prefetching and select_related. However, the actual number of queries still depends on the size of the dataset.

image

performance_test/test_zaak_inzage.py
abegailsieffers: With the current changes. I ran the performance test  does not have much difference with_relationship and without relationship comparison.

with `assert_relationship`
image

without `asset_relationship`
image

src/openzaak/management/commands/generate_data.py
abegailsieffers: Thank you for your feedback. 
I agree! Let's keep the test as it in this PR until we get more information about the data.

performance_test/test_zaak_inzage.py
stevenbal: The only part that is being tested for performance is what's passed to the `benchmark` call, any code outside of that will not be included

src/openzaak/components/zaken/api/viewsets.py
```diff
@@ -2565,3 +2577,117 @@ def _create_audit_logs(self, response, serializer, zaak_version_before_edit):
             basename="resultaat",
             main_object=serializer.data["zaak"]["url"],
         )
+
+
+@extend_schema_view(
+    retrieve=extend_schema(
+        "zaakinzage",
+        summary="Geef inzage in een zaak",
+        description=mark_experimental(
```
stevenbal, line 2586: Could you also mention this endpoint in the documentation `docs/api/experimental.rst`? This makes it easier to keep track of all the experimental changes Open Zaak implements compared to the VNG standard

src/openzaak/components/zaken/api/viewsets.py
stevenbal: Looking much better already :+1:

src/openzaak/components/zaken/api/viewsets.py
```diff
@@ -2565,3 +2577,117 @@ def _create_audit_logs(self, response, serializer, zaak_version_before_edit):
...
+):
+    serializer_class = ZaakInzageSerializer
+    lookup_field = "uuid"
+    permission_classes = (ZaakAuthRequired,)
+    required_scopes = {
+        "retrieve": SCOPE_ZAKEN_ALLES_LEZEN
+        & SCOPE_CATALOGI_READ
+        & SCOPE_BESLUITEN_ALLES_LEZEN
+    }
+    permission_main_object = "zaak"
```
stevenbal, line 2603: Did you get this to work locally with a non superuser token? I configured this, but I keep getting 403 permission denied it seems. I think the issue is that it currently checks all these scopes under the ZRC (zaken API component), but in practice these scopes can only be configured under ZRC, BRC and ZTC currently.

Image

stevenbal (changes requested):

src/openzaak/components/zaken/api/viewsets.py
abegailsieffers: I have not tried testing it with a non superuser token.  I just tested it now locally and it seems I also get 403.
I currently use `ZaakAuthRequired` permission_class but it seem it only checks ZRC component. 
Is there an existing permission on which I can use to check different components? 

PS: In this PR, I cannot save new authorizations in the admin page.

src/openzaak/components/zaken/api/viewsets.py
abegailsieffers: Added!

PR #2540, opened 2026-09-07
:pushpin: [#2484] Use commonground-api-common ETag deletion fix

### LINKED ISSUE #2484: DELETE /zaken/{uuid} returns 500 after successful deletion due to post-commit ETag recalculation

1.28.2

Triggered by: Open Archiefbeheer 2.0.0
Endpoint: DELETE /zaken/api/v1/zaken/{uuid}

Deleting a Zaak through the Zaken API returns an HTTP 500, even though the Zaak and its related resources have already been deleted successfully.

During the cascading deletion, Open Zaak schedules ETag recalculations for the Zaak and related resources such as Rollen, Statussen and Resultaten. After the database transaction has been committed, an ETag callback attempts to process one of these deleted model instances.

Because the instance no longer has a primary key, the callback raises:

TypeError: Model instances without primary key value are unhashable

The zaak_deleted event is logged before the exception occurs. A subsequent GET confirms that the Zaak no longer exists.

1. Create a Zaak.
2. Add related resources, including at least a Status and Resultaat.
3. Send: DELETE /zaken/api/v1/zaken/{uuid}
4. Observe that the response is HTTP 500.
5. Request the same Zaak again and observe that it has already been deleted.

We reproduced this with two separate Zaken while using Open Archiefbeheer.

### Independent reproduction

The issue can also be reproduced independently of Open Archiefbeheer by sending the DELETE request directly to Open Zaak using Bruno (or another API-client).

```http
DELETE /zaken/api/v1/zaken/{uuid}
```

The request returns:

```http
500 Internal Server Error
```

A subsequent GET request for the same Zaak returns `404 Not Found`, confirming that the Zaak was successfully deleted despite the `500` response.

This confirms that the error occurs within Open Zaak itself and is not caused by Open Archiefbeheer.

The DELETE request returns HTTP 204 No Content after the Zaak and its related resources have been deleted.

No ETag recalculation should be attempted for model instances that were deleted within the same transaction.

### Impact

- Open Zaak reports that the deletion failed;
- the Zaak has actually been deleted;
- retrying the request results in a 404;
- clients cannot reliably determine whether the deletion succeeded.

For Open Archiefbeheer this is a blocking issue. OAB receives the 500 and marks the destruction as failed, so the subsequent destruction report and archive Zaak are not created, even though the original Zaak has already been removed.

[stacktrace500Error.txt](https://github.com/user-attachments/files/30505383/stacktrace500Error.txt)

OlhaZahoruiko:
Closes #2484

**Changes**

[Describe the changes here]

**Checklist**

- Commit hygiene

- Architectural design record

  - [x] If a design decision was made that changes functionality, create an architectural design record for it [here](https://github.com/open-zaak/open-zaak/wiki/Architectural-Decision-Records-(ADR))

### DIFF
--- a/src/openzaak/components/zaken/tests/test_zaak_delete.py
+++ b/src/openzaak/components/zaken/tests/test_zaak_delete.py
@@ -6,7 +6,10 @@ from django.test import override_settings, tag
 
 import requests_mock
 from rest_framework import status
-from rest_framework.test import APITestCase
+from rest_framework.test import APITestCase, APITransactionTestCase
+from vng_api_common.authorizations.models import Applicatie
+from vng_api_common.authorizations.utils import generate_jwt
+from vng_api_common.models import JWTSecret
 from vng_api_common.tests import get_validation_errors, reverse
 from zgw_consumers.constants import APITypes, AuthTypes
 from zgw_consumers.test.factories import ServiceFactory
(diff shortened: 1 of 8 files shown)


src/openzaak/components/zaken/tests/test_zaak_delete.py
```diff
@@ -169,3 +172,39 @@ def test_zaak_delete_oio_removed(self, m):
 
         delete_call = next(req for req in m.request_history if req.method == "DELETE")
         self.assertEqual(delete_call.url, zio._objectinformatieobject_url)
+
+
+class ZaakDeleteEtagRegressionTests(APITransactionTestCase):
```
stevenbal, line 178:
```suggestion
@tag("gh-2484")
class ZaakDeleteEtagRegressionTests(APITransactionTestCase):
```

stevenbal (approved):
Glad to see it didn't take as many days as we estimated 😅

PR #2550, opened 2026-09-10
:bug: [#2506] Fix bug daysduration doorlooptijd field in statustype

### LINKED ISSUE #2506: Setting StatusType.doorlooptijd returns 500 (TypeError: relativedelta vs timedelta)

1.29.1 (Docker, `docker compose up` from the repo's own `docker-compose.yml`)

POSTing or PATCHing any non-null value to `doorlooptijd` on `/catalogi/api/v1/statustypen` returns HTTP 500.

The OpenAPI schema advertises the field as writable (`readOnly: false`, `nullable: true`, `format: duration`). Omitting the field, or sending an explicit `null`, works fine — but every actual duration value fails. Tested `P5D`, `P1D`, `P1W`, `P0D`, `P30D`, `PT48H`; all return 500.

Server traceback:

```
File "rest_framework/fields.py", line 553, in run_validators
    validator(value)
File "django/core/validators.py", line 395, in __call__
    if self.compare(cleaned, limit_value):
File "django/core/validators.py", line 420, in compare
    return a > b
TypeError: '>' not supported between instances of 'relativedelta' and 'datetime.timedelta'
```

**Likely root cause.** `StatusType.doorlooptijd` uses `DaysDurationField` from `commonground-api-common` (2.13.0, current release), which attaches validators that compare against `timedelta`:

```python
# vng_api_common/fields.py
class DaysDurationField(models.DurationField):
    def __init__(self, *args, **kwargs):
        ...
        self.default_validators = [
            MinValueValidator(timedelta(days=self.min_duration)),
            MaxValueValidator(timedelta(days=self.max_duration)),
        ]
```

The serializer parses the incoming ISO-8601 duration into a `relativedelta`, and `relativedelta` and `timedelta` are not comparable, so the validator raises `TypeError` rather than producing a validation error.

`ZaakType.doorlooptijd_behandeling` is unaffected — it uses `DurationField` (a `RelativeDeltaField` subclass) instead. As far as I can tell `StatusType.doorlooptijd` is the only field in Open Zaak using `DaysDurationField`.

Possibly the same class of problem as #557 (nullable `brondatumArchiefprocedure` returning 500 on resultaattype create).

**Practical consequence:** `StatusType.doorlooptijd` cannot be populated through the Catalogi API at all, so per-step doorlooptijd data cannot be loaded or migrated via the API.

1. Start the stack from the repository: `docker compose up -d`
2. Create a zaaktype in the default catalogus (any otherwise-valid payload)
3. Create a statustype referencing it, including a `doorlooptijd`:

```http
POST /catalogi/api/v1/statustypen
Content-Type: application/json

{
  "omschrijving": "Behandeling",
  "zaaktype": "<zaaktype url>",
  "volgnummer": 1,
  "doorlooptijd": "P5D"
}
```

4. Observe HTTP 500. The identical request **without** `doorlooptijd`, or with `"doorlooptijd": null`, returns 201.
5. `PATCH`ing `doorlooptijd` onto an existing statustype fails in the same way.

A valid ISO-8601 duration within the field's min/max bounds should be accepted and stored (201/200). A value outside those bounds should return 400 with a validation error — not 500.

danielmursa-dev:
Closes https://github.com/open-zaak/open-zaak/issues/2506

**Changes**

[Describe the changes here]

**Checklist**

- Commit hygiene

- Architectural design record

  - [x] If a design decision was made that changes functionality, create an architectural design record for it [here](https://github.com/open-zaak/open-zaak/wiki/Architectural-Decision-Records-(ADR))


src/openzaak/components/catalogi/tests/test_statustype.py
```diff
@@ -77,7 +78,7 @@ def test_get_detail(self):
             "isEindstatus": True,
             "informeren": False,
             "catalogus": f"http://testserver{reverse(zaaktype.catalogus)}",
-            "doorlooptijd": None,
+            "doorlooptijd": "P10DT00H00M00S",
```
stevenbal: Does the `DurationField` add the hours/minutes/seconds portion as well?

danielmursa-dev: @stevenbal Yes because `DurationField` return a `timedelta`: https://docs.djangoproject.com/en/6.1/ref/models/fields/#durationfield

```
>>> from datetime import timedelta
>>> timedelta(days=1)
datetime.timedelta(days=1)
>>> str(timedelta(days=1))
'1 day, 0:00:00'
```

src/openzaak/components/catalogi/migrations/0018_auto_20230919_1225.py
```diff
@@ -157,11 +157,9 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name="statustype",
             name="doorlooptijd",
-            field=vng_api_common.fields.DaysDurationField(
+            field=openzaak.utils.fields.DurationField(
```
stevenbal: If we mark DaysDurationField as deprecated, we don't have to make this change. Instead what we'll do is when we release Open Zaak 3.0, we squash the migrations to avoid having to make this change

stevenbal (changes requested):
Looks good, only suggesting the change regarding deprecating the DaysDurationField in the library instead of removing it right away.

Because it was essentially impossible to have values for `doorlooptijd` (didn't work via the API and the admin didn't expose it), we won't need a datamigration in my opinion.

stevenbal (approved):

